PR #2, opened 2024-10-22
Stub adjustments

benbz:
Some adjustments now that #1 is merged
* GitHub didn't like the symlinks for the LICENSE files. Replace with actual files (named `LICENSE.<whatever>` rather than `LICENSE-<whatever>` like https://github.com/element-hq/element-internal/issues/635 wishes until https://github.com/fsfe/reuse-tool/pull/1041 is merged
* Switch to `pull_request_target` for the license check now that we've got it on the target branch as it is more secure

### DIFF
--- a/.github/workflows/licensing.yml
+++ b/.github/workflows/licensing.yml
@@ -5,7 +5,7 @@
 name: Licensing checks
 
 on:
-  pull_request:
+  pull_request_target:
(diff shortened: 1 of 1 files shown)


benbz:
#7 moved the licenses back to `LICENSE-<whatever>`

PR #3, opened 2024-11-06
Initial Element Web Helm chart

benbz:
This PR adds an `element-web` sub-chart to the stub `matrix-stack` chart

Design decisions:
* Built output is included in the git history
  * The rationale here is that we want values files (and schemas) to be visible to people browsing the repo.
  * There is some potential pain here for developers in that they've got to remember to run `scripts/construct_helm_scripts` before pushing but that is solved with a CI validation job to prevent it being accidentally missed and we could add pre-commit hooks as well
* The sub-charts for `matrix-stack` are at `charts/<sub-chart>` rather than `charts/matrix-stack/charts/<sub-chart>`
  * The tarballs in `charts/matrix-stack/charts` were causing problems for `ct lint` when we looked at that directory
  * It is also very nested otherwise
* Jinja for values file generation. Various approaches have been tried
  * Generating the default values file from the schemas was very involved. One of the main problems was that the key a given sub-schema was attached to was in the parent schema but the description to go above it is in the sub-schema. Also trying to construct commented out YAML based on `default` was fraught, esp for arrays. I've got various local tests of this so we can revisit later
  * `ruamel-yaml` while it is great at round-tripping commented YAML, it is inscrutable when trying to constructed constructed YAML
  * `yq` removes blank lines the second it touches a file, we want these for readability
  * https://github.com/hairyhenderson/gomplate extra binary but possibly acceptable to reduce the templating languages involved
  * https://github.com/harsh-98/go-template/tree/master looks abandoned but again possibly acceptable to reduce the templating languages involved, if it supports including templates from other templates
    * https://github.com/myback/go-template forked this and looks more up-to-date?

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,3 +6,32 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 # Element Server Suite - Helm Charts
 
 Helm Charts to deploy the Element Server Suite
+
+## Developing
+
+Requirements for development:
+* Python 3.11, 3.12 or 3.13
+  * [poetry](https://python-poetry.org/)
+* [Helm](https://helm.sh/docs/intro/install/)
+* [yq](https://github.com/mikefarah/yq)
+
+Optional Tools:
+* [chart-testing](https://github.com/helm/chart-testing) for Helm linting
+* [reuse](https://reuse.software/)
+* [shellcheck](https://www.shellcheck.net/)
+
+Changes to chart templates are directly made to `chart/<chart>/templates`.
+
+`chart/<chart>/values.yaml` and `chart/<chart>/values.schema.json` are generated files
+and should not be directly edited. Changes to chart values and the values schema are
+made in `chart/<chart>/source`. This is then built by running
+`scripts/construct_helm_charts.sh charts <version>`.
+
+The rationale for this is so that shared values & schema snippets can be shared between
+components without copy-pasting. Shared schema snippets can be found at
+`chart/matrix-stack/sub_schemas/*.json`. Shared values snippets can be found in
+`chart/matrix-stack/sub_schemas/sub_schemas.values.yaml.j2`
+
+The output of `construct_helm_charts.sh` must be committed to Git or CI fails. The rationale
+for this is so that the values files and schemas can be easily viewed in the repo and diffs
+seen in PRs


.github/workflows/build-test.yml
```diff
@@ -0,0 +1,55 @@
+# Copyright 2024 New Vector Ltd
```
kegsay, line 1: > Built output is included in the git history 

How big is the output? It'll be in git history forever so will impact `git clone` times.

I would also personally generate the output to a clearly named `/generated` directory, otherwise it's really hard when reviewing PRs to know which bits are autogenned vs not.

.github/workflows/build-test.yml
```diff
@@ -0,0 +1,55 @@
...
+#
```
kegsay, line 2: > The sub-charts for matrix-stack are at charts/<sub-chart> rather than charts/matrix-stack/charts/<sub-chart>

Sounds good to me, especially if linters were getting upset.

.github/workflows/build-test.yml
```diff
@@ -0,0 +1,55 @@
...
+name: Helm Chart Building tests
+on:
+  pull_request:
+  push:
+    branches:
+    - main
+
+jobs:
+  # We build from source and commit all generated file changes so that we can see the impact in PRs
+  # We want to ensure that the commit of built changes does happen, so fail if building creates changes
```
kegsay, line 18: This assumes building is deterministic, which I hope is true?

charts/element-web/ci/minimal-values.yaml
```diff
@@ -0,0 +1,9 @@
+# Copyright 2024 New Vector Ltd
```
kegsay, line 1: What is the purpose of this file `minimal-values.yaml`? As an example, or actually used somewhere?

If example, what's stopping this getting out of sync with reality? If used somewhere, it's not really clear.

charts/element-web/source/values.schema.json
```diff
@@ -0,0 +1,65 @@
...
+      "$ref": "file://./../../matrix-stack/sub_schemas/podSecurityContext.json"
+    },
+    "resources": {
+      "$ref": "file://./../../matrix-stack/sub_schemas/resources.json"
+    },
+    "tolerations": {
+      "$ref": "file://./../../matrix-stack/sub_schemas/tolerations.json"
+    },
+    "annotations": {
+      "$ref": "file://./../../matrix-stack/sub_schemas/workloadAnnotations.json"
```
kegsay, line 62: The idea behind this is that we're going to be referencing the same blobs all the time for many different components (EW, Synapse, MAS, etc) so we're offloading the schema for those bits to a shared file?

I would personally invert the ordering in that case, so at the top of the file are all the EW specific options and then the bottom is all the generic common cruft.

charts/element-web/source/values.yaml.j2
```diff
@@ -0,0 +1,30 @@
...
+defaultMatrixServer: {}
+  ## URL where the Matrix server is hosted
+  # baseUrl:
+
+  ## Server name of the default Matrix server (if distinct from the baseUrl)
+  # serverName:
+
+# Number of Element Web replicas to start up
+replicas: 2
+{{ sub_schema_values.image(registry='docker.io', repository='vectorim/element-web', tag='v1.11.82') }}
```
kegsay: Confused as I would have thought the chart's appVersion would be used rather than this hard-coded here?

charts/element-web/source/values.yaml.j2
```diff
@@ -0,0 +1,30 @@
...
+{{ sub_schema_values.ingress() }}
+{{ sub_schema_values.containersSecurityContext() }}
+{{ sub_schema_values.extraEnv() }}
+{{ sub_schema_values.nodeSelector() }}
+{{ sub_schema_values.podSecurityContext(user_id='10004', group_id='10004') }}
+{{ sub_schema_values.resources(requests_memory='50Mi', requests_cpu='50m', limits_memory='200Mi') }}
```
kegsay, line 28: Why 10004, 50Mi and 50m?

charts/element-web/values.schema.json
```diff
@@ -0,0 +1,326 @@
+{
```
kegsay, line 1: I assume this is autogenned from `charts/element-web/source/values.schema.json`?

charts/matrix-stack/sub_schemas/ingress.json
```diff
@@ -0,0 +1,35 @@
+{
```
kegsay, line 1: Are these also autogenned? It's hard to tell.

scripts/construct_helm_schema.py
```diff
@@ -0,0 +1,100 @@
...
+import json
+from pathlib import Path
+from typing import Any, Callable
+
+import typer
+
+
+def schema_walker(
+    schema_part: dict[Any], callable: Callable[[dict[any]], dict[any]]
+) -> dict[Any]:
```
kegsay: Needs docs.

charts/matrix-stack/sub_schemas/ingress.json
benbz: These are generated from the operator YAML OpenAPI schemas and sent through a YAML -> JSON and touched up. But they just as appropriately have been written by hand. They aren't going to be auto-generated again, they are allowed to be edited, they are common schema fragments to be used to generated the full chart schemas

charts/element-web/values.schema.json
benbz: Correct but I can't comment inside JSON files like I can with `charts/element-web/values.yaml`

charts/element-web/source/values.yaml.j2
benbz: These are ESS's current values for Element Web

.github/workflows/build-test.yml
benbz: Output size is trivial. The output is `charts/*/values.yaml` and `charts/*/values.schema.json`

.github/workflows/build-test.yml
benbz: It is so far. If it stops being deterministic this job will immediately fail and that's a bug to fix

charts/element-web/source/values.yaml.j2
```diff
@@ -0,0 +1,30 @@
+{#
+Copyright 2024 New Vector Ltd
+
+SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+#}
+{% import 'sub_schema_values.yaml.j2' as sub_schema_values -%}
+{{ sub_schema_values.generatedFileWarning() }}
+
+## Arbitrary extra config to inject into Element Web's config.json a a YAML object
```
gaelgatelement: Because json is a subset of yaml, it should actually be possible to write extra config as json here

charts/element-web/ci/minimal-values.yaml
benbz: `ct lint` uses all `*-values.yaml` files in a chart's `ci` folder to lint against. If the values get out of sync with the template/schema the linting will fail.

It is possible that the filename gets out of sync with reality, i.e. some of these values aren't required and so this isn't really minimal. I don't think there's anyway to protect against that

charts/element-web/source/values.schema.json
benbz: > The idea behind this is that we're going to be referencing the same blobs all the time for many different components (EW, Synapse, MAS, etc) so we're offloading the schema for those bits to a shared file?

Correct. We want our components to have consistent configuration options.

> I would personally invert the ordering in that case, so at the top of the file are all the EW specific options and then the bottom is all the generic common cruft.

:+1:

charts/element-web/source/values.yaml.j2
benbz: We're actually setting the chart's appVersion to this too in `scripts/construct_helm_charts.sh`. I could omit this here for the "main" image in a sub-chart and fall through to `appVersion` but non-main images (e.g. Redis/HAProxy for the Synapse sub-chart) will need tags specified in similar values templates. I defaulted to treating all images consistently, i.e. specifying a tag

charts/element-web/source/values.yaml.j2
```diff
@@ -0,0 +1,30 @@
...
+additional: {}
```
gaelgatelement, line 10: I think we need a way for matrix-stack to inject its own element web config into ElementWeb.

Because of how Helm works, with current structure we are only able to do : 


```yml
matrix-stack:
  element-web:
    additional: # its up to the user to properly configure values here, matrix stack cannot inject anything
```

I would suggest the following mechanism for example, which comes in addition to this `additional` value (pseudocode). However I believe this will not work because the context of the subchart wont be able to look at the parent chart, and it would probably fail (needs testing) 

`element-web/values.yaml` : 
```
additionalHelper: # will call this helper with the $ context
```

In `element-web/templates/_helpers.tpl`, `element-io.element-web.config`  : 

```
{{- set $config := (merge $config .Values.additional }}
{{- set $config := (merge $config (include $.Values.additionalHelper "context" $) }}
```

In `matrix-stack/values.yaml` :  
```
element-web:
  additionalHelper: element-io.matrix-stack.element-web.config
```



`matrix-stack/_element-web-helpers.tpl` : 
```

{{- define "element-io.matrix-stack.element-web.config" }}
{{- $context := .context }}
{{- $config := dict }}
{{- if $context.matrix_content_scanner.enabled }}
{{ $_ := set $config "content_scanner" (dict "enabled" true) }}
{{- end }}
{{ $config }}
{{- end }}
```

How would you envision it this usecase?

scripts/test_construct_helm_schema.py
```diff
@@ -0,0 +1,187 @@
+# Copyright 2024 New Vector Ltd
+#
+# SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+
```
gaelgatelement, line 4: Shouldn't it be tested in a github wokflow as well?

charts/element-web/source/values.yaml.j2
gaelgatelement: I agree with @kegsay, in order for the chart to follow best-practices once built, I  think `image` template should default to `appVersion` if tag is not defined for example.

benbz:
> Overall this seems pretty good.
> 
> I have some reservations around the supplementary tooling though. We have some hand-rolled python scripts to combine sub-schemas, and we have Jinja templates to generate `values.yaml`.

I think our options here are:
* Copy/pasting between components (and sub-components) to maintain `values.yaml` & `values.schema.json` for each sub-chart
* Hand-rolled Python/jq/yq/bash/something to maintain `values.yaml` & `values.schema.json` based on common fragments

Given the potential number of components and sub-components I've opted for the 2nd so things get out of sync. 

I will double check however that the move from  `charts/matrix-stack/charts/<sub-chart>` to `charts/<sub-chart>` doesn't open up the possibility of `$ref: file://...` just working by default from both `matrix-stack` and the sub-chart. If it does, I think we could do the `additionalProperties: false` manually and skip the need for a schema generation script. Otherwise I think that implies hand-rolled script in something for the schemas.

> Rather than introducing a new templating language (Jinja) can we instead lean on the JSON schema to generate the values.yml? Back in the day, we used custom `x-` keys in JSON schema to automatically populated "default" values for the Matrix Spec, so I'm wondering if that could work here?

This was my initial attempt. As discussed in room the main problems were
* Attaching the description above the key the sub-schema attaches to
  * The description could go below, but I think this is less readable and not Helm-like
* Basing when to comment out properties vs having them as actual YAML based on `default` values
* Descriptions for lists

These are solvable problems and I had solutions working in various states. It was a lot more Python and IMO less clear as to what would be generated than what's here now.

charts/element-web/source/values.yaml.j2
benbz: I'm leaving this problem out of this MR. Until we have multiple components we don't have injecting from the `matrix-stack` chart. Until we do a release we can easily change the schema

scripts/test_construct_helm_schema.py
benbz: Now added

scripts/construct_helm_schema.py
benbz: Docs added

charts/element-web/source/values.yaml.j2
benbz: Now defaulting to `appVersion` with `image.tag` being optional

charts/element-web/source/values.schema.json
benbz: EW properties brought to the top of the schema

charts/element-web/source/values.yaml.j2
benbz: Commented updated with this info

benbz:
To say one more thing on generating `values.yaml` from `values.schema.json` vs templating it like here. After the last 2 years I'm much more sceptical of auto-deriving things from schemas. I worry that it is going to be fraught and we won't have the control we end up wanting to make things as readable as possible for people. By enriching objects in the schema with `additionalProperties: false` we allow `ct lint/helm lint` to detect when provided values (including the default `values.yaml` embedded in the chart) are not declared in the schema. Given `ct lint/helm lint` also detect if not enough values are provided to render templates we've got guarantees that the non-commented templated properties are valid. We admittedly have no guarantees that the commented out examples are valid by hand-writing this vs generating from the schema but I think that's an acceptable trade-off.

I'm more concerned that by introducing Jinja templates we've now got 2 different templating languages in this repo vs doing something with Go templates we've only got 2 different dialects (as Helm will add its own functions) of the same templating language

benbz:
> I will double check however that the move from `charts/matrix-stack/charts/<sub-chart>` to `charts/<sub-chart>` doesn't open up the possibility of `$ref: file://...` just working by default from both `matrix-stack` and the sub-chart. If it does, I think we could do the `additionalProperties: false` manually and skip the need for a schema generation script. Otherwise I think that implies hand-rolled script in something for the schemas.

This makes things marginally better, but not good enough. I can now do both `cd charts/element-web && helm lint .` or `cd charts/matrix-stack && helm lint .`. I still can't `helm lint charts/element-web` or `helm lint charts/matrix-stack` at the same time without changing the relative file reference.

https://github.com/helm/helm/pull/12553 / https://github.com/helm/helm/issues/10481 would be required. So I think we keep the schema generation script to do the inlining

gaelgatelement:
> To say one more thing on generating `values.yaml` from `values.schema.json` vs templating it like here. After the last 2 years I'm much more sceptical of auto-deriving things from schemas. I worry that it is going to be fraught and we won't have the control we end up wanting to make things as readable as possible for people. By enriching objects in the schema with `additionalProperties: false` we allow `ct lint/helm lint` to detect when provided values (including the default `values.yaml` embedded in the chart) are not declared in the schema. Given `ct lint/helm lint` also detect if not enough values are provided to render templates we've got guarantees that the non-commented templated properties are valid. We admittedly have no guarantees that the commented out examples are valid by hand-writing this vs generating from the schema but I think that's an acceptable trade-off.
> 
> I'm more concerned that by introducing Jinja templates we've now got 2 different templating languages in this repo vs doing something with Go templates we've only got 2 different dialects (as Helm will add its own functions) of the same templating language

I think that's fine as we are going to have python in this repo for build & testing, and helm / go-templates for runs. 

I believe this is a situation that is acceptable. I would worry a lot more if 2 template languages were mixed up in the same phase.

.github/workflows/build-test.yml
benbz: Resolving now that I've documented the generated files in the readme

.github/workflows/build-test.yml
benbz: Resolving now that the CI job for this is working and it isn't flaking

charts/element-web/ci/minimal-values.yaml
benbz: Resolving now that the CI job for `ct lint` is working

benbz:
Should be said that the Docker image build is manually done (but scripted) rather than in CI. I don't want to deal with this right now :innocent:

charts/element-web/source/values.yaml.j2
benbz: FTR as discussed in the room, this approach does not work. There can be no dynamic variable passing from parent -> subchart unless all the dynamic variables are in `global`

PR #4, opened 2024-11-07
CI: the licensing check doesn't need to be pull_request_target anymore as we're not using the DIND runners

benbz:
#3 3 switched over the `reuse` check to a custom container image and moved away from the DIND runners. We can now safely move this job to `pull_request` not `pull_request_target`

### DIFF
--- a/.github/workflows/licensing.yml
+++ b/.github/workflows/licensing.yml
@@ -5,7 +5,7 @@
 name: Licensing checks
 
 on:
-  pull_request_target:
+  pull_request:
   push:
     branches:
     - main


gaelgatelement:
OOI what's the difference between those WRT DIND runners safety?

benbz:
> OOI what's the difference between those WRT DIND runners safety?

https://runs-on.com/github-actions/pull-request-vs-pull-request-target/.  `pull_request_target` runs in the context of the branch you're targetting. So if you attempted to introduce changes in the jobs in the workflow that attempted to exploit the DIND runner they wouldn't be used.

I suspect I was doing this wrong as well, as the whole checkout is of the target branch, so the licensing check was against the target branch. The checkout job should have explicitly checked out the source branch. Anyway it is irrelevant now

PR #5, opened 2024-11-08
Expose the image digest setting

benbz:
#3 included the logic for setting the digest on the Element Web deployment. It didn't include it in the schema or the values files


benbz:
i.e. https://github.com/element-hq/ess-helm/blob/main/charts/element-web/templates/deployment.yaml#L67-L75

PR #9, opened 2024-11-14
Synapse: initial import

benbz:
Notes:
* Can split the signing key script, sub-schemas and server_name changes out into their own MRs for ease of review if required
* Compared with the `Synapse` CRD:
  * We allow specification of credentials differently. With the CRD there was 1 `Secret` per `Synapse` and specific keys must be provided. With this Helm chart each credential can either be specified directly in Helm (for linting/simplicity) or via a `Secret`+key per credential with no hard-coding of what either must be. This allows the chart to easily consume `Secrets` constructed out of band
  * No credentials are embedded into the bulk of the config templates so they can go into a `ConfigMap`
  * We don't support additional config on a per-worker-type basis, only additional config for the whole instance. We don't believe this was ever used
  * `PersistentVolumeClaim` construction is now supported in the chart (in addition to an existing claim name) so you only need to provide a Postgres
* There are 2 usages of `lookup` only. This is to construct a hash of application service registration files so that `Pods` auto-restart if the registration file changes. This will not work in e.g. ArgoCD but will gracefully degrade to a constant hash so I feel it is acceptable to keep for `helm install` users.
* A usable Postgres can be constructed (that works with the CI values files) with `helm upgrade -i postgres oci://registry-1.docker.io/bitnamicharts/postgresql --set fullnameOverride=ess-postgres,auth.database=synapse,auth.username=synapse_user,auth.password=SuperSecure,primary.initdb.args='--locale=C --encoding=UTF8'`

Potential refactorings I want to do:
* Do `path_map_file` and `path_map_file_get` like the stream writers or user secrets named templates are done. There is an evolution in what I discovered was possible
* Restructure the context passed into `element-io.synapse.process.labels` such that it looks like all the other label templates
* Consider moving away from `n` process-type config templates to at most 3: `main`, `media-repository` and `generic_worker.` To do this we'd need to always binding the HTTP and replication ports in all generic workers. This is probably acceptable they just won't get any traffic on those ports


charts/matrix-stack/sub_schemas/sub_schema_values.yaml.j2
```diff
@@ -34,6 +34,19 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
...
 
+{% macro credential(credential_description, key) %}
+## {{ credential_description }}.
+## It can either be provided inline in the Helm chart e.g.:
+## {{ key }}:
+##   value: SecretValue
+##
+## Or it can be provided via an existing Secret e.g.:
+## {{ key }}:
+##   secret: existing-secret
```
gaelgatelement: This does not match, `postgres-libpq schema` ? What's the reasoning behind using a different pattern for this one?

```

      "required": [
        "password"
      ],
      "not": {
        "required": [
          "passwordSecret",
          "passwordSecretKey"
        ]
```

charts/synapse/templates/_helpers.tpl
```diff
@@ -0,0 +1,45 @@
...
+
+{{- define "element-io.synapse.haproxy.labels" -}}
+app.kubernetes.io/part-of: matrix-stack
+app.kubernetes.io/component: matrix-server-ingress
+app.kubernetes.io/name: synapse-haproxy
+app.kubernetes.io/instance: {{ .Release.Name }}-synapse-haproxy
+app.kubernetes.io/version: {{ .Values.haproxy.image.tag }}
+{{- end }}
+
+{{- define "element-io.synapse.pvcName" -}}
```
gaelgatelement, line 39: Are we happy keeping synapse claim under the old schema, or do we want to move this under a more standard "storage" property, like for all other components ?

Anyway it's probably worth preparing a generic schema for all components needing a claim

charts/synapse/templates/synapse_statefulset.yaml
```diff
@@ -0,0 +1,247 @@
...
+                That would be an acceptable mechanism for getting user secrets in though*/}}
+          export SYNAPSE_MACAROON=$(cat /secrets/{{ .Values.macaroon.secret | default (printf "%s-synapse" $.Release.Name) }}/{{ .Values.macaroon.secretKey | default "MACAROON" }});
+          export SYNAPSE_POSTGRES_PASSWORD=$(cat /secrets/{{ .Values.postgres.password.secret | default (printf "%s-synapse" $.Release.Name) }}/{{ .Values.postgres.password.secretKey | default "POSTGRES_PASSWORD" }});
+          export SYNAPSE_REGISTRATION_SHARED_SECRET=$(cat /secrets/{{ .Values.registrationSharedSecret.secret | default (printf "%s-synapse" $.Release.Name) }}/{{ .Values.registrationSharedSecret.secretKey | default "REGISTRATION_SHARED_SECRET" }});
+
+          set -x;
+          export APPLICATION_NAME=$(hostname -s | sed 's/.*-synapse-//g');
+          cp /config-templates/log_config.yaml /conf/log_config.yaml;
+
+          python3 /config-templates/render_config_with_secrets.py \
```
gaelgatelement, line 110: I think this script might be used by other components, I wonder if we want to build a generic script that can be mounted in any component needing the mechanism.

charts/synapse/templates/synapse_statefulset.yaml
```diff
@@ -0,0 +1,247 @@
...
+{{- with .Values.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
+{{- end }}
+{{- with .Values.hostAliases }}
+      hostAliases:
+        {{- toYaml . | nindent 8 }}
+{{- end }}
+      initContainers:
+      - name: render-config
```
gaelgatelement, line 77: Can you add a comment about the reasoning about why we need to render config in an init container instead of helm helpers ?

charts/synapse/templates/synapse_statefulset.yaml
```diff
@@ -0,0 +1,247 @@
...
+        securityContext:
+          {{- toYaml . | nindent 10 }}
+{{- end }}
+        command:
+        - "bash"
+        - "-ce"
+        - >
+          {{- /*These could be done as env vars with valueFrom.secretKeyRef, but that triggers CKV_K8S_35.
+                That would be an acceptable mechanism for getting user secrets in though*/}}
+          export SYNAPSE_MACAROON=$(cat /secrets/{{ .Values.macaroon.secret | default (printf "%s-synapse" $.Release.Name) }}/{{ .Values.macaroon.secretKey | default "MACAROON" }});
```
gaelgatelement, line 102: The secrets are rendered like this because the user can provide them from an external secret I believe. I think it would be great to have some explanation in comments or the README about this mechanism, since it's probably going to be applied to other components.

charts/synapse/templates/synapse_statefulset.yaml
```diff
@@ -0,0 +1,247 @@
...
+            port: synapse-health
+          periodSeconds: 2
+          failureThreshold: 10
+        livenessProbe:
+          httpGet:
+            path: /health
+            port: synapse-health
+          periodSeconds: 6
+          timeoutSeconds: 2
+          failureThreshold: {{ ternary 8 3 (eq "isSingle" (include "element-io.synapse.process.isSingle" $processType)) }}
```
gaelgatelement, line 194: Why 8 or 3 depending of if the worker is a singleton?

charts/matrix-stack/sub_schemas/sub_schema_values.yaml.j2
gaelgatelement: Mmmh, whhy did it keep this outdated comment..

charts/synapse/templates/synapse_statefulset.yaml
benbz: I'm going to push back on this for now. The script is here due to the lack of `envsubstr` (for replacements) and `yq` (for merging) in the Synapse image. We're relying on this container image a) having Python and b) having YAML installed in `site-packages`. That's all fine for Synapse, it isn't fine for arbitrary components.

If we want to do something generic for arbitrary components, I'd suggest we'd then construct a minimal image with the tooling we need and use that consistently across components that need it.

Let's only do all that if/when we need it though

charts/synapse/templates/synapse_statefulset.yaml
benbz: I'm going to add details to the comments above but I'm pushing back on this until we have other components needing a similar mechanism

charts/synapse/templates/_helpers.tpl
benbz: I'll split out a sub-schema for a claim. I think it will still be rooted under `media` in this case. The rationale being:
* We could require more storage types later, so having `media` in the hierarchy is desirable
* We may want to put additional settings for media that don't relate to the PVC here too, so `media` being above `storage` or `claim` is desirable

charts/synapse/templates/synapse_statefulset.yaml
benbz: Current operator behaviour, but am documenting here

charts/synapse/templates/synapse_statefulset.yaml
benbz: Done

PR #12, opened 2024-11-14
Postgres, credential and hostAlias sub-schemas

benbz:
Extract from #9 for ease of review.

To see usage of these schemas please look at Synapse in #9

### DIFF
--- a/charts/matrix-stack/sub_schemas/hostAliases.json
+++ b/charts/matrix-stack/sub_schemas/hostAliases.json
@@ -1,0 +1,17 @@
+{
+  "type": "array",
+  "items": {
+    "type": "object",
+    "properties": {
+      "ip": {
+        "type": "string"
+      },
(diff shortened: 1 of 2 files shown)


charts/matrix-stack/sub_schemas/postgres-libpq.json
```diff
@@ -0,0 +1,68 @@
...
+      "enum": [
+        "disable",
+        "allow",
+        "prefer",
+        "require",
+        "verify-ca",
+        "verify-full"
+      ]
+    },
+    "password": {
```
gaelgatelement, line 35: Why don't you reuse the credentials schema here?

benbz: Can do. At the time the credential stuff didn't exist and I was trying to minimise the differences vs the CRDs. Happy to change

PR #13, opened 2024-11-14
Add generate-signing-key script

benbz:
Simple script to generate a signing key for Synapse, etc

### DIFF
--- a/scripts/generate-signing-key.py
+++ b/scripts/generate-signing-key.py
@@ -1,0 +1,12 @@
+#!/usr/bin/env python3
+# Copyright 2024 New Vector Ltd
+#
+# SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+import signedjson.key
+
+signing_key = signedjson.key.generate_signing_key(0)
+print(f"Signing key: {signing_key.alg} {signing_key.version} {signedjson.key.encode_signing_key_base64(signing_key)}")
+print(
+    f"Verify key: {signing_key.alg} {signing_key.version} \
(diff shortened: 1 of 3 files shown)


scripts/generate-signing-key.py
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env python3
+# Copyright 2024 New Vector Ltd
+#
+# SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+import signedjson.key
+
+signing_key = signedjson.key.generate_signing_key(0)
+print(f"Signing key: {signing_key.alg} {signing_key.version} {signedjson.key.encode_signing_key_base64(signing_key)}")
```
gaelgatelement, line 8: I wonder if we could also print it as the values file format expect

benbz: That would be component specific though.

In the Synapse PR I could add a comment about this script?

gaelgatelement: FTR the following comment on TI Messenger chart : 

```

  synapse:
    # python3 -c "import signedjson.key; signing_key = signedjson.key.generate_signing_key(0); print(f\"{signing_key.alg} {signing_key.version} {signedjson.key.encode_signing_key_base64(signing_key)}\")"
    # This value must be unique for each deployment
    signingKey: <signing key>
```


led to confusion because people extracted the key part without the alg / version.

I'm wondering how we could make sure people stop making mistake, which might be by writing a python line which pritns the full value `signingKey: ed25519 0 ...` .

This should probably happen against Synapse PR anyway.

PR #14, opened 2024-11-14
Specify server_name via globals

benbz:
Extracted from #9

Lots of components will need access to it. Make it a global


charts/element-web/ci/checkov-values.yaml
```diff
@@ -2,6 +2,9 @@
 #
 # SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 
+global:
+  server_name: ess.localhost
```
gaelgatelement: Since `global` is a key that would be shared with other charts, I wonder if we should scope it under `global.ess` or `global.matrix_stack`

PR #16, opened 2024-11-18
Synapse fixes and other improvements

benbz:
Some fixes following on from #9 and other general improvements coming from testing.

JSON schema changes are consistently auto-formatting at 2 space indentation (we should have linting for this)


charts/synapse/templates/synapse_configmap.yaml
```diff
@@ -64,6 +64,13 @@ data:
     import yaml
 
 
+    def deep_merge_dicts(source, destination):
```
gaelgatelement: Looking at this with fresh eyes, I think it would be better if we could import the python file with a `File` call.

benbz: Possibly Monday morning slowness, but I'm not sure I follow

benbz: Oh right. Put the Python script in its own distinct file and Helm `.Files` it?

gaelgatelement: Yes ! :+1:

gaelgatelement: I very badly expressed it, I forgot that Helm was giving access through a `.Files` property, I thought it was some sort of lookup like ansible does.

PR #22, opened 2024-11-20
Add pytest integration tests

gaelgatelement:
- This PR adds pytest integration tests, call it with `pytest tests`
- The github workflow uses docker-in-docker : as such, it needs maintainer approval to run and will run against the target branch workflow code.
- Environments to tests are configured in `tests/env` directory

### DIFF
--- a/tests/fixtures/cluster.py
+++ b/tests/fixtures/cluster.py
@@ -50,25 +50,30 @@ async def kube_client(cluster):
 
 
 @pytest.fixture(autouse=True, scope="session")
-async def ingress(cluster, kube_client: AsyncClient):
-    async with httpx.AsyncClient() as client:
-        response = await client.get(
-            "https://raw.githubusercontent.com/kubernetes/ingress-nginx/main/deploy/static/provider/kind/deploy.yaml"
-        )
-        ingress_controller_yaml = response.text
-    resources = codecs.load_all_yaml(ingress_controller_yaml)
-    await batch_apply_ordered(kube_client, resources)
-    await asyncio.gather(
-        kube_client.wait(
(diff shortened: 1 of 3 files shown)


charts/synapse/ci/pytest-values.yaml
benbz, line 1: Unused for now?

.github/workflows/pytest.yml
```diff
@@ -0,0 +1,45 @@
...
+
+name: Run Pytest tests
+
+on:
+  pull_request:
+
+
+jobs:
+  pytest:
+    runs-on: dind-privileged-m
```
benbz, line 17: Can we add suitable warnings that we can't go public with this like it is

tests/fixtures/cluster.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+@pytest.fixture(scope="session")
+async def kube_client(cluster):
+    kube_config = KubeConfig.from_file(cluster.kubeconfig)
+    yield AsyncClient(config=kube_config)
+
+
+@pytest.fixture(autouse=True, scope="session")
+async def ingress(cluster, kube_client: AsyncClient):
```
benbz: Not going to block the PR on this but it might be nice to use the ingress chart like we do in the `setup_test_cluster.sh` script

gaelgatelement:
> > The github workflow uses docker-in-docker : as such, it needs maintainer approval to run and will run against the target branch workflow code.
> 
> What in the workflow makes this so?

I need to `/s/pull_request/pull_request_target` first. But because the workflow was not existing on main, I could not test it here. 

I'll drop https://github.com/element-hq/ess-helm/pull/22/commits/21d48a43f3eb295d7bbdc3e1525992137ee9a438 once the PR is validated

PR #24, opened 2024-11-21
Add ServiceMonitors

benbz:


charts/matrix-stack/sub_schemas/sub_schema_values.yaml.j2
```diff
@@ -248,7 +248,14 @@ global:
     memory: {{ limits_memory }}
 {%- endmacro %}
 
-{% macro tolerations(key='tolerations')%}
+{% macro serviceMonitors(key='serviceMonitors') %}
+## Whether to deploy ServiceMonitors into the cluster for this component
+## Requires the ServiceMonitor CRDs to be in the cluster
+{{ key }}:
```
gaelgatelement, line 254: What about "auto" mode where it would use Helm feature detection capabilities?

benbz: That's effectively what we have as we have the capabilities check in the manifest template and the values fragments default to true. We're missing the force mode to ignore the capability check. I think that's acceptable for now. We can always add a `serviceMonitor.force: true` to override the capability check in future if needed

gaelgatelement: OK, so `enabled` = `auto` actually. LGTM for now

PR #29, opened 2024-11-21
tests: implement basic synapse deployment tests

gaelgatelement:
- This just starts synapse and checks that it answers on its unauthenticated api endpoint

### DIFF
--- a/.github/workflows/scripts-linting.yml
+++ b/.github/workflows/scripts-linting.yml
@@ -53,14 +53,12 @@ jobs:
         echo "$(poetry env info -p)/bin" >> "${GITHUB_PATH}"
 
     - name: Run ruff check
-      uses: astral-sh/ruff-action@v1
-      with:
-        args: "check"
+      run: |
+        ruff check
 
     - name: Run ruff format
-      uses: astral-sh/ruff-action@v1
-      with:
-        args: "format --check"
+      run: |
+        ruff format --check
 
     - name: pytest scripts
       run: |


.github/workflows/pytest.yml
```diff
@@ -5,7 +5,7 @@
 name: Run Pytest tests
 
 on:
-  pull_request_target:
+  pull_request:
```

tests/services/postgresql.py
```diff
@@ -0,0 +1,88 @@
...
+                        "scripts": {
+                            "init.sh": "#!/bin/sh\n"
+                            "(echo -n $POSTGRES_POSTGRES_PASSWORD | psql -W -U postgres -tc "
+                            '"SELECT 1 FROM pg_database WHERE datname = '
+                            f"'{self.database}' "
+                            '" | grep -q 1) || '
+                            "(echo -n $POSTGRES_POSTGRES_PASSWORD | createdb --encoding=UTF8 "
+                            "--locale=C --template=template0 "
+                            f"--owner={self.user} {self.database} -U postgres) "
+                        }
```
benbz: What does this give us vs `primary.initdb.args: --locale=C --encoding=UTF8` like the `setup_test_cluster.sh` script does? i.e. should that be changed to this?

tests/services/postgresql.py
```diff
@@ -0,0 +1,88 @@
...
+                    },
+                    "persistence": {"enabled": True, "size": "8Gi"},
```
benbz: Do we need this for tests?

tests/services/postgresql.py
```diff
@@ -0,0 +1,88 @@
...
+
+    async def setup(self, helm_client: Client, kube_client):
+        await kube_client.create(
+            Secret(
+                metadata=ObjectMeta(
+                    name=f"{self.name}-postgres-db",
+                    namespace=self.namespace,
+                    labels={"app.kubernetes.io/created-by": "pytest"},
+                ),
+                stringData={"adminPassword": secrets.token_urlsafe(36), "password": self.password},
```
benbz, line 37: Any reason to use `token_urlsafe` for this and all the other passwords? We've deliberately tested non-urlsafe creds in the past to confirm we're escaping things correctly

tests/fixtures/data.py
```diff
@@ -36,6 +38,18 @@ class ESSData:
     def ess_namespace(self):
         return f"ess-{self.secrets_random}"
 
+    def ess_secret(self):
+        return Secret(
+            metadata=ObjectMeta(
+                name="ess-secrets", namespace=self.ess_namespace, labels={"app.kubernetes.io/created-by": "pytest"}
```
benbz: Appreciate we've used this previously lots, but it appears this label is [deprecated](https://kubernetes.io/docs/reference/labels-annotations-taints/#app-kubernetes-io-created-by-deprecated).

Probably should be `app.kubernetes.io/managed-by` ?

charts/synapse/ci/pytest-values.yaml
benbz, line 1: I realise now that this and #22 are in the sub-charts. Is that where we want this to happen or should we do `charts/matrix-stack/ci/pytest-<component>-values.yaml`?

Or both in someway?

tests/services/postgresql.py
gaelgatelement: This gives us idempotency - it was useful in molecule/ansible context, but .probably useless in the context of pytest. I'll simplify

charts/synapse/ci/pytest-values.yaml
gaelgatelement: I think we should do both. Tests for individual charts, and tests for matrix-stack chart.

tests/fixtures/data.py
```diff
@@ -19,6 +21,11 @@ def generate_signing_key():
     return value
 
 
+def unsafe_token(size):
+    alphabet = string.ascii_letters + string.digits
```
benbz: Isn't this alphabet actually safe? Not seeing any symbols or the like in here?

charts/synapse/ci/pytest-values.yaml
gaelgatelement: (I'll do matrix-stack test chart in the context of multi-components deployment, so for SBG)

PR #37, opened 2024-11-26
tests: refactor and rely on matrix-stack only

gaelgatelement:
- Env files now contain the path to the pytest value file
 - All charts deployment now rely on `matrix-stack` to get ready for charts flattening

### DIFF
--- a/charts/matrix-stack/ci/pytest-element-web-values.yaml
+++ b/charts/matrix-stack/ci/pytest-element-web-values.yaml
@@ -35,3 +35,6 @@ element-web:
             "feature_element_call_video_rooms": true,
           },
       }
+
+synapse:
+  enabled: false


charts/synapse/values.yaml
benbz, line 1: Shouldn't be disappearing?

tests/fixtures/synapse.py
```diff
@@ -13,18 +13,19 @@
...
-logging.getLogger("pyhelm3.commands").setLevel(logging.ERROR)
+logging.getLogger("pyhelm3").setLevel(logging.ERROR)
 
 
 @pytest.fixture(scope="session")
 async def synapse_ready(
     cluster, kube_client: AsyncClient, helm_client: pyhelm3.Client, ingress, ess_namespace, generated_data: ESSData
 ):
     counter = 0
+
```
benbz: Can all the below be updated to use the `revision_deployed` stuff?

charts/synapse/values.yaml
gaelgatelement: It's going to happen in next PR

benbz: I just expect that this disappearing is causing lots of your CI failures

gaelgatelement: ah I read "should it be disappearing ?" :facepalm:

tests/fixtures/synapse.py
benbz: Have I misunderstood what `revision_deployed` should do? I'd have thought lines 22-38 could be replaced by it somehow similar to what EW does

gaelgatelement: oh you are right, this should be simplifyed greatly indeed.

.github/workflows/build-test.yml
```diff
@@ -92,6 +95,9 @@ jobs:
         poetry install
         echo "$(poetry env info -p)/bin" >> "${GITHUB_PATH}"
 
+    - name: Update helm dependencies recursively
```
benbz: Why is this needed as well as l68-70?

PR #39, opened 2024-11-27
Flatten Element Web Chart - Step 1, refactor to move everything under a top level key ElementWeb

gaelgatelement:
- Step 1 is refactoring to move everything under a top level key
- Enhance element-web helpers to standardize input with global & local context
- Dont break everything in one PR to ease reviewing


gaelgatelement:
> I imagine most of the oddness (double `element-web` when used from `matrix-stack`, supporting multiple formats in `ess-library`, `global.ess` + `ess`) here is a result of the intermediate steps and will disappear in a latter PR?

Indeed!

PR #41, opened 2024-11-27
Flatten Synapse structure

gaelgatelement:
- Due to credentials being present in root part, we are now enforcing their proper usage with a helper
 - Templates also need to be called with the `(dict "global" $ "context" .)` pattern
 - Similarly to https://github.com/element-hq/ess-helm/pull/39, Step 1 is refactoring to move everything under a top level key

### DIFF
--- a/charts/synapse/configs/haproxy/haproxy.cfg.tpl
+++ b/charts/synapse/configs/haproxy/haproxy.cfg.tpl
@@ -1,6 +1,8 @@
-# Copyright 2024 New Vector Ltd
-#
-# SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+{{- /*
+Copyright 2024 New Vector Ltd
+
+SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+*/ -}}
 
 {{- $global := .global -}}
 {{- with required "haproxy.cfg.tpl missing context" .context -}}


charts/ess-library/templates/_credentials.tpl
```diff
@@ -0,0 +1,24 @@
+# Copyright 2024 New Vector Ltd
+#
+# SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+
+{{- define "element-io.ess-library.check-credential" -}}
+{{- $global := .global -}}
+{{- with required "element-io.ess-library.check-credential" .context -}}
```
benbz:
```suggestion
{{- with required "element-io.ess-library.check-credential missing context" .context -}}
```

 Is what you've done in the EW helpers

charts/ess-library/templates/_credentials.tpl
```diff
@@ -0,0 +1,24 @@
...
+{{- $secretPath := .secretPath -}}
+{{- $secretProperty := .secretProperty -}}
+{{- if and .secretProperty.value (or .secretProperty.secret .secretProperty.secretKey) -}}
+{{- fail printf "The secret %s must either have a value, or a secret/secretKey properties" $secretPath -}}
+{{- else if and .secretProperty.secret (not .secretProperty.secretKey) -}}
+{{- fail printf "The secret %s has a secret but no secretKey properties" $secretPath -}}
```
benbz:
```suggestion
{{- fail printf "The secret %s has a secret but no secretKey property" $secretPath -}}
```

charts/ess-library/templates/_credentials.tpl
```diff
@@ -0,0 +1,24 @@
...
+{{- else if and .secretProperty.secretKey (not .secretProperty.secret) -}}
+{{- fail printf "The secret %s has a secretKey but no secret properties" $secretPath -}}
```
benbz:
```suggestion
{{- fail printf "The secret %s has a secretKey but no secret property" $secretPath -}}
```

charts/ess-library/templates/_credentials.tpl
```diff
@@ -0,0 +1,24 @@
...
+#
+# SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+
+{{- define "element-io.ess-library.check-credential" -}}
+{{- $global := .global -}}
+{{- with required "element-io.ess-library.check-credential" .context -}}
+{{- $secretPath := .secretPath -}}
+{{- $secretProperty := .secretProperty -}}
+{{- if and .secretProperty.value (or .secretProperty.secret .secretProperty.secretKey) -}}
+{{- fail printf "The secret %s must either have a value, or a secret/secretKey properties" $secretPath -}}
```
benbz:
```suggestion
{{- fail printf "The secret %s must either have a value, or both secret & secretKey properties" $secretPath -}}
```

PR #42, opened 2024-11-27
Normalize root & context passing to helpers

gaelgatelement:
- `global` should really be named `root` to avoid confusion with `$.global` special variable
- we use the dict pattern to enforce parameters in all helpers, including ess-library

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -76,6 +76,14 @@ jobs:
         git config --global --add safe.directory "$GITHUB_WORKSPACE"
         ct lint --config ct.yaml
 
+    - name: Ensure we don't use the helm $ root token in .tpl files
+      shell: bash
+      run: |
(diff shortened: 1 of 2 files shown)


charts/matrix-stack/sub_schemas/sub_schema_values.yaml.j2
```diff
@@ -80,7 +80,9 @@ global:
 ## {{ key }}:
 ##   secret: existing-secret
 ##   secretKey: key-in-secret
-{{ key }}: {}
+{{ key }}:
+   secret: change-me
+   secretKey: change-me-key
```
benbz: I don't think we should have this change

gaelgatelement: It was indeed a rebase mistake

charts/synapse/values.yaml
```diff
@@ -73,7 +73,9 @@ synapse:
   ## signingKey:
   ##   secret: existing-secret
   ##   secretKey: key-in-secret
-  signingKey: {}
+  signingKey:
```
benbz: Needs regenerating

PR #43, opened 2024-11-27
Merge Element Web down into Matrix-Stack

gaelgatelement:
- Follow up of https://github.com/element-hq/ess-helm/pull/39 - Element Web is now part of matrix-stack chart


charts/matrix-stack/ci/example-default-enabled-components.values.yaml
```diff
@@ -8,17 +8,16 @@ global:
...
-      defaultMatrixServer:
-        baseUrl: https://synapse.ess.localhost
+  elementWeb:
+    defaultMatrixServer:
+      baseUrl: https://synapse.ess.localhost
 
-      ingress:
-        host: element.ess.localhost
+    ingress:
+      host: element.ess.localhost
```
benbz: This doesn't look like the others / is still doubly nested

PR #45, opened 2024-11-27
charts: merge synapse down into matrix-stack

gaelgatelement:
- Follow up of https://github.com/element-hq/ess-helm/pull/41 - Synapse is now part of matrix-stack chart

### DIFF
--- a/REUSE.toml
+++ b/REUSE.toml
@@ -11,7 +11,7 @@ SPDX-FileCopyrightText = "Copyright 2024 New Vector Ltd"
 SPDX-License-Identifier = "AGPL-3.0-only OR LicenseRef-Element-Commercial"
 
 [[annotations]]
-path = "charts/**/source/sub_schemas/*.json"
+path = "charts/**/sub_schemas/*.json"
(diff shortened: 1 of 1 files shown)


REUSE.toml
```diff
@@ -15,6 +15,11 @@ path = "charts/**/source/sub_schemas/*.json"
 SPDX-FileCopyrightText = "Copyright 2024 New Vector Ltd"
 SPDX-License-Identifier = "AGPL-3.0-only OR LicenseRef-Element-Commercial"
 
+[[annotations]]
+path = "charts/**/sub_schemas/*.json"
```
benbz, line 14: Unsure if the component specific sub-schemas should go directly in `source/<component>` rather than `source/<component>/sub_schemas`. And possibly the common ones in `source/common` or `source/library` so everything is consistent. That can be a separate PR though

Regardless I think the annotation above disappears

gaelgatelement: I like your reorg suggestion. I'll do so in a follow-up PR.

PR #49, opened 2024-11-28
elementWeb: use Synapse ingress host for matrix server base url

gaelgatelement:
- Let's profit from actually sharing the same chart

### DIFF
--- a/charts/matrix-stack/templates/element-web/_helpers.tpl
+++ b/charts/matrix-stack/templates/element-web/_helpers.tpl
@@ -27,12 +27,13 @@ app.kubernetes.io/version: {{ .image.tag | default $root.Chart.AppVersion }}
 {{- with required "element-io.element-web.config missing context" .context -}}
 {{- $config := dict -}}
 {{- $serverName := required "Element Web requires .ess.serverName set" $root.Values.ess.serverName -}}
+{{- $mHomeserver := dict "server_name" $serverName }}
 {{- if $root.Values.synapse.enabled }}
 {{- $baseUrl := "https://{{ $root.Values.synapse.ingress.host }}" -}}
-{{- $mHomeserver := dict "base_url" $baseUrl "serverName" $serverName -}}
+{{- $mHomeserver := merge (dict "base_url" $baseUrl) $mHomeserver }}
+{{- end }}
 {{- $defaultServerConfig := dict "m.homeserver" $mHomeserver -}}
 {{- $_ := set $config "default_server_config" $defaultServerConfig -}}
-{{- end }}
 {{- toPrettyJson (merge $config .additional) -}}
 {{- end }}
 {{- end }}


charts/matrix-stack/templates/element-web/_helpers.tpl
```diff
@@ -25,14 +25,19 @@ app.kubernetes.io/version: {{ .image.tag | default $root.Chart.AppVersion }}
...
 {{- with required "element-io.element-web.config missing context" .context -}}
-{{- $config := dict }}
-{{- $serverName := required "Element Web requires .ess.serverName set" $root.Values.ess.serverName }}
-{{- with required "elementWeb.defaultMatrixServer is required" .defaultMatrixServer }}
-{{- $baseUrl := required "elementWeb.defaultMatrixServer.baseUrl is required" .baseUrl -}}
-{{- $mHomeserver := dict "base_url" $baseUrl "serverName" $serverName }}
+{{- $config := dict -}}
+{{- $serverName := required "Element Web requires .ess.serverName set" $root.Values.ess.serverName -}}
+{{- if .defaultMatrixServer -}}
+{{- $baseUrl := required "elementWeb.defaultMatrixServer.baseUrl is required" .defaultMatrixServer.baseUrl -}}
```
benbz: Not sure this is the test we want to use. If we add another property to `defaultMatrixServer` and that has a value this check gets hit.

I'm tempted to say let's go further than this PR. Remove `defaultMatrixServer`. Populate the config with the serverName and optionally the Synapse hostname. Additional config can provide the remote homeserver if the Synapse isn't enabled. The `defaultMatrixServer` in EW's config isn't required, it just won't have a default if not configured

charts/matrix-stack/templates/element-web/_helpers.tpl
```diff
@@ -25,14 +25,14 @@ app.kubernetes.io/version: {{ .image.tag | default $root.Chart.AppVersion }}
...
 {{- $root := .root -}}
 {{- with required "element-io.element-web.config missing context" .context -}}
-{{- $config := dict }}
-{{- $serverName := required "Element Web requires .ess.serverName set" $root.Values.ess.serverName }}
-{{- with required "elementWeb.defaultMatrixServer is required" .defaultMatrixServer }}
-{{- $baseUrl := required "elementWeb.defaultMatrixServer.baseUrl is required" .baseUrl -}}
-{{- $mHomeserver := dict "base_url" $baseUrl "serverName" $serverName }}
+{{- $config := dict -}}
+{{- $serverName := required "Element Web requires .ess.serverName set" $root.Values.ess.serverName -}}
+{{- if $root.Values.synapse.enabled }}
```
benbz, line 31: `server_name` should be independently set of whether Synapse is enabled

charts/matrix-stack/templates/element-web/_helpers.tpl
```diff
@@ -25,14 +25,14 @@ app.kubernetes.io/version: {{ .image.tag | default $root.Chart.AppVersion }}
...
+{{- $baseUrl := "https://{{ $root.Values.synapse.ingress.host }}" -}}
+{{- $mHomeserver := dict "base_url" $baseUrl "serverName" $serverName -}}
```
benbz:
```suggestion
{{- $mHomeserver := dict "base_url" $baseUrl "server_name" $serverName -}}
```

PR #53, opened 2024-12-02
synapse: enabled workers detection helpers dont need local context, only root

gaelgatelement:
- `Values` is not used in helpers anymore, since https://github.com/element-hq/ess-helm/pull/42
- Remove `context` from helpers which dont need it under `synapse`, as it simplifies a lot how we pass values to workers templates

### DIFF
--- a/charts/matrix-stack/templates/synapse/_synapse_details.tpl
+++ b/charts/matrix-stack/templates/synapse/_synapse_details.tpl
@@ -98,7 +98,6 @@ responsibleForMedia
 
 {{- define "element-io.synapse.streamWriterWorkers" -}}
 {{- $root := .root -}}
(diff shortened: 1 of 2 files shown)


charts/matrix-stack/templates/synapse/_synapse_details.tpl
```diff
@@ -100,7 +100,7 @@ responsibleForMedia
 {{- $root := .root -}}
 {{- with required "element-io.synapse.streamWriterWorkers missing context" .context -}}
```
benbz: This isn't needed now either

benbz: Also there's a trailing `q` on line 109 but I can't comment on it

PR #74, opened 2024-12-06
synapse: fix additional configuration pass

gaelgatelement:


charts/matrix-stack/templates/synapse/synapse_secret.yaml
```diff
@@ -14,8 +14,13 @@ metadata:
   name: {{ $.Release.Name }}-synapse
   namespace: {{ $.Release.Namespace }}
 data:
-{{- with .additional.config }}
-  02-homeserver-helm-user-provided.yaml: {{ . | b64enc }}
+{{- with .additional }}
+{{- range $key := (. | keys | uniq | sortAlpha) }}
+{{- $prop := index $.Values.synapse.additional $key }}
+{{- if $prop.config }}
+  {{ $key }}: {{ $prop.config | b64enc }}
```
benbz: Can we prefix this with `user-` or something so there's no chance of clashing

gaelgatelement: Sure!

PR #86, opened 2024-12-12
synapse: wait for postgres to avoid restarts

gaelgatelement:


charts/matrix-stack/templates/synapse/_helpers.tpl
```diff
@@ -92,3 +92,20 @@ app.kubernetes.io/version: {{ .image.tag }}
 {{- end -}}
 {{- end }}
 {{- end }}
+
+{{- define "element-io.synapse.env" }}
+{{- $root := .root -}}
+{{- with required "element-io.sygnal.labels missing context" .context -}}
```
```suggestion
{{- with required "element-io.synapse.labels missing context" .context -}}
```

charts/matrix-stack/templates/synapse/synapse_statefulset.yaml
```diff
@@ -162,12 +162,11 @@ We have an init container to render & merge the config for several reasons:
         - "-cex"
         - >
           ls -lR /secrets;
+          while ! bash -c "cat < /dev/null > /dev/tcp/$POSTGRES_HOST/$POSTGRES_PORT" 2>/dev/null; do sleep 1; done;
```
benbz: Can we put an `echo` in the loop please

PR #88, opened 2024-12-13
tests: implement ingress/services networking verification

gaelgatelement:
- Verify that all ingress services backends points to endpoints with an address after all ingresses became ready
- The verification that ingresses points to services is done in the manifest tests

### DIFF
--- a/tests/integration/test_networking.py
+++ b/tests/integration/test_networking.py
@@ -51,24 +51,17 @@ async def test_services_have_endpoints(
     kube_client: AsyncClient,
     generated_data: ESSData,
 ):
-    endpoints_by_name = dict[str, Endpoints]()
-    async for endpoint in kube_client.list(Endpoints, namespace=generated_data.ess_namespace):
-        assert endpoint.metadata is not None, f"Encountered an endpoint without metadata : {endpoint}"
-        endpoints_by_name[endpoint.metadata.name] = endpoint
     async for service in kube_client.list(
         Service, namespace=generated_data.ess_namespace, labels={"app.kubernetes.io/part-of": op.in_(["matrix-stack"])}
     ):
         assert service.metadata is not None, f"Encountered a service without metadata : {service}"
-        endpoint = endpoints_by_name[service.metadata.name]
-        assert endpoint.metadata is not None, f"Encountered an endpoint without metadata : {endpoint}"
         await asyncio.to_thread(
             cluster.wait,
-            name=f"endpoints/{endpoint.metadata.name}",
+            name=f"endpoints/{service.metadata.name}",
             namespace=generated_data.ess_namespace,
             waitfor="jsonpath='{.subsets[].addresses}'",
         )
-        # We refresh the endpoint to get the latest state
-        endpoint = await kube_client.get(Endpoints, name=endpoint.metadata.name, namespace=generated_data.ess_namespace)
+        endpoint = await kube_client.get(Endpoints, name=service.metadata.name, namespace=generated_data.ess_namespace)
         assert endpoint.metadata is not None, f"Encountered an endpoint without metadata : {endpoint}"
         assert endpoint.subsets, f"Endpoint {endpoint.metadata.name} has no subsets"
 


tests/integration/fixtures/__init__.py
```diff
@@ -23,4 +23,5 @@
     "ESSData",
     "matrix_stack",
     "helm_prerequisites",
+    "all_ingresses_ready",
```
benbz: Something that's been annoying me for a bit and isn't the fault of this PR, can we sort this list. I found 2x of one function in a previous PR

tests/integration/test_networking.py
```diff
@@ -0,0 +1,44 @@
...
+    endpoints_by_name = dict[str, Endpoints]()
+    async for endpoint in kube_client.list(Endpoints, namespace=generated_data.ess_namespace):
+        assert endpoint.metadata is not None, f"Encountered an endpoint without metadata : {endpoint}"
+        endpoints_by_name[endpoint.metadata.name] = endpoint
+    async for service in kube_client.list(
+        Service, namespace=generated_data.ess_namespace, labels={"app.kubernetes.io/part-of": op.in_(["matrix-stack"])}
+    ):
+        assert service.metadata is not None, f"Encountered a service without metadata : {service}"
+        endpoint = endpoints_by_name[service.metadata.name]
+        assert endpoint.subsets, f"Endpoint {service.metadata.name} has no subsets"
```
benbz: Just because the ingress has been registered with the ingress controller, doesn't mean the Pods behind the services are ready / have endpoints

I think the fixture used in this test should be `all_pods_ready`

gaelgatelement: The `all_ingresses_ready` depends on `matrix_stack` which waits for the helm to be deployed - All pods should be ready I think?

I think ingresses dont get ready if the services have an issue behind. But actually we might not want to wait on this, in order to detect services having issues if an ingress never comes up. WDYT?

benbz:
> The verification that ingresses points to services is done in the manifest tests

This is not in the current manifest tests FTR. Was this missed off this PR?

tests/integration/test_networking.py
benbz:
```
$ kubectl get ingress -n ess ess-element-web -o jsonpath='{.status}' 
{"loadBalancer":{"ingress":[{"ip":"192.168.112.2"}]}}                                                                                                                                                                                    $ kubectl delete deployment ess-element-web -n ess
deployment.apps "ess-element-web" deleted
$ kubectl get ingress -n ess ess-element-web -o jsonpath='{.status}'
{"loadBalancer":{"ingress":[{"ip":"192.168.112.2"}]}}
$ kubectl delete service ess-element-web -n ess           
service "ess-element-web" deleted
$ kubectl get ingress -n ess ess-element-web -o jsonpath='{.status}'
{"loadBalancer":{"ingress":[{"ip":"192.168.112.2"}]}}
```

The status check for an `Ingress` is literally just whether it has been registered with the controller. 

The `matrix-stack` `wait=True` is more interesting as that should wait for `Waits until all Pods are in a ready state, PVCs are bound, Deployments have minimum (Desired minus maxUnavailable) Pods in ready state and Services have an IP address` which should be enough and we not need any additional fixture here. Which makes me wonder why https://github.com/element-hq/ess-helm/actions/runs/12313317890/job/34367084805?pr=88 is failing.

benbz: https://github.com/element-hq/ess-helm/actions/runs/12313317890/job/34367084805?pr=88#step:6:50:
`all_ingresses_ready = <_GatheringFuture pending>`

Which appears as though the fixture isn't actually waiting for the ingresses to be ready which also makes me wonder whether the `matrix_stack` fixture hasn't finished waiting

tests/integration/fixtures/__init__.py
gaelgatelement: I removed the fixture but sorted the list

tests/integration/test_networking.py
gaelgatelement: It's because I returned the wrong thing : `
    return asyncio.gather(*waits)` should have been `
    return await asyncio.gather(*waits)`

Anyway, this is now useless as we discussed, so I removed the fixture.

gaelgatelement:
> > The verification that ingresses points to services is done in the manifest tests
> 
> This is not in the current manifest tests FTR. Was this missed off this PR?

Well, I saw one of the 3 tests already implemented (the `Prefix` check), and I assumed too quickly that all were. I added the missing tests.

tests/manifests/test_ingresses.py
```diff
@@ -204,3 +204,34 @@ async def test_component_ingressClassName_beats_global(component, values, make_t
         if template["kind"] == "Ingress":
             assert "ingressClassName" in template["spec"]
             assert template["spec"]["ingressClassName"] == "component"
+
+
+@pytest.mark.parametrize("component", components_with_ingresses)
+@pytest.mark.asyncio_cooperative
+async def test_ingress_services(component, templates):
```
benbz: Needs rebasing as otherwise will break on merge. Needs to look something like

```suggestion
@pytest.mark.parametrize("values_file", values_files_with_ingresses)
@pytest.mark.asyncio_cooperative
async def test_ingress_services(templates):
```

tests/integration/test_networking.py
```diff
@@ -0,0 +1,89 @@
...
+async def test_services_have_endpoints(
+    cluster,
+    matrix_stack,
+    kube_client: AsyncClient,
+    generated_data: ESSData,
+):
+    endpoints_by_name = dict[str, Endpoints]()
+    async for endpoint in kube_client.list(Endpoints, namespace=generated_data.ess_namespace):
+        assert endpoint.metadata is not None, f"Encountered an endpoint without metadata : {endpoint}"
+        endpoints_by_name[endpoint.metadata.name] = endpoint
```
benbz:
```suggestion
```

Don't think there's any point in doing this anymore

tests/integration/test_networking.py
```diff
@@ -0,0 +1,89 @@
...
+    async for service in kube_client.list(
+        Service, namespace=generated_data.ess_namespace, labels={"app.kubernetes.io/part-of": op.in_(["matrix-stack"])}
+    ):
+        assert service.metadata is not None, f"Encountered a service without metadata : {service}"
+        endpoint = endpoints_by_name[service.metadata.name]
+        assert endpoint.metadata is not None, f"Encountered an endpoint without metadata : {endpoint}"
+        await asyncio.to_thread(
+            cluster.wait,
+            name=f"endpoints/{endpoint.metadata.name}",
```
benbz:
```suggestion
        await asyncio.to_thread(
            cluster.wait,
            name=f"endpoints/{service.metadata.name}",
```

tests/integration/test_networking.py
```diff
@@ -0,0 +1,83 @@
...
+        Service, namespace=generated_data.ess_namespace, labels={"app.kubernetes.io/part-of": op.in_(["matrix-stack"])}
+    ):
+        assert service.metadata is not None, f"Encountered a service without metadata : {service}"
+        await asyncio.to_thread(
+            cluster.wait,
+            name=f"endpoints/{service.metadata.name}",
+            namespace=generated_data.ess_namespace,
+            waitfor="jsonpath='{.subsets[].addresses}'",
+        )
+        # We refresh the endpoint to get the latest state
```
benbz:
```suggestion
```

Whops sorry missed this

PR #94, opened 2024-12-13
Global tolerations and tests

benbz:
* Support setting `tolerations` for all components in the chart
  * If (sub)component `tolerations` and global `tolerations` are set, they're merged.
* Remove `taints` and `tolerations` at the integration test layer, the manifest tests cover all components, inc new ones as they're added (via `test_all_components_covered`)


tests/manifests/test_tolerations.py
```diff
@@ -0,0 +1,89 @@
...
+        values[component].setdefault(sub_component, {}).setdefault("tolerations", []).append(specific_toleration)
+
+    for template in await make_templates(values):
+        if template["kind"] in ["Deployment", "StatefulSet"]:
+            id = f"{template["kind"]}/{template["metadata"]["name"]}"
+
+            pod_spec = template["spec"]["template"]["spec"]
+            assert "tolerations" in pod_spec, f"No tolerations for {id}"
+            assert len(pod_spec["tolerations"]) == 1, f"Wrong number of tolerations for {id}"
+            assert pod_spec["tolerations"][0] == specific_toleration, f"Toleration isn't as expected for {id}"
```
gaelgatelement, line 50: Could error messages show expected / actual, else it's a bit painful to diagnose ?

benbz: You get them for free, e.g. after removing the `uniq` call
```
>               assert len(pod_spec["tolerations"]) == 2, f"Wrong number of tolerations for {id}"
E               AssertionError: Wrong number of tolerations for Deployment/pytest-synapse-haproxy
E               assert 3 == 2
E                +  where 3 = len([{'effect': 'NoSchedule', 'key': 'component', 'operator': 'Equals', 'value': 'pytest'}, {'effect': 'NoSchedule', 'key'...tor': 'Equals', 'value': 'pytest'}, {'effect': 'NoSchedule', 'key': 'global', 'operator': 'Equals', 'value': 'pytest'}])

tests/manifests/test_tolerations.py:85: AssertionError

```

gaelgatelement: OK, strangely on some conditions I don't see this error message. I'm unsure why. But in this case that works.

PR #95, opened 2024-12-16
Check ServiceMonitors

gaelgatelement:
- We check service monitors networking in integration tests
- We check that service monitors can be turned off in manifests tests
- Manifests tests now render templates with ServiceMonitors API enabled

### DIFF
--- a/tests/integration/test_networking.py
+++ b/tests/integration/test_networking.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 
 import asyncio
+import os
 
 import pytest
 from lightkube import AsyncClient
@@ -83,6 +84,9 @@ async def test_services_have_endpoints(
                 assert port.port in port_numbers
 
 
+@pytest.mark.skipif(
+    os.environ.get("SKIP_SERVICE_MONITORS_CRDS", "false") == "true", reason="ServiceMonitors not deployed"
+)
 @pytest.mark.asyncio_cooperative
 @pytest.mark.usefixtures("matrix_stack")
 async def test_pods_monitored(


tests/manifests/__init__.py
```diff
@@ -49,3 +57,22 @@ def _enrich_components_to_test() -> Dict[str, Any]:
...
+    + [
+        values_file
+        for values_file, component in values_files_to_components.items()
+        if len(
+            [
+                sub_component_details.get("has_service_monitor")
+                for sub_component_details in component_details[component]["sub_components"].values()
+            ]
+        )
+        > 0
```
benbz:
```suggestion
        if component_details[component]["has_service_monitor"] or len(
            [
                sub_component_details.get("has_service_monitor")
                for sub_component_details in component_details[component]["sub_components"].values()
            ]
        )
        > 0
```

tests/manifests/__init__.py
```diff
@@ -3,20 +3,25 @@
...
         "additional_values_files": [
             "synapse-worker-example-values.yaml",
         ],
-        "sub_components": [
-            "haproxy",
-            "redis",
-        ],
+        "services_monitors_names": ["{release_name}-synapse"],
+        "sub_components": {
+            "haproxy": {"services_monitors_names": ["{release_name}-synapse-haproxy"]},
```
benbz:
```suggestion
        "sub_components": {
            "haproxy": {},
```

I don't think we should have these. I think the test should assert the name of the `ServiceMonitor` corresponds to the pod controller name (with the expected override for Synapse)

tests/manifests/test_service_monitors.py
```diff
@@ -0,0 +1,58 @@
...
+
+    # Second, we disable subcomponent service monitors independently
+    # And make sure top-level component service monitors are rendered,
+    # as well as independent subcomponent service monitors
+    for turned_off_sub_component in component_details[component]["sub_components"]:
+        # Subcomponents that don't have service monitors don't need to be tested
+        if not component_details[component]["sub_components"][turned_off_sub_component]["has_service_monitor"]:
+            continue
+
+        values = origin_values.copy()
```
benbz: Probably should be a deep copy as the changes are nested?

tests/manifests/__init__.py
gaelgatelement: I could not find a way to determine the name of the pod controller from the component name. I ended up with quite a convoluted test, but hopefully with the comments its logic and goal are clear enough : 

Step 1 : Detection of the service monitors names to find later
I enable each component and subcomponents `ServiceMonitor` templating independently, one by one. For each ServiceMonitor I find, I add to the list of expected service monitors the name of the workloads its selector target.

Step 2 : For each component and subcomponent, we render all but exclude it. We make sure that when we render all other components & subcomponent, its own `ServiceMonitor` is absent. For all ServiceMonitor found, we make sure that it matches the name expected during the detection done in step 1.

benbz: Yeah I'm happy with an override as it is probably only Synapse that will make use of it

PR #100, opened 2024-12-18
HAProxy as a main component chart

gaelgatelement:
In order to avoid deploying a WellKnown pod for serving static files, we are going to implement it as a feature of HAProxy stack.

The HAProxy can now be shared between components. In this PR, only Synapse uses it for its workers routing

### DIFF
--- a/charts/matrix-stack/templates/synapse/synapse_http_service.yaml
+++ b/charts/matrix-stack/templates/synapse/synapse_http_service.yaml
@@ -10,8 +10,8 @@ apiVersion: v1
 kind: Service
 metadata:
   labels:
-    {{- include "element-io.synapse-haproxy.labels" (dict "root" $ "context" $.Values.haproxy) | nindent 4 }}
-  name: {{ $.Release.Name }}-synapse-haproxy
+    {{- include "element-io.synapse-ingress.labels" (dict "root" $ "context" $.Values.haproxy) | nindent 4 }}
+  name: {{ $.Release.Name }}-synapse
   namespace: {{ $.Release.Namespace }}
 spec:
   type: {{ .ingress.service.type | default $.Values.ingress.service.type }}
--- a/charts/matrix-stack/templates/synapse/synapse_ingress.yaml
+++ b/charts/matrix-stack/templates/synapse/synapse_ingress.yaml
@@ -11,8 +11,8 @@ kind: Ingress
 metadata:
 {{- include "element-io.ess-library.ingress.annotations" (dict "root" $ "context" .ingress.annotations) | nindent 2 }}
   labels:
-    {{- include "element-io.synapse.labels" (dict "root" $ "context" $.Values.haproxy) | nindent 4 }}
-  name: {{ $.Release.Name }}-synapse-haproxy
+    {{- include "element-io.synapse-ingress.labels" (dict "root" $ "context" $.Values.haproxy) | nindent 4 }}
+  name: {{ $.Release.Name }}-synapse
   namespace: {{ $.Release.Namespace }}
 spec:
 {{- include "element-io.ess-library.ingress.tlsSecret" (dict "root" $ "context" (dict "hosts" (list (required "synapse.ingress.host is required" .ingress.host)) "tlsSecret" .ingress.tlsSecret)) | nindent 2 }}
@@ -35,7 +35,7 @@ spec:
         pathType: Prefix
         backend:
           service:
-            name: "{{ $.Release.Name }}-synapse-haproxy"
+            name: "{{ $.Release.Name }}-synapse"
             port:
               name: haproxy-403
 {{- end }}
@@ -59,7 +59,7 @@ spec:
         pathType: Prefix
         backend:
           service:
-            name: "{{ $.Release.Name }}-synapse-haproxy"
+            name: "{{ $.Release.Name }}-synapse"
             port:
               name: haproxy-http
 {{- end }}
--- a/charts/matrix-stack/templates/synapse/_helpers.tpl
+++ b/charts/matrix-stack/templates/synapse/_helpers.tpl
@@ -16,13 +16,13 @@ k8s.element.io/synapse-instance: {{ $root.Release.Name }}-synapse
 {{- end }}
 {{- end }}
 
-{{- define "element-io.synapse-haproxy.labels" -}}
+{{- define "element-io.synapse-ingress.labels" -}}
 {{- $root := .root -}}
-{{- with required "element-io.synapse-haproxy.labels missing context" .context -}}
+{{- with required "element-io.synapse-ingress.labels missing context" .context -}}
 {{ include "element-io.ess-library.labels.common" (dict "root" $root "context" .labels) }}
 app.kubernetes.io/component: matrix-stack-ingress
-app.kubernetes.io/name: synapse-haproxy
-app.kubernetes.io/instance: {{ $root.Release.Name }}-synapse-haproxy
+app.kubernetes.io/name: synapse
+app.kubernetes.io/instance: {{ $root.Release.Name }}-synapse
 k8s.element.io/target-name: haproxy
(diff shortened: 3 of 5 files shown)


charts/matrix-stack/templates/synapse/synapse_ingress.yaml
```diff
@@ -11,8 +11,8 @@ kind: Ingress
 metadata:
 {{- include "element-io.ess-library.ingress.annotations" (dict "root" $ "context" .ingress.annotations) | nindent 2 }}
   labels:
-    {{- include "element-io.synapse-haproxy.labels" (dict "root" $ "context" .haproxy) | nindent 4 }}
-  name: {{ $.Release.Name }}-synapse
+    {{- include "element-io.synapse.labels" (dict "root" $ "context" .) | nindent 4 }}
+  name: {{ $.Release.Name }}-synapse-haproxy
```
benbz: Not convinced the ingress changing name is needed or desirable. Any particular reason to do so?

charts/matrix-stack/templates/synapse/synapse_http_service.yaml
benbz, line 1: I think this `Service` should have been moved as a whole to `charts/matrix-stack/templates/haproxy/service.yaml` and the Synapse `Ingress` updated to use the new name.

We can get rid of `element-io.synapse-haproxy.labels` then as well

tests/manifests/__init__.py
```diff
@@ -36,6 +36,7 @@ def _enrich_components_to_test() -> Dict[str, Any]:
 
         _component_details[component].setdefault("has_ingress", True)
         _component_details[component].setdefault("has_service_monitor", True)
+        _component_details[component].setdefault("has_workloads", True)
```
benbz: This is for #101 rather than this PR?

I think I'd want to see a test that confirms there are no workloads when this is `False` as there's suddenly lots of stuff gated on this

tests/manifests/test_ingresses.py
```diff
@@ -28,7 +28,14 @@ async def test_ingress_is_expected_host(component, values, templates):
...
                 assert "host" in rule
-                assert rule["host"] == values[component]["ingress"]["host"]
+                if component == "wellKnownDelegation":
+                    if not values[component].setdefault("ingress", {}).get("host"):
+                        expected_host = values["serverName"]
+                    else:
+                        expected_host = values[component].setdefault("ingress", {})["host"]
+                else:
+                    expected_host = values[component]["ingress"]["host"]
+                assert rule["host"] == expected_host
```
benbz: This is for #101 rather than this PR?

tests/manifests/test_service_monitors.py
```diff
@@ -84,6 +89,8 @@ async def test_service_monitored_as_appropriate(component, values: dict, make_te
             values[component].setdefault(sub_component, {}).setdefault("labels", {}).setdefault(
                 "servicemonitor", "none"
             )
+    for shared_component in component_details[component].get("shared_components", []):
+        values.setdefault(shared_component, {}).setdefault("serviceMonitors", {}).setdefault("enabled", False)
```
benbz, line 93: I think we should be checking that shared_components have ServiceMonitors too

tests/manifests/test_pod_securityContext.py
```diff
@@ -38,9 +38,14 @@ async def test_sets_nonRoot_uids_gids_in_pod_securityContext_by_default(template
 @pytest.mark.parametrize("values_file", values_files_to_test)
 @pytest.mark.asyncio_cooperative
 async def test_can_nuke_pod_securityContext_ids(component, values, make_templates):
-    values[component].setdefault("podSecurityContext", {"runAsUser": None, "runAsGroup": None, "fsGroup": None})
-    for sub_component in component_details[component]["sub_components"]:
-        values[component].setdefault(sub_component, {}).setdefault(
+    if component_details[component]["has_workloads"]:
```
benbz: This looks like it is going to be a very common pattern: We want to do the same "something" with all the component, sub-components and shared components if they have workloads. I feel like we're going to miss bits and so should have some function that does the iterating, allows you to provide a function to do the setting / etc. Bonus points if we can do the `ServiceMonitor` test with it.

Some untested, pseudo-ish code as an example of what I'm thinking
```python
def iterate_component_workload_parts(component, setter):
    if component_details[component]["has_workloads"]:
        setter(values[component], values)
        for sub_component in component_details[component]["sub_components"]
           setter(values[component].setdefault(sub_component, {}), values)
    for shared_component in component_details[component].get("shared_components", []):
```

And then
```python
async def test_can_nuke_pod_securityContext_ids(component, values, make_templates):
  def nuker(part_values, values):
     part_values.setdefault("podSecurityContext", {"runAsUser": None, "runAsGroup": None, "fsGroup": None})

  iterate_component_workload_parts(component, nuker)
  ...
```

The reason for also providing the complete `values` into `setter` is that I think it would be needed for `test_service_monitored_as_appropriate` so that could turn the `ServiceMonitor` on, generate the templates with the complete values and then turn the `ServiceMonitor` back off

tests/manifests/test_service_monitors.py
gaelgatelement: Because the only shared component is haproxy, and because it has ServiceMonitors, I avoided on purpose to add a gate here. I think this can be added when required.

tests/manifests/test_pod_securityContext.py
gaelgatelement: Good idea, I'll implement it in 101 (and remove `has_workloads` gate from 100)

charts/matrix-stack/templates/synapse/synapse_http_service.yaml
gaelgatelement: The problem is that the service is related to the ingress because of `type: {{ .ingress.service.type | default $.Values.ingress.service.type }}`

This is why I ended up keeping it in `synapse` but naming it `synapse-haproxy` to make it obvious that it's a service shared between components

benbz: Right that makes sense. I think I'd still keep it as either `synapse` or `synapse-http` and using Synapse's labels.

gaelgatelement: At first I used Synapse labels indeed, but I encountered a small issue. The problem with using Synapse labels is that the Service Monitor tests starts failing. Synapse `ServiceMonitor` `labelSelector` links up to both Services , Synapse-Headless & Synapse-HaProxy. It detects haproxy workload as being monitored despite its ServiceMonitor being disabled.

benbz: Ah I see. I'd be tempted to add an additional label to `element-io.synapse.process.labels` and then update the `ServiceMonitor` to look at that. Alternatively rename `element-io.synapse-haproxy.labels` to `element-io.synapse-ingress.labels` and remove discussion of HAProxy?

charts/matrix-stack/templates/synapse/synapse_ingress.yaml
gaelgatelement: Just making sure service & ingress have the same name

benbz: I guess as a user of the chart, if I go `kubectl get ingress -n ess` it feels odd that Synapse will be `...-synapse-haproxy` rather than `...-synapse`. The service less so as there's >1 service for Synapse

charts/matrix-stack/templates/synapse/synapse_http_service.yaml
gaelgatelement: I  went for the alternative, as adding a label felt a bit bizarre

Let me know!

charts/matrix-stack/templates/synapse/synapse_ingress.yaml
gaelgatelement: Indeed. Anywa now both are named `-synapse` !

PR #101, opened 2024-12-18
Implement Well Known Delegation

gaelgatelement:
- Well Known is now hosted on the common chart HAProxy

### DIFF
--- a/charts/matrix-stack/templates/well-known/_helpers.tpl
+++ b/charts/matrix-stack/templates/well-known/_helpers.tpl
@@ -11,7 +11,7 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 app.kubernetes.io/component: matrix-delegation
 app.kubernetes.io/name: well-known-delegation
 app.kubernetes.io/instance: {{ $root.Release.Name }}-well-known-delegation
-app.kubernetes.io/version: {{ .image.tag }}
+app.kubernetes.io/version: {{ $root.Release.Version }}
 {{- end }}
 {{- end }}
 
--- a/charts/matrix-stack/templates/well-known/configmap.yaml
+++ b/charts/matrix-stack/templates/well-known/configmap.yaml
@@ -10,7 +10,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   labels:
-    {{- include "element-io.well-known-delegation.labels" (dict "root" $ "context" $.Values.haproxy) | nindent 4 }}
+    {{- include "element-io.well-known-delegation.labels" (dict "root" $ "context" .) | nindent 4 }}
   name: {{ $.Release.Name }}-well-known-haproxy
   namespace: {{ $.Release.Namespace }}
 data:


charts/matrix-stack/templates/haproxy/deployment.yaml
```diff
@@ -70,25 +70,32 @@ spec:
...
-          name: haproxy-http
+          name: haproxy-synapse
           protocol: TCP
         - containerPort: 8009
           name: haproxy-403
           protocol: TCP
+{{- end }}
+{{- if $.Values.wellKnownDelegation.enabled }}
+        - containerPort: 8010
+          name: haproxy-wkd
```
benbz, line 83: Does `haproxy-wellknown` fit or is that too long?

charts/matrix-stack/configs/synapse/partial-haproxy.cfg.tpl
```diff
@@ -0,0 +1,177 @@
...
+  # special synchrotron backend for initialsyncs
+  acl is_sync path -m reg ^/_matrix/client/(r0|v3)/sync$
+  acl is_sync path -m reg ^/_matrix/client/(api/v1|r0|v3)/events$
+  use_backend initial-synchrotron if is_sync { urlp("full_state") -m str true }
+  use_backend initial-synchrotron if is_sync !{ urlp("since") -m found } !{ urlp("from") -m found }
+{{- end }}
+
+  use_backend %[var(req.backend)]
+
+backend main
```
benbz: Shall we prefix all these backends with`synapse-`?

charts/matrix-stack/templates/well-known/_helpers.tpl
```diff
@@ -0,0 +1,66 @@
...
+*/ -}}
+
+{{- define "element-io.well-known-delegation.labels" -}}
+{{- $root := .root -}}
+{{- with required "element-io.well-known-delegation.labels missing context" .context -}}
+{{ include "element-io.ess-library.labels.common" (dict "root" $root "context" .labels) }}
+app.kubernetes.io/component: matrix-delegation
+app.kubernetes.io/name: well-known-delegation
+app.kubernetes.io/instance: {{ $root.Release.Name }}-well-known-delegation
+app.kubernetes.io/version: {{ .image.tag }}
```
benbz: Has no tag, so the chart version?

charts/matrix-stack/templates/well-known/ingress.yaml
```diff
@@ -0,0 +1,39 @@
...
+*/ -}}
+
+{{- with $.Values.wellKnownDelegation -}}
+{{- if .enabled -}}
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+{{- include "element-io.ess-library.ingress.annotations" (dict "root" $ "context" .ingress.annotations) | nindent 2 }}
+  labels:
+    {{- include "element-io.synapse.labels" (dict "root" $ "context" $.Values.haproxy) | nindent 4 }}
```
benbz:
```suggestion
    {{- include "element-io.well-known-delegation.labels" (dict "root" $ "context" .) | nindent 4 }}
```

charts/matrix-stack/templates/well-known/_helpers.tpl
```diff
@@ -0,0 +1,66 @@
...
+{{- define "element-io.well-known-haproxy.labels" -}}
+{{- $root := .root -}}
+{{- with required "element-io.haproxy.labels missing context" .context -}}
+{{ include "element-io.ess-library.labels.common" (dict "root" $root "context" .labels) }}
+app.kubernetes.io/component: matrix-stack-ingress
+app.kubernetes.io/name: well-knownhaproxy
+app.kubernetes.io/instance: {{ $root.Release.Name }}-well-known-haproxy
+app.kubernetes.io/version: {{ .image.tag }}
+{{- end }}
+{{- end }}
```
benbz: Same comment as #100. I think this should go and if needed we add additional labels to distinguish

tests/manifests/test_serviceaccounts.py
```diff
@@ -99,36 +88,39 @@ async def test_uses_serviceaccount_named_as_values_if_specified(component, value
 @pytest.mark.parametrize("values_file", values_files_to_test)
 @pytest.mark.asyncio_cooperative
 async def test_does_not_create_serviceaccounts_if_configured_not_to(component, values, make_templates):
```
benbz, line 90: I guess this test needs lots of updating to use `iterate_component_workload_parts` given the `deepcopy`

charts/matrix-stack/templates/haproxy/deployment.yaml
gaelgatelement: It is 17 characters unfortunately :/

charts/matrix-stack/templates/well-known/_helpers.tpl
gaelgatelement: Service & Pod labels have to match, so I need to target the haproxy pod version ?

out
benbz, line 1: In, out, in, out, shake it all about :p

charts/matrix-stack/templates/well-known/configmap.yaml
```diff
@@ -0,0 +1,24 @@
...
+*/ -}}
+
+{{- with $.Values.wellKnownDelegation -}}
+{{- if .enabled -}}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  labels:
+    {{- include "element-io.well-known-delegation.labels" (dict "root" $ "context" $.Values.haproxy) | nindent 4 }}
+  name: {{ $.Release.Name }}-well-known-haproxy
```
benbz, line 14:
```suggestion
  name: {{ $.Release.Name }}-well-known
```

charts/matrix-stack/templates/well-known/_helpers.tpl
benbz: But this is only used by the `ConfigMap` though?

tests/manifests/test_serviceaccounts.py
gaelgatelement: I'm not even sure it would make the code clearer. I still refactored it a little bit to use an inline function.

charts/matrix-stack/templates/well-known/_helpers.tpl
gaelgatelement: I renamed this like in #100 , `well-known-delegation-ingress.labels` helper.

charts/matrix-stack/templates/well-known/_helpers.tpl
gaelgatelement: Right!

out
gaelgatelement: :laughing:

PR #111, opened 2024-12-19
add IDE files to gitignore

gaelgatelement:
We don't want these files committed accidentally.


gaelgatelement:
> No objection to this but `git config --global core.excludesfile ~/.gitignore_global` is a relatively standard way of handling this by having a local cross-repo ignore file

TIL! We can ignore this then

PR #112, opened 2024-12-19
Fix well knowns

benbz:
* Correct the Element well-known path
* Ensure we have the CORS (and security) headers on responses
* Restart HAProxy with config changes

### DIFF
--- a/charts/matrix-stack/templates/haproxy/deployment.yaml
+++ b/charts/matrix-stack/templates/haproxy/deployment.yaml
@@ -17,6 +17,12 @@ metadata:
   labels:
     {{- include "element-io.haproxy.labels" (dict "root" $ "context" .) | nindent 4 }}
     k8s.element.io/confighash: "{{ include (print $.Template.BasePath "/haproxy/configmap.yaml") $ | sha1sum }}"
+{{- if $.Values.synapse.enabled }}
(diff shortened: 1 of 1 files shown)


charts/matrix-stack/configs/well-known/partial-haproxy.cfg.tpl
```diff
@@ -9,21 +9,26 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
...
   use_backend well-known-client if well-known-client
-  acl well-known-element path_beg /.well-known/matrix/element.json
+
+  acl well-known-element path_beg /.well-known/element/element.json
   use_backend well-known-element if well-known-element
 
 backend well-known-server
   mode http
-  http-request return status 200 content-type "application/json" file "/well-known/server"
+  http-request return errorfile "/well-known/server"
```
gaelgatelement, line 24: OOC, Why  is `errorfile` better here? I hesitated between both approach but the `status 200` file seemed cleaner

benbz: Because I couldn't find a way to split multiple headers over multiple lines in the HAProxy config. I didn't feel one super long line was maintainable

PR #119, opened 2025-01-22
Verify secrets consistency in containers commands & fix synapse POSTGRES_PASSWORD

gaelgatelement:
- This test checks for consistency between the secrets used in containers commands and secrets mounted in the pods.
- It detects properly the error with synapse postgres password value, so we fix it.

### DIFF
--- a/newsfragments/119.fixed.md
+++ b/newsfragments/119.fixed.md
@@ -1,0 +1,1 @@
+Fix an issue where the secret key was wrong when using synapse.postgres.value.


tests/manifests/utils.py
```diff
@@ -82,3 +82,10 @@ def iterate_component_workload_parts(component, values, setter):
             setter(values[component].setdefault(sub_component, {}), values)
     for shared_component in component_details[component].get("shared_components", []):
         setter(values.setdefault(shared_component, {}), values)
+
+def get_or_empty(d, key):
+    res = d.get(key, {})
+    if res is not None:
```
benbz, line 89: Is the only way this could be false is if `d` explicitly had `None` in it for `key`?

gaelgatelement: If d is an empty dict or empty array, we want to return it. We only want to avoid the situation where `None` is explicitely set. Do you think of any other issue?

benbz: No that's fine. I just wanted to understand why we needed this function / `d.get(key, {})` wasn't good enough

gaelgatelement: Problem is that secrets with empty `data:` have `None` instead of `{}`. I can document it.

PR #122, opened 2025-01-22
pytest: add tests of volumes mounts names

gaelgatelement:
- This discover wrongly configured volumes mounts, pointing to a volume not mounted in the workload

### DIFF
--- a/tests/manifests/test_volumes_mounts.py
+++ b/tests/manifests/test_volumes_mounts.py
@@ -1,0 +1,29 @@
+# Copyright 2025 New Vector Ltd
+#
+# SPDX-License-Identifier: LicenseRef-Element-Commercial
+
+import pytest
+
+from . import values_files_to_test
+
+
+@pytest.mark.parametrize("values_file", values_files_to_test)
+@pytest.mark.asyncio_cooperative
+async def test_volumes_mounts_exists(templates):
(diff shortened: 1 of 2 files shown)


tests/manifests/test_volumes_mounts.py
```diff
@@ -0,0 +1,29 @@
...
+    secrets_names = [t["metadata"]["name"] for t in templates if t["kind"] == "Secret"]
+    for template in templates:
+        if template["kind"] in ["Deployment", "StatefulSet"]:
+            volumes_names = []
+            for volume in template["spec"]["template"]["spec"].get("volumes", []):
+                volumes_names.append(volume["name"])
+                if "secret" in volume:
+                    assert volume["secret"]["secretName"] in secrets_names
+                if "configMap" in volume:
+                    assert volume["configMap"]["name"] in configmaps_names
```
benbz, line 23: This only works if the values files don't reference external `Secrets`/`ConfigMaps`, yes?

If we need to do that in future I suppose we can add additionals to the workload to skip specific volumes

gaelgatelement: Indeed. In the future we could add some thing like `external_resources` with a list of secrets / configmaps if we need to pass such a test.

PR #125, opened 2025-01-22
tests: fix syntax error in format strings

gaelgatelement:
I will skip towncrier on this one.


gaelgatelement:
> Actual changes (apart from `tests/manifests/test_secrets_consistency.py`) look good but licenses have gone wrong

And your new test just catches :1st_place_medal:

PR #126, opened 2025-01-23
document the common sections of the values file

gaelgatelement:
This behaviour is significant and should be documented in the README.

### DIFF
--- a/charts/matrix-stack/README.md
+++ b/charts/matrix-stack/README.md
@@ -12,11 +12,11 @@ This Helm chart deploys a coherent Matrix Stack. It currently includes the follo
 
 ## Common
 
-The components deployed in the chart can share some configuration. You'll find below the relevant sections of values.yaml.
+The components deployed in the chart can share some configuration. You'll find below the relevant base sections of values.yaml that you can configure at the top of the chart.
 
 ### Labels
 
-The components deployed in the chart can share labels using the `labels` base section. Any value can be configured here to apply them globally, and will be merged into the per components labels. You can unset a common label on a per-component basis by setting it to `null`.
+The components deployed in the chart can share labels using the `labels` base section. Configure any value here to set global labels, which will then be applied to each component's labels. You can override them on a per component basis. You can unset a common label on a per-component basis by setting it to `null`.
 
 ```yaml
 labels:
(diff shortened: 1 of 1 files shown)


charts/matrix-stack/README.md
```diff
@@ -10,6 +10,65 @@ This Helm chart deploys a coherent Matrix Stack. It currently includes the follo
 * [Synapse](https://github.com/element-hq/synapse) as a Matrix homeserver
 * [Element Web](https://github.com/element-hq/element-web) as a Matrix client
 
+## Common
+
+The components deployed in the chart can share some configuration. You'll find below the relevant sections of values.yaml.
+
+### Labels
+
+The components deployed in the chart can share labels using the `labels` base section. Any value can be configured here to apply them globally, and will be merged into the per components labels.
```
benbz: Do we want to talk about unsetting any global labels on a per-component basis by setting `null`?

charts/matrix-stack/README.md
```diff
@@ -10,6 +10,65 @@ This Helm chart deploys a coherent Matrix Stack. It currently includes the follo
...
+```yaml
+containersSecurityContext:
+  allowPrivilegeEscalation: false
+  capabilities:
+    drop:
+    - ALL
+  readOnlyRootFilesystem: true
+  seccompProfile:
+    type: RuntimeDefault
+```
```
benbz: There is no global `containersSecurityContext`

charts/matrix-stack/README.md
```diff
@@ -10,6 +10,65 @@ This Helm chart deploys a coherent Matrix Stack. It currently includes the follo
...
+
+### Tolerations and Topology Spread Constraints configuration
+
+Workloads of the individual components in the chart can share the same configuration using the `tolerations` and `topologySpreadConstraints` base section. Any value can be configured here to apply them globally, but can be overridden on a per component basis.
```
benbz: * `tolerations` are appended to by individual components rather than overwriting, `topologySpreadConstraints` are indeed overwritten
* `topologySpreadConstraints` automatically adds an appropriate `labelSelector.matchLabels` (based on the `app.kubernetes.io/instance` label) to each entry if one isn't specified. This should be highlighted
* `topologySpreadConstraints` automatically adds an appropriate `matchLabelKeys` (based on `pod-template-hash` for `Deployments` to each entry if one isn't specified. This should also be highlighted

charts/matrix-stack/README.md
```diff
@@ -10,6 +10,65 @@ This Helm chart deploys a coherent Matrix Stack. It currently includes the follo
...
+
+```yaml
+labels:
+  my-deployment: ess-deployment
+```
+
+### Ingress Configuration
+
+Ingresses of the individual components in the chart can share the same configuration using the `ingress` base section.
+Any `annotations`, `className`, `tlsSecret` can be configured here to apply them globally, but can be overridden on a per component basis.
```
benbz: Talk about unsetting any global annotations on a per-component basis by setting `null`?

charts/matrix-stack/README.md
gaelgatelement: Uh, I must have hallucinate this!

benbz: Heh, I didn't know you were a LLM :p

PR #131, opened 2025-01-24
Render config with secrets : Build a dedicated image

gaelgatelement:
- This image will use golang, and might be use for any feature we need (generating keys, ulid, merging config, etc)
- The image is using pure go, so that we don't need to inject any bash interpreter and other external binaries to the image to make the mechanism work
- It's going to be built and released on the custom tag `matrix-tools-*`.
- The helm release now only happens on semver tags

### DIFF
--- a/charts/matrix-stack/source/common/sub_schema_values.yaml.j2
+++ b/charts/matrix-stack/source/common/sub_schema_values.yaml.j2
@@ -13,7 +13,7 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 
 ## The matrix-tools iamge, used in multiple components
 matrixTools:
-{{ image(registry="ghcr.io", repository="element-hq/matrix-tools", tag="0.0.2") | indent(2) }}
+{{ image(registry="ghcr.io", repository="element-hq/matrix-tools", tag="0.1.0") | indent(2) }}
 
 ## The server name of the Matrix Stack. This gets embedded in user IDs & room IDs
 ## It can not change after the initial deployment.
--- a/charts/matrix-stack/values.yaml
+++ b/charts/matrix-stack/values.yaml
@@ -20,7 +20,7 @@ matrixTools:
 
     ## The tag of the container image to use.
     ## Defaults to the Chart's appVersion if not set
-    tag: 0.0.2
+    tag: 0.1.0
 
     ## Container digest to use. Used to pull the image instead of the image tag / Chart appVersion if set
     # digest:


matrix-tools/internal/pkg/args/args.go
```diff
@@ -0,0 +1,53 @@
...
+type Options struct {
+	Files   []string
+	Output  string
+	Address string
+}
+
+func ParseArgs(args []string) (*Options, error) {
+	var options Options
+
+	if len(args) < 2 || args[1] != "--render-config" {
```
kegsay: Just use [flag](https://pkg.go.dev/flag) from the stdlib. Example usage: https://github.com/element-hq/chaos/blob/main/cmd/chaos/main.go#L13

matrix-tools/internal/pkg/renderer/renderer.go
```diff
@@ -0,0 +1,120 @@
...
+		fileContent, err := os.ReadFile(sourceFilename)
+		if err != nil {
+			return nil, errors.New("failed to read file: " + sourceFilename)
+		}
+
+		envVarNames := extractEnvVarNames(string(fileContent))
+		for _, envVar := range envVarNames {
+			if val, ok := os.LookupEnv(envVar); !ok {
+				return nil, errors.New(envVar + " is not present in the environment")
+			} else {
```
kegsay: The `return` makes this else block nothing but needless indentation. I'd un-inline the `LookupEnv` so you can save the horizontal space.

matrix-tools/cmd/main.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+	result, err := renderer.RenderConfig(options.Files)
+	if err != nil {
+		fmt.Println("Error:", err)
+		os.Exit(1)
+	}
+
+	if options.Address != "" {
+		tcpwait.WaitForTCP(options.Address)
+	}
```
benbz, line 28: I feel this should be a separate mode for this binary or a different binary in the same image rather than bundling into rendering config

matrix-tools/internal/pkg/args/args.go
```diff
@@ -0,0 +1,53 @@
...
+		fmt.Println("Usage: go run main.go --render-config <file1> [<file2>, <...>] --output <file> [--tcpwait <server> <port>]")
+		os.Exit(1)
+	}
+
+	for i := 1; i < len(args); i++ {
```
benbz: Is there no idiomatic Go arg parsing package?

.github/workflows/matrix-tools.yml
```diff
@@ -0,0 +1,101 @@
...
+          tags: |
+            type=raw,value=${{ env.VERSION }},enable=${{ github.ref_type == 'tag' }}
+            type=sha,prefix=sha-,format=short
+
+      - name: Build and Push Image
+        id: build_image
+        uses: docker/bake-action@v4.1.0
+        with:
+          # Only for main branch or releases
+          push: "${{ github.ref_type == 'tag' }}"
```
benbz, line 89: Comment and this line disagree

charts/matrix-stack/ci/pytest-synapse-values.yaml
```diff
@@ -4,6 +4,10 @@
 
 serverName: ess.localhost
 
+matrixTools:
+  image:
+    pullPolicy: Never
```
benbz: Can this go once we've got the first tagged image?

matrix-tools/internal/pkg/renderer/renderer.go
```diff
@@ -0,0 +1,120 @@
...
+					replacementValue, err = json.Marshal(val)
+				}
+				if err != nil {
+					return nil, err
+				}
+				fileContent = bytes.ReplaceAll(fileContent, []byte("${"+envVar+"}"), replacementValue)
+			}
+		}
+
+		var data map[string]interface{}
```
kegsay: For post-generics Go, `map[string]any` is more consise and literally the same thing.

matrix-tools/internal/pkg/tcpwait/tcpwait.go
```diff
@@ -0,0 +1,23 @@
...
+	"time"
+)
+
+// WaitForTCP waits for a TCP connection on the specified address.
+func WaitForTCP(address string) {
+	for {
+		fmt.Println("Waiting for TCP connection on " + address)
+		conn, err := net.DialTimeout("tcp", address, 5*time.Second)
+		if err != nil {
+			time.Sleep(time.Second)
```
kegsay, line 19: This will loop forever if the addr never dials. Typically you would make the caller pass in a `context.Context` and then periodically check to see if that has been cancelled and then give up if it has. This allows the caller to stop waiting.

matrix-tools/internal/pkg/renderer/renderer_test.go
```diff
@@ -0,0 +1,112 @@
...
+		expected map[string]interface{}
+		err      bool
+	}{
+		{
+			name:  "Single File",
+			files: []string{"testdata/single_file.yml"},
+			env: map[string]string{
+				"TEST_ENV":      "value",
+				"SECRET_KEY":    "secret://testdata/secret_key",
+				"THIS_HOSTNAME": "hostname://",
```
benbz: Can we have a test with a similar setup to Synapse (where it removes `-synapse-` from the hostname)?

matrix-tools/internal/pkg/renderer/renderer_test.go
```diff
@@ -0,0 +1,112 @@
...
+				t.Errorf("expected: %v, got: %v", tc.expected, result)
+			}
+			for k := range tc.env {
+				os.Unsetenv(k)
+			}
+		})
+	}
+}
+
+func deepEqual(a, b interface{}) bool {
```
kegsay: For testing you can also use `reflect.DeepEqual(a,b)`

matrix-tools/internal/pkg/renderer/renderer_test.go
```diff
@@ -0,0 +1,112 @@
...
+			expected: map[string]interface{}{
+				"key":       "value",
+				"secretKey": "secret_value",
+				"hostname":  hostname,
+			},
+			err: false,
+		},
+		{
+			name:  "Multiple Files",
+			files: []string{"testdata/multiple_files_1.yml", "testdata/multiple_files_2.yml"},
```
benbz: These files should probably have some clashing data so that the test includes what "wins"

matrix-tools/cmd/main.go
```diff
@@ -0,0 +1,40 @@
...
+	"os"
+
+	"github.com/element-hq/ess-helm/matrix-tools/internal/pkg/args"
+	"github.com/element-hq/ess-helm/matrix-tools/internal/pkg/renderer"
+	"github.com/element-hq/ess-helm/matrix-tools/internal/pkg/tcpwait"
+	"gopkg.in/yaml.v3"
+)
+
+func main() {
+	options, err := args.ParseArgs(os.Args)
```
kegsay, line 20: Are these trusted inputs? If not, then some of this code needs to change to make it handle malicious inputs.

charts/matrix-stack/templates/synapse/synapse_statefulset.yaml
```diff
@@ -79,43 +79,40 @@ We have an init container to render & merge the config for several reasons:
...
+        - "{{ (tpl .postgres.host $) }}:{{ .postgres.port }}"
         env:
+          - name: SYNAPSE_MACAROON
+            value: secret:///secrets/{{ tpl (.macaroon.secret | default (printf "%s-synapse" $.Release.Name)) $ }}/{{ .macaroon.secretKey | default "MACAROON" }}
+          - name: SYNAPSE_POSTGRES_PASSWORD
+            value: secret://secrets/{{ tpl (.postgres.password.secret | default (printf "%s-synapse" $.Release.Name)) $ }}/{{ .postgres.password.secretKey | default "POSTGRES_PASSWORD" }}
+          - name: SYNAPSE_REGISTRATION_SHARED_SECRET
+            value: secret:///secrets/{{ tpl (.registrationSharedSecret.secret | default (printf "%s-synapse" $.Release.Name)) $ }}/{{ .registrationSharedSecret.secretKey | default "REGISTRATION_SHARED_SECRET" }}
+          - name: APPLICATION_NAME
+            value: hostname://-synapse-
```
benbz: The nature of these env vars has changed. Previously they contained the actual values (Secrets or Application name). Now they are either paths to the Secrets or a pattern to replace. I feel that the env var names should reflect this.

An alternative would be t suppress `CKV_K8S_35` and use `valueFrom` for most of these. `APPLICATION_NAME` would still need to be handled differently

matrix-tools/internal/pkg/renderer/renderer.go
```diff
@@ -0,0 +1,120 @@
...
+		if err := yaml.Unmarshal(fileContent, &data); err != nil {
+			return nil, errors.New("Post-processed YAML of " + sourceFilename + " is invalid: " + string(fileContent) + " with error: " + err.Error())
+		}
+
+		if err := deepMergeMaps(data, output); err != nil {
```
kegsay, line 121: It's not really clear _why_ you want to do this. The source files will be sharing a global namespace which feels Bad?

gaelgatelement: The idea is to be able to inject multiple configuration files in our containers, dynamically, and merge them down to one single file.

This is for example used [in ti-messenger-chart](https://github.com/element-hq/ti-messenger-chart/blob/main/charts/tim-backend/templates/synapse-ti-messenger-specs-secret.yaml) to inject a config fragment dedicated to TI-M specs. The TI-M backend comes with a secret hardcoded in the chart that is injected in addition to the default configuration of ess-helm.

matrix-tools/internal/pkg/tcpwait/tcpwait.go
gaelgatelement: This is expected, as kubernetes will kill the init container process and crashes it if it never exits (after 5 min by default)

charts/matrix-stack/templates/synapse/synapse_statefulset.yaml
gaelgatelement: > The nature of these env vars has changed. Previously they contained the actual values (Secrets or Application name). Now they are either paths to the Secrets or a pattern to replace. I feel that the env var names should reflect this.
In the config files, these variables names describe what you expect to have here : 

`macaroon: ${SYNAPSE_MACAROON}`.

How they are injected here depends on `matrix-tools` behaviour, which uses special variable parsing to do its job.  But it could be another tool (like with our previous bash/python script). I think I prefer to keep current naming.

matrix-tools/internal/pkg/renderer/renderer.go
gaelgatelement: Without the else, go returns the error `val is unused` in line 67, and `undefined: val` line 71

matrix-tools/cmd/main.go
gaelgatelement: With `flag` package it should now be fine ?

matrix-tools/internal/pkg/renderer/testdata/000_override.yml
benbz, line 1: Can this file also have 1 non-overridden key-value so that the tests show things are merged not just replaced

charts/matrix-stack/templates/synapse/synapse_statefulset.yaml
```diff
@@ -79,43 +79,38 @@ We have an init container to render & merge the config for several reasons:
...
-        - "-ce"
-        - >
+        - "/matrix-tools"
           {{- /* The filesystem structure is `/secrets`/<< secret name>>/<< secret key >>.
                  The non-defaulted values are handling the case where the credential is provided by an existing Secret
                  The default values are handling the case where the credential is provided plain in the Helm chart and we add it to our Secret with a well-known key.
 
                  These could be done as env vars with valueFrom.secretKeyRef, but that triggers CKV_K8S_35.
                  If we wanted to allow for replacement of values in the user-provided config with external Secrets we could extend this init-container
                  to allow extraEnv with valueFrom.secretKeyRef. */}}
```
benbz: This should move down to the `env` for the `initContainer` and possibly be reworded a bit

charts/matrix-stack/templates/synapse/synapse_statefulset.yaml
benbz: Ack. I suppose people could provide secret values in env (either `value` or `valueFrom`) and then still use the `${FOO}` syntax here as the replacement can be a plain value and not just `secret://` or `hostname://`

charts/matrix-stack/templates/synapse/synapse_statefulset.yaml
```diff
@@ -79,43 +79,40 @@ We have an init container to render & merge the config for several reasons:
...
+        - /config-templates/05-{{ $processType }}.yaml
         env:
+          {{- /* The filesystem structure is `/secrets`/<< secret name>>/<< secret key >>.
+                 The non-defaulted values are handling the case where the credential is provided by an existing Secret
+                 The default values are handling the case where the credential is provided plain in the Helm chart and we add it to our Secret with a well-known key.
+
+                 These could be done as env vars with valueFrom.secretKeyRef, but that triggers CKV_K8S_35.
+                 The prefixes `secret://` and `hostname://` are mechanisms handled in the matrix-tools binary:
+                 - `secret://` is used to read the secret from the filesystem and inject in the rendered config file
+                 - `hostname://` is used to drop the pattern indicated from the hostname and inject resulting value in the config file
```
benbz:
```suggestion
                 These could be done as env vars with valueFrom.secretKeyRef, but that triggers CKV_K8S_35.
                 The prefixes `secret://` and `hostname://` are mechanisms handled in the matrix-tools binary:
                 - `secret://` is used to read the secret from the filesystem and inject in the rendered config file
                 - `hostname://` is used to drop the pattern indicated from the hostname and inject resulting value in the config file
                 - Anything else is handled as a simple string replacement
```

matrix-tools/internal/pkg/renderer/renderer.go
kegsay:
```go
val, ok := os.LookupEnv(envVar)
if !ok {
    return nil, fmt.Errorf("%s is not present in the environment", envVar)
}
// .. else block
```
You should be using `val`.

matrix-tools/cmd/main.go
kegsay: No, in the general case are the inputs trusted? There's more than just arg parsing to consider.

matrix-tools/internal/pkg/renderer/renderer.go
kegsay: If you have the same keys in the map then it'll be very sensitive to the order each file is read (last one wins). This should at the very least be documented/tested.

gaelgatelement: I've added a test for this yes (`"Overrides in order")

matrix-tools/cmd/main.go
gaelgatelement: No, in the general case are the inputs trusted? 

I mean, admins of the chart can inject any input they want here. Can you describe what problem do you envision ?

matrix-tools/cmd/main.go
```diff
@@ -0,0 +1,49 @@
...
+	if err != nil {
+		fmt.Println(err)
+		return
```
kegsay: Probably `os.Exit(1)`?

matrix-tools/cmd/main.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	if options.Address != "" {
+		tcpwait.WaitForTCP(options.Address)
+	}
+
+	if options.Output != "" && len(options.Files) > 0 {
+		outputYAML, _ := yaml.Marshal(result)
+		fmt.Printf("Rendering config to file: %v", options.Output)
+		err = os.WriteFile(options.Output, outputYAML, 0644)
```
kegsay, line 56: It would be more unixy to just splat to stdout and let the caller pipe it to disk? Maybe if there is no `output.Output` specified?

matrix-tools/internal/pkg/args/args.go
```diff
@@ -0,0 +1,52 @@
...
+	var options Options
+
+	renderConfigSet := flag.NewFlagSet("render-config", flag.ExitOnError)
+	output := renderConfigSet.String("output", "", "Output file for rendering")
+
+	tcpWaitSet := flag.NewFlagSet("tcpwait", flag.ExitOnError)
+	tcpWait := tcpWaitSet.String("address", "", "Address to listen on for TCP connections")
+
+	if args[1] == "render-config" {
+		err := renderConfigSet.Parse(args[2:])
```
kegsay, line 28: This is an unusual way of using the `flag` package.

The flag package keeps global internal state to store parsed results in pointers provided before parsing, which allows you to do things like:
```go
outputPointer := flag.String("output", "", "path to the output YAML")
// outputPointer is nil until you parse the args
flag.Parse() // implicitly parses os.Args
fmt.Println(*outputPointer) // dereference as it's a pointer
```
Realistically this style should provide all you need.

matrix-tools/internal/pkg/renderer/renderer.go
```diff
@@ -0,0 +1,120 @@
...
+				destination[key] = value
+			}
+		} else {
+			destination[key] = value
+		}
+	}
+	return nil
+}
+
+func RenderConfig(sourceFiles []string) (map[string]any, error) {
```
kegsay: Really needs documentation.

I also don't understand why you wouldn't use `text/template`.

I would also have a slight preference for `sourceFiles []string` to be `sourceConfigs []io.Reader`. Currently, `RenderConfig` is doing more than it should because it is:
 - absolute pathing the source file then reading it,
 - actually doing the replacement logic.

If you made it `io.Reader` then you could provide in-memory configs (useful for testing) and allow other sources of config input (e.g stdin). YMMV though, go with what makes sense for your purposes.

matrix-tools/internal/pkg/renderer/renderer.go
```diff
@@ -0,0 +1,120 @@
...
+			sourceFilename = absName
+		}
+
+		if fileInfo, err := os.Stat(sourceFilename); err == nil && fileInfo.IsDir() {
+			continue
+		}
+
+		fileContent, err := os.ReadFile(sourceFilename)
+		if err != nil {
+			return nil, errors.New("failed to read file: " + sourceFilename)
```
kegsay: `fmt.Errorf()` is more ergonomic than `errors.New` as you can use `%s`.

matrix-tools/internal/pkg/args/args.go
gaelgatelement: I could not figure out how to unit tests the flags behaviour, because `flag.Parse()` does not accepts a flag list. This is why I ended up doing it like this.

matrix-tools/cmd/main.go
gaelgatelement: I'm building a binary which does not need any shell, so that the container can be as minimal as possible.

matrix-tools/internal/pkg/renderer/renderer.go
gaelgatelement: OK this should be better now. We are now using `text/template` to parse environment variable content instead of our customs `secrets://` and `hostname://` mechanisms.

The function has been documented as such.

Render Config is also using a read list now. We have a defer function called to be sure the files are closed properly.

matrix-tools/internal/pkg/renderer/renderer.go
```diff
@@ -0,0 +1,139 @@
...
+		envVarNames := extractEnvVarNames(string(fileContent))
+		for _, envVar := range envVarNames {
+			val, ok := os.LookupEnv(envVar)
+			if !ok {
+				return nil, errors.New(envVar + " is not present in the environment")
+			}
+			var replacementValue []byte
+			tmpl, err := template.New("matrix-tools").Funcs(funcMap).Parse(val)
+			if err != nil {
+				return nil, err
```
kegsay: Go doesn't provide stack traces on errors. This means you need to provide context when you return errors. `err` won't know anything about the env var being used for example, it only knows the `val`, seeing something like "failed to parse template: unmatched }" lacks important context (which env var). You should include this here by nesting errors:
```go
return nil, fmt.Errorf("failed to parse template for env var %s: %w", envVar, err)`
```
Same applies everywhere you just return `nil, err`. It's a pain to do, but you'll thank me later.

PR #137, opened 2025-01-29
matrix-tools: add quote filter

gaelgatelement:
We need to be able to quote random values to handle special chars.

### DIFF
--- a/matrix-tools/internal/pkg/renderer/renderer_test.go
+++ b/matrix-tools/internal/pkg/renderer/renderer_test.go
@@ -28,18 +28,21 @@ func TestRenderConfig(t *testing.T) {
 			name: "Single File",
 			readers: []io.Reader{
 				bytes.NewBuffer([]byte(`key: ${TEST_ENV}
+quotedValue: ${SPECIAL_CHARS}
 secretKey: ${SECRET_KEY}
 hostname: ${THIS_HOSTNAME}`)),
 			},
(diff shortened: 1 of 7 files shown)


matrix-tools/internal/pkg/renderer/renderer_test.go
```diff
@@ -28,18 +28,21 @@ func TestRenderConfig(t *testing.T) {
...
 				"SECRET_KEY":    "{{ readfile \"testdata/secret_key\" }}",
 				"THIS_HOSTNAME": fmt.Sprintf("{{ hostname | replace \"%s\" \"\" }}", droppedFromHostname),
 			},
 			expected: map[string]any{
-				"key":       "value",
-				"secretKey": "secret_value",
-				"hostname":  hostname[0:2] + hostname[4:],
+				"key":         "value",
+				"secretKey":   "secret_value",
+				"quotedValue": "!][",
```
benbz: I'm a bit confused. What would this give without the `quote` filter?

gaelgatelement: Because `!` is a yaml character, it gets interpreted : 

`        renderer_test.go:130: expected: map[hostname:gag-laptop key:value quotedValue:!][ secretKey:secret_value], got: map[hostname:gag-laptop key:value quotedValue: secretKey:secret_value]`

PR #141, opened 2025-01-30
fix project name to ESS Community Helm Chart

gaelgatelement:
As agreed with Marketing.


gaelgatelement:
> What's here LGTM but there's a couple of places near the top of the readme that should also be adjusted I think

Mmmh I changed the title in the main README. Do you see anything else?

PR #142, opened 2025-01-30
init-secrets job 

gaelgatelement:
This job is dedicated to initializing secrets when they are missing from the chart.

The job is enabled by default. It makes some secrets optional as they are going to be handled automatically during initial chart setup.

So far this PR makes the following secret optional : 
- `synapse.macaroon`

The PR adds 2 new helpers to include secrets in `matrix-tools` init containers.

### DIFF
--- a/charts/matrix-stack/ci/pytest-element-web-values.yaml
+++ b/charts/matrix-stack/ci/pytest-element-web-values.yaml
@@ -4,6 +4,11 @@
 
 serverName: ess.localhost
 
+matrixTools:
+  image:
+    pullPolicy: Never
+    digest: ""
+
 elementWeb:
   ingress:
     host: element.{{ $.Values.serverName }}
--- a/tests/manifests/__init__.py
+++ b/tests/manifests/__init__.py
@@ -5,12 +5,12 @@
 import copy
 from typing import Any, Dict
 
+_raw_shared_components_details = {
+    "initSecrets": {"hyphened_name": "init-secrets", "has_service_monitor": False, "has_ingress": False},
+    "haproxy": {},
+}
+
 _raw_component_details = {
-    "initSecrets": {
-        "hyphened_name": "init-secrets",
-        "has_service_monitor": False,
-        "has_ingress": False,
-    },
     "elementWeb": {
         "hyphened_name": "element-web",
         "has_service_monitor": False,
@@ -24,7 +24,7 @@ _raw_component_details = {
                 "has_service_monitor": False,
             },
         },
-        "shared_components": ["haproxy"],
+        "shared_components": ["initSecrets", "haproxy"],
     },
     "wellKnownDelegation": {
         "hyphened_name": "well-known",
@@ -35,9 +35,9 @@ _raw_component_details = {
 }
 
 
-def _enrich_components_to_test() -> Dict[str, Any]:
-    _component_details = copy.deepcopy(_raw_component_details)
-    for component in _raw_component_details:
+def _enrich_components_to_test(details) -> Dict[str, Any]:
+    _component_details = copy.deepcopy(details)
+    for component in details:
         _component_details[component].setdefault("hyphened_name", component)
 
         values_files = _component_details[component].setdefault("additional_values_files", [])
@@ -54,7 +54,8 @@ def _enrich_components_to_test() -> Dict[str, Any]:
     return _component_details
 
 
-component_details = _enrich_components_to_test()
+component_details = _enrich_components_to_test(_raw_component_details)
+shared_components_details = _enrich_components_to_test(_raw_shared_components_details)
 
 values_files_to_components = {
     values_file: component
--- a/tests/manifests/test_basic.py
+++ b/tests/manifests/test_basic.py
@@ -16,6 +16,7 @@ async def test_nothing_enabled_renders_nothing(templates):
 @pytest.mark.parametrize("values_file", values_files_to_test)
 @pytest.mark.asyncio_cooperative
 async def test_values_file_renders_only_itself(component, templates):
+    # init-secrets does not render any manifest without any component needing it
     assert len(templates) > 0
 
     allowed_starts_with = [
--- a/tests/manifests/test_manifest_test_infrastructure.py
+++ b/tests/manifests/test_manifest_test_infrastructure.py
@@ -6,7 +6,7 @@ from pathlib import Path
 
 import pytest
 
-from . import component_details, values_files_to_test
+from . import component_details, shared_components_details, values_files_to_test
 
 
 def test_all_components_covered():
@@ -16,7 +16,9 @@ def test_all_components_covered():
     for contents in templates_folder.iterdir():
         if not contents.is_dir():
             continue
-        if contents.name in ("haproxy", "ess-library"):
+        if contents.name in ("ess-library",) + tuple(
+            k if not v["hyphened_name"] else v["hyphened_name"] for k, v in shared_components_details.items()
+        ):
             continue
 
         assert contents.name in expected_folders
--- a/tests/manifests/test_service_monitors.py
+++ b/tests/manifests/test_service_monitors.py
@@ -6,7 +6,7 @@ from typing import Any, Dict, Iterator
 
 import pytest
 
-from . import component_details, values_files_to_test
+from . import component_details, shared_components_details, values_files_to_test
 
 
 def selector_match(labels: Dict[str, str], selector: Dict[str, str]) -> bool:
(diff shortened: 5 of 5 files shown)


charts/matrix-stack/configs/matrix-authentication-service/config.yaml.tpl
benbz, line 1: Should be part of this MR?

charts/matrix-stack/templates/init-secrets/role.yaml
```diff
@@ -0,0 +1,20 @@
+{{- /*
+Copyright 2024 New Vector Ltd
+
+SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+*/ -}}
+{{- with .Values.initSecrets -}}
+{{- if .enabled -}}
+apiVersion: rbac.authorization.k8s.io/v1
```
benbz, line 8: https://helm.sh/docs/chart_best_practices/rbac/ should be togglable behind a `initSecrets.rbac.create` flag (that defaults to `true`)

charts/matrix-stack/templates/init-secrets/rolebinding.yaml
```diff
@@ -0,0 +1,25 @@
+{{- /*
+Copyright 2024 New Vector Ltd
+
+SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+*/ -}}
+
+{{- with .Values.initSecrets -}}
+{{- if .enabled -}}
+apiVersion: rbac.authorization.k8s.io/v1
```
benbz, line 9: https://helm.sh/docs/chart_best_practices/rbac/ should be togglable behind a `initSecrets.rbac.create` flag (that defaults to `true`)

matrix-tools/internal/pkg/secret/secret.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+func GenerateSecret(client kubernetes.Interface, namespace string, name string, key string, secretType args.SecretType, override bool) error {
+	ctx := context.Background()
+
+	secretsClient := client.CoreV1().Secrets(namespace)
+	// Fetch the existing secret or initialize an empty one
+	existingSecret, err := secretsClient.Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		existingSecret, err = secretsClient.Create(ctx, &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: namespace}, Data: nil}, metav1.CreateOptions{})
```
benbz: Suspect we need args to label this secret

charts/matrix-stack/templates/synapse/synapse_secret.yaml
```diff
@@ -22,7 +22,6 @@ data:
 {{- end }}
 {{- end }}
 {{- end }}
-{{- include "element-io.ess-library.check-credential" (dict "root" $ "context" (dict "secretPath" "synapse.macaroon" "secretProperty" .macaroon)) -}}
```
benbz, line 25: Should the check-credential helper be updated to carry on doing its current job but also look at `root.initSecrets.enabled`?

charts/matrix-stack/templates/init-secrets/job.yaml
```diff
@@ -0,0 +1,67 @@
...
+Copyright 2024 New Vector Ltd
+
+SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+*/ -}}
+
+{{- with .Values.initSecrets -}}
+{{- if .enabled -}}
+apiVersion: batch/v1
+kind: Job
+metadata:
```
benbz, line 11: Needs the annotations to act as a Helm hook IIUC

charts/matrix-stack/templates/init-secrets/job.yaml
```diff
@@ -0,0 +1,67 @@
...
+  manualSelector: false
+  parallelism: 1
+  podReplacementPolicy: TerminatingOrFailed
+  template:
+{{- with .annotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+{{- end }}
+    spec:
+      automountServiceAccountToken: false
```
benbz: Won't we need to mount the SA token to interact with k8s?

charts/matrix-stack/templates/init-secrets/job.yaml
```diff
@@ -0,0 +1,67 @@
...
+{{- end }}
+{{- with .nodeSelector }}
+      nodeSelector:
+        {{- toYaml . | nindent 8 }}
+{{- end }}
+      {{ include "element-io.ess-library.pods.tolerations" (dict "root" $ "context" .tolerations) | nindent 6 }}
+{{- with .hostAliases }}
+      hostAliases:
+        {{- tpl (toYaml . | nindent 8) $ }}
+{{- end }}
```
benbz: Not needed as won't talk to anything other than k8s?

charts/matrix-stack/templates/init-secrets/job.yaml
gaelgatelement: What will the annotation bring ? It seems to work fine as-is :thinking:

benbz: My understanding is that at present you've got a 1-shot `Job`. This won't be run on subsequent `helm upgrade/install`. The `Job` also stays around in cluster and I suspect we want it cleaned up before the next Job run (`"helm.sh/hook-delete-policy: before-hook-creation` annotation?)

charts/matrix-stack/templates/synapse/_synapse_details.tpl
```diff
@@ -111,6 +111,9 @@ responsibleForMedia
 {{- $root := .root -}}
 {{- with required "element-io.synapse.configSecrets missing context" .context -}}
 {{ $configSecrets := list (printf "%s-synapse" $root.Release.Name) }}
+{{- if $root.Values.initSecrets.enabled }}
+{{ $configSecrets = append $configSecrets (printf "%s-init-secrets" $root.Release.Name) }}
```
benbz:
```suggestion
{{ $configSecrets = append $configSecrets (printf "%s-generated" $root.Release.Name) }}
```

Might make it clearer where these `Secrets` come from?

charts/matrix-stack/templates/synapse/synapse_secret.yaml
```diff
@@ -22,7 +22,7 @@ data:
...
+{{- include "element-io.ess-library.check-credential" (dict "root" $ "context" (dict "secretPath" "synapse.postgres.password" "secretProperty" .postgres.password "initIfAbsent" true)) -}}
```
benbz:
```suggestion
{{- include "element-io.ess-library.check-credential" (dict "root" $ "context" (dict "secretPath" "synapse.macaroon" "secretProperty" .macaroon "initIfAbsent" true)) -}}
```

We're not init'ing Postgres yet and we otherwise want to check the Macaroon. IIUC

charts/matrix-stack/ci/init-secrets-checkov-values.yaml
```diff
@@ -0,0 +1,19 @@
+# Copyright 2024 New Vector Ltd
+#
+# SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+
+initSecrets:
+  enabled: true
+  annotations:
+    checkov.io/skip1: CKV_K8S_11=We deliberately don't set CPU limits. Pod is BestEffort not Guaranteed
+    checkov.io/skip2: CKV_K8S_43=No digests
```
benbz, line 9: We should have digests

charts/matrix-stack/templates/init-secrets/job.yaml
```diff
@@ -0,0 +1,79 @@
...
+  backoffLimit: 6
+  completionMode: NonIndexed
+  completions: 1
+  manualSelector: false
+  parallelism: 1
+  podReplacementPolicy: TerminatingOrFailed
+  template:
+    annotations:
+      "helm.sh/hook": pre-install,pre-upgrade
+      "helm.sh/hook-delete-policy": before-hook-creation
```
benbz: As I understand it, the annotations need to be on the `Job` not on the created `Pod`

charts/matrix-stack/ci/init-secrets-checkov-values.yaml
gaelgatelement: Not on FOSS though?

.github/workflows/matrix-tools.yml
```diff
@@ -5,7 +5,7 @@
 name: Build Docker Images
 
 on:
-  pull_request_target:
+  pull_request:
```
benbz: To be reverted before merging

charts/matrix-stack/ci/pytest-element-web-values.yaml
```diff
@@ -4,6 +4,11 @@
 
 serverName: ess.localhost
 
+matrixTools:
+  image:
+    pullPolicy: Never
+    digest: ""
+
```
benbz, line 11: Why is this needed in the EW integration test values?

PR #145, opened 2025-02-04
Dependency update

benbz:
`poetry up --latest --pinned --with=dev` mainly to bring in latest Jinja to shut dependabot up but also generally bump


benbz:
The updates brought in a new `ruff` version which required formatting and other fixes

PR #147, opened 2025-02-06
ci: add gotestfmt

gaelgatelement:
It's a classic go tool to format test output properly in diverse CI contexts. 

https://github.com/GoTestTools/gotestfmt


newsfragments/133.internal.md
benbz, line 1: 147.internal.md

PR #148, opened 2025-02-06
pytest: build matrix-tools in CI runs

gaelgatelement:
At the moment we rely on manually tagging matrix-tools before the CI tests are able to run against it. It also involves us manually loading `matrix-tools` image in the kind cluster.

A future optimization might be to pre-build `matrix-tools` but we will have to pass it around the tests. Let's try this for now.

### DIFF
--- a/tests/integration/fixtures/helm.py
+++ b/tests/integration/fixtures/helm.py
@@ -90,8 +90,12 @@ retention:
 
 @pytest.fixture(autouse=True, scope="session")
 async def matrix_stack(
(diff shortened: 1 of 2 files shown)


tests/integration/fixtures/__init__.py
benbz, line 1: Are the changes in here needed given there's no usage outside of this package?

gaelgatelement: It is for pytest to discover the new fixtures when loading the python modules.

PR #155, opened 2025-02-07
fix imagePullSecrets missing for matrixTools containers

gaelgatelement:
This tests makes sure that all required pull secrets are properly available in the component manifests.

### DIFF
--- a/charts/matrix-stack/ci/matrix-authentication-service-minimal-values.yaml
+++ b/charts/matrix-stack/ci/matrix-authentication-service-minimal-values.yaml
@@ -8,7 +8,7 @@ elementWeb:
   enabled: false
 
 initSecrets:
-  enabled: false
+  enabled: true
 
 matrixAuthenticationService:
   enabled: true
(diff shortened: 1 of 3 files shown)


tests/manifests/__init__.py
```diff
@@ -52,6 +58,7 @@ def _enrich_components_to_test(details) -> dict[str, Any]:
         _component_details[component].setdefault("has_ingress", True)
         _component_details[component].setdefault("has_service_monitor", True)
         _component_details[component].setdefault("has_workloads", True)
+        _component_details[component].setdefault("has_image", True)
```
benbz: We could default this to `has_workloads` as if something doesn't have a workload it can't have an image either?

tests/manifests/utils.py
```diff
@@ -36,6 +36,11 @@ def component(values_file):
     return values_files_to_components[values_file]
 
 
+@pytest.fixture(scope="session")
+def base_values(values_file) -> dict[str, Any]:
+    return yaml.safe_load(Path("charts/matrix-stack/values.yaml").read_text("utf-8"))
```
benbz, line 41: Hmm if we doing this, I wonder whether it is worth parsing this for `has_workloads` & `has_image`. I don't think it should be done for `has_servicemonitor` as we want to catch components missing service monitors

tests/manifests/utils.py
```diff
@@ -92,6 +97,16 @@ def iterate_component_workload_parts(component, values, setter):
         setter(values.setdefault(shared_component, {}), values)
 
 
+def iterate_component_image_parts(component, values, setter):
```
benbz, line 176: Implement this and `iterate_component_workload_parts` in terms of a common function as they should be the same except for the test. That would give `iterate_component_workload_parts` the shared component `has_workloads` test

tests/manifests/__init__.py
gaelgatelement: `initSecrets` has a workload but uses the image from `matrixTools`

benbz: What I'm suggesting is `_component_details[component].setdefault("has_image", _component_details[component]["has_workloads"])` to avoid a) having to configure things for WKD and b) have a situation where we've said you have an image but not a workload

gaelgatelement: oh yes indeed!

tests/manifests/utils.py
```diff
@@ -83,13 +116,14 @@ async def _make_templates(values):
     return _make_templates
 
 
-def iterate_component_workload_parts(component, values, setter):
-    if component_details[component]["has_workloads"]:
+def iterate_component_workload_parts(component, values, setter, if_condition="has_workloads"):
```
benbz: Sorry I think I wasn't clear. I was happy to have 2 different iterate functions, I was just suggesting they are implemented in terms of a common function that has the `if_condition` arg too.

tests/manifests/utils.py
gaelgatelement: Problem if we do this is that the test might miss workload keys misplaced. `has_workloads` would be automatically turned off, and the key value changes would never be tested.

benbz: Right I suppose that's possible too

PR #158, opened 2025-02-10
manifest tests: fix behaviour when list contains non-dicts

gaelgatelement:
Encountered a small issue while testing locally.

### DIFF
--- a/tests/manifests/utils.py
+++ b/tests/manifests/utils.py
@@ -96,7 +96,7 @@ def generated_secrets(release_name: str, values: Any | None, helm_generated_temp
 
 def external_secrets(values):
     def find_credential(values_fragment):
-        for value in values_fragment.values():
+        for value in values_fragment.values() if isinstance(values_fragment, dict) else values_fragment:
             if isinstance(value, dict):
                 if "secret" in value and "secretKey" in value and len(value) == 2:
                     yield (value["secret"], value["secretKey"])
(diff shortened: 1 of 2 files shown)


tests/manifests/utils.py
```diff
@@ -106,7 +106,8 @@ def find_credential(values_fragment):
                     yield from find_credential(value)
             elif isinstance(value, list):
                 for inner in value:
-                    yield from find_credential(inner)
+                    if isinstance(inner, dict):
```
benbz: Should we also need to check if the value is a list?

tests/manifests/utils.py
```diff
@@ -106,7 +106,8 @@ def find_credential(values_fragment):
                     yield from find_credential(value)
             elif isinstance(value, list):
                 for inner in value:
-                    yield from find_credential(inner)
+                    if isinstance(inner, (dict, list)):
+                        yield from find_credential(inner)
```
benbz, line 110: Do we need this change anymore

gaelgatelement: Yes because inner can be `string`, `int` etc

benbz: But then it will recurse in, hit the else you've added above and then do nothing

benbz: Still needed as we can't / shouldn't iterate over a scalar value returned by the `else` above and so we still need to filter down here

PR #159, opened 2025-02-10
CI values files rationalisation

benbz:
Have almost all of `charts/matrix-stack/ci/*-values.yaml` be generated from fragments that get merged together. This allows us to
* Consistently use fragments (e.g. minimal config, checkov suppressions) between files
* Remove Secrets that can be generated from all values files in 1 go
* Build all values files on top of `nothing-enabled-values.yaml` to ensure components aren't accidentally enabled
* Reduced c/p'ing a lot

When can also now enforce testing of Helm inlined credentials and external Secrets for components that have Secrets in addition to generated credentials


tests/manifests/__init__.py
```diff
@@ -62,6 +62,15 @@ def _enrich_components_to_test(details) -> dict[str, Any]:
         _component_details[component].setdefault("sub_components", {})
         for sub_component in _component_details[component]["sub_components"]:
             _component_details[component]["sub_components"][sub_component].setdefault("has_service_monitor", True)
+
+        _component_details[component]["secret_values_files"] = []
+        if "initSecrets" in _component_details[component].setdefault("shared_components", []):
```
gaelgatelement, line 67: Components not using `initSecrets` job might still need to verify values vs external values?

benbz: I've made is possible to explicitly specify some files for any component but components using `initSecrets` are known to have generate-able Secrets and so will always get these 2 files added

gaelgatelement: Right, if they use secret, they *must* also support helm values / external secrets. Whereas for other components, it might not mandatory (they might not use secrets at all).

gaelgatelement:
The main README probably needs an update to describe the new CI values files generation.

README.md
```diff
@@ -44,6 +44,33 @@ Similarly the version number of the chart can be changed with
...
+`nothing-enabled-values.yaml` such that no default-enabled components are configured.
+
+For each component there must be a values file named
+* `<component>-minimal-values.yaml` - this should contain the absolute minimal values
+  required to get a working instance of the component. Any credentials should be
+  generated if possible
+* `<component>-checkov-values.yaml` - this should contain the minimal values to get
+  checkov to pass on a working instance of the component
+* If the component has credentials the following values are also required:
+  * `<component>-secrets-in-helm-value.yaml` - where any credentials are provide in
```
gaelgatelement: (nit)
```suggestion
  * `<component>-secrets-in-helm-value.yaml` - where any credentials are provided in
```

PR #163, opened 2025-02-10
Allow generation of the registration shared secret

benbz:
Also we can load both `registration_shared_secret` (since 1.67.0) & `macaroon` (since 1.121.0) from files rather than having to do env var string substitution.

### DIFF
--- a/charts/matrix-stack/templates/synapse/_helpers.tpl
+++ b/charts/matrix-stack/templates/synapse/_helpers.tpl
@@ -122,21 +122,6 @@ app.kubernetes.io/version: {{ .image.tag }}
 {{- define "element-io.synapse.matrixToolsEnv" }}
 {{- $root := .root -}}
 {{- with required "element-io.synapse.matrixToolsEnv missing context" .context -}}
-- name: SYNAPSE_MACAROON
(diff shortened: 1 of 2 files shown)


tests/integration/fixtures/users.py
```diff
@@ -13,7 +13,7 @@
 
 
 @pytest.fixture(scope="session")
-async def users(request, matrix_stack, kube_client, generated_data: ESSData, ssl_context, ingress_ready):
+async def users(request, matrix_stack, secrets_generated, generated_data: ESSData, ssl_context, ingress_ready):
```
gaelgatelement, line 16: Doesn't `secrets_generated` and `matrix_stack` have the same purpose, as `matrix_stack` will become ready only after the job has completed?

benbz: Yes I suppose it does, `ingress_ready` would also imply this to. I'm happy to remove it, I'm also happy to keep it if we want to be explicit that the chart needs to be deployed first.

gaelgatelement: Ah it's the same pattern as users actually, I missed next line. LGTM

PR #171, opened 2025-02-11
matrix-tools: internal command handling refactoring

gaelgatelement:
Minor cleanups.

### DIFF
--- a/matrix-tools/cmd/main.go
+++ b/matrix-tools/cmd/main.go
@@ -34,11 +34,11 @@ func getKubernetesClient() (kubernetes.Interface, error) {
 }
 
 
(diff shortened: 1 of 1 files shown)


matrix-tools/cmd/main.go
```diff
@@ -32,19 +33,31 @@ func getKubernetesClient() (kubernetes.Interface, error) {
 	return clientset, nil
 }
 
+
+func readFiles(configFiles []string) ([]io.Reader, []func() error, error) {
```
AndrewFerr: Nitpick: maybe rename this to `paths`, because
1) they could be any kind of file, not just config files
2) renaming it makes it clear that it's a different type from `files`

PR #172, opened 2025-02-12
Add internal PostgreSQL support

gaelgatelement:
- This creates a PostgreSQL statefulset if a component needs it, and its postgres database is not configured
- It is enabled by default
- Init Secrets will also initialize the internal postgres password if necessary
- Each component has its own user, generated password and database in this internal postgres

### DIFF
--- a/charts/matrix-stack/ci/fragments/postgres-synapse-secrets-in-helm.yaml
+++ b/charts/matrix-stack/ci/fragments/postgres-synapse-secrets-in-helm.yaml
@@ -3,8 +3,7 @@
 # SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 
 postgres:
-  enabled: true
+  essPasswords:
+    synapse:
+      value: CHANGEME-yuhaoshiupahmaGheiheiloJuone3aim
 
-  adminPassword:
-    secret: "{{ $.Release.Name }}-pg-external"
-    secretKey: adminPasswordShared
--- a/charts/matrix-stack/ci/matrix-authentication-service-postgres-secrets-externally-values.yaml
+++ b/charts/matrix-stack/ci/matrix-authentication-service-postgres-secrets-externally-values.yaml
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 #
-# source_fragments: matrix-authentication-service-minimal.yaml matrix-authentication-service-secrets-externally.yaml postgres-secrets-externally.yaml
+# source_fragments: matrix-authentication-service-minimal.yaml matrix-authentication-service-secrets-externally.yaml postgres-secrets-externally.yaml postgres-matrix-authentication-service-secrets-externally.yaml
 # DO NOT EDIT DIRECTLY. Edit the fragment files to add / modify / remove values
 
 initSecrets:
@@ -43,9 +43,6 @@ postgres:
     matrixAuthenticationService:
       secret: "{{ $.Release.Name }}-pg-external"
       secretKey: masPasswordShared
-    synapse:
-      secret: "{{ $.Release.Name }}-pg-external"
-      secretKey: synapsePasswordShared
 synapse:
   enabled: false
 wellKnownDelegation:
--- a/charts/matrix-stack/ci/matrix-authentication-service-postgres-secrets-in-helm-values.yaml
+++ b/charts/matrix-stack/ci/matrix-authentication-service-postgres-secrets-in-helm-values.yaml
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 #
-# source_fragments: matrix-authentication-service-minimal.yaml matrix-authentication-service-secrets-externally.yaml postgres-secrets-in-helm.yaml
+# source_fragments: matrix-authentication-service-minimal.yaml matrix-authentication-service-secrets-externally.yaml postgres-secrets-in-helm.yaml postgres-matrix-authentication-service-secrets-in-helm.yaml
 # DO NOT EDIT DIRECTLY. Edit the fragment files to add / modify / remove values
 
 initSecrets:
@@ -41,8 +41,6 @@ postgres:
   essPasswords:
     matrixAuthenticationService:
       value: CHANGEME-aiT7eisheim3Ojo3rongikuo1eiV3Ooh
-    synapse:
-      value: CHANGEME-PahpeeMeesha5aileGhaebaiceabeime
 synapse:
   enabled: false
 wellKnownDelegation:
--- a/charts/matrix-stack/ci/synapse-postgres-secrets-externally-values.yaml
+++ b/charts/matrix-stack/ci/synapse-postgres-secrets-externally-values.yaml
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 #
-# source_fragments: synapse-minimal.yaml synapse-secrets-externally.yaml postgres-secrets-externally.yaml
+# source_fragments: synapse-minimal.yaml synapse-secrets-externally.yaml postgres-secrets-externally.yaml postgres-synapse-secrets-externally.yaml
 # DO NOT EDIT DIRECTLY. Edit the fragment files to add / modify / remove values
 
 initSecrets:
@@ -17,9 +17,6 @@ postgres:
     secret: "{{ $.Release.Name }}-pg-external"
     secretKey: adminPasswordShared
   essPasswords:
-    matrixAuthenticationService:
-      secret: "{{ $.Release.Name }}-pg-external"
-      secretKey: masPasswordShared
     synapse:
       secret: "{{ $.Release.Name }}-pg-external"
       secretKey: synapsePasswordShared
--- a/charts/matrix-stack/ci/synapse-postgres-secrets-in-helm-values.yaml
+++ b/charts/matrix-stack/ci/synapse-postgres-secrets-in-helm-values.yaml
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 #
-# source_fragments: synapse-minimal.yaml synapse-secrets-in-helm.yaml postgres-secrets-in-helm.yaml
+# source_fragments: synapse-minimal.yaml synapse-secrets-in-helm.yaml postgres-secrets-in-helm.yaml postgres-synapse-secrets-in-helm.yaml
 # DO NOT EDIT DIRECTLY. Edit the fragment files to add / modify / remove values
 
 initSecrets:
@@ -16,10 +16,8 @@ postgres:
   adminPassword:
     value: CHANGEME-phiaPh8iu9tiivaiWahquaeg8ohcub4a
   essPasswords:
-    matrixAuthenticationService:
-      value: CHANGEME-aiT7eisheim3Ojo3rongikuo1eiV3Ooh
     synapse:
-      value: CHANGEME-PahpeeMeesha5aileGhaebaiceabeime
+      value: CHANGEME-yuhaoshiupahmaGheiheiloJuone3aim
 synapse:
   ingress:
     host: synapse.ess.localhost
--- a/charts/matrix-stack/ci/fragments/postgres-matrix-authentication-service-secrets-externally.yaml
+++ b/charts/matrix-stack/ci/fragments/postgres-matrix-authentication-service-secrets-externally.yaml
@@ -1,0 +1,10 @@
+# Copyright 2025 New Vector Ltd
+#
+# SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+
+postgres:
+  essPasswords:
+    matrixAuthenticationService:
+      secret: "{{ $.Release.Name }}-pg-external"
+      secretKey: masPasswordShared
+
--- a/charts/matrix-stack/ci/fragments/postgres-matrix-authentication-service-secrets-in-helm.yaml
+++ b/charts/matrix-stack/ci/fragments/postgres-matrix-authentication-service-secrets-in-helm.yaml
@@ -3,8 +3,7 @@
 # SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 
 postgres:
-  enabled: true
+  essPasswords:
+    matrixAuthenticationService:
+      value: CHANGEME-aiT7eisheim3Ojo3rongikuo1eiV3Ooh
 
-  adminPassword:
-    secret: "{{ $.Release.Name }}-pg-external"
-    secretKey: adminPasswordShared
--- a/charts/matrix-stack/ci/fragments/postgres-secrets-in-helm.yaml
+++ b/charts/matrix-stack/ci/fragments/postgres-secrets-in-helm.yaml
@@ -7,10 +7,3 @@ postgres:
 
   adminPassword:
     value: CHANGEME-phiaPh8iu9tiivaiWahquaeg8ohcub4a
-
-  essPasswords:
-    matrixAuthenticationService:
-      value: CHANGEME-aiT7eisheim3Ojo3rongikuo1eiV3Ooh
-    synapse:
-      value: CHANGEME-PahpeeMeesha5aileGhaebaiceabeime
-
--- a/charts/matrix-stack/ci/fragments/postgres-synapse-secrets-externally.yaml
+++ b/charts/matrix-stack/ci/fragments/postgres-synapse-secrets-externally.yaml
@@ -3,8 +3,8 @@
 # SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 
 postgres:
-  enabled: true
+  essPasswords:
+    synapse:
+      secret: "{{ $.Release.Name }}-pg-external"
+      secretKey: synapsePasswordShared
 
-  adminPassword:
-    secret: "{{ $.Release.Name }}-pg-external"
-    secretKey: adminPasswordShared


charts/matrix-stack/README.md
```diff
@@ -10,16 +10,27 @@ This Helm chart deploys a coherent Matrix Stack. It currently includes the follo
...
-* If using Synapse or Matrix Authentication Service, a Postgres instance
 * If using Synapse, the ability to create `PersistentVolumeClaims` to store media
 
+### Recommandations
+
+The chart comes with an internal postgres database which will be automatically setup by default.
+
+Although we are trying to update the database reliably with the chart, it is provided as-is without warranty of any kind.
+
+On a production deployment, we advise you to host your own Postgres instance, and configure it accordingly
```
benbz:
```suggestion
On a production deployment, we advise you to host your own Postgres instance, and configure it accordingly. The chart's built in Postgres is primarily for demonstration
```

Given we don't provide a warranty for anything on FOSS

charts/matrix-stack/configs/matrix-authentication-service/config.yaml.tpl
```diff
@@ -34,10 +34,17 @@ http:
...
+
 database:
+{{- if .postgres }}
+{{- with .postgres }}
   uri: "postgresql://{{ .user }}:${POSTGRES_PASSWORD}@{{ tpl .host $root }}:{{ .port }}/{{ .database }}?{{ with .sslMode }}sslmode={{ . }}&{{ end }}application_name=matrix-authentication-service"
 {{- end }}
+{{- else if $root.Values.postgresql.enabled }}
+  uri: "postgresql://ess_user:${POSTGRES_PASSWORD}@{{ $root.Release.Name }}-postgresql.{{ $root.Release.Namespace }}.svc.cluster.local:5432/mas?sslmode=prefer&application_name=matrix-authentication-service"
+{{ else }}
+  {{ fail "MAS requires .matrixAuthenticationService.postgres.* configured, or internal chart .postgres to be enabled" }}
```
benbz:
```suggestion
  {{ fail "MAS requires matrixAuthenticationService.postgres.* configured, or internal chart postgres to be enabled with postgresql.enabled: true" }}
```

charts/matrix-stack/source/values.schema.json
```diff
@@ -52,6 +52,9 @@
     "matrixAuthenticationService": {
       "$ref": "file://matrixAuthenticationService.json"
     },
+    "postgresql": {
```
benbz:
```suggestion
    "postgres": {
```

We use `postgres` under `synapse` and `matrixAuthenticationService` and so should probably be consistent

charts/matrix-stack/source/postgresql.yaml.j2
```diff
@@ -0,0 +1,27 @@
...
+postgresExporter:
+  {{- sub_schema_values.image(registry='docker.io', repository='prometheuscommunity/postgres-exporter', tag='v0.16.0') | indent(2) }}
+  {{- sub_schema_values.resources(requests_memory='100Mi', requests_cpu='100m', limits_memory='500Mi')| indent(2) }}
+  {{- sub_schema_values.containersSecurityContext()| indent(2) }}
+
+{{- sub_schema_values.credential("Postgres Admin Password", "adminPassword", initIfAbsent=true) }}
+{{- sub_schema_values.credential("ESS Role Password", "essPassword", initIfAbsent=true) }}
+
+{{- sub_schema_values.persistentVolumeClaim("storage") }}
+{{- sub_schema_values.image(registry='docker.io', repository='postgres', tag='16-alpine') }}
```
benbz:
```suggestion
{{- sub_schema_values.image(registry='docker.io', repository='postgres', tag='17-alpine') }}
```

Let's safe ourselves the pain of Postgres major version upgrades for as long as possible

charts/matrix-stack/source/postgresql.yaml.j2
```diff
@@ -0,0 +1,27 @@
+{#
+Copyright 2024 New Vector Ltd
+
+SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+#}
+enabled: true
+
+postgresExporter:
+  {{- sub_schema_values.image(registry='docker.io', repository='prometheuscommunity/postgres-exporter', tag='v0.16.0') | indent(2) }}
+  {{- sub_schema_values.resources(requests_memory='100Mi', requests_cpu='100m', limits_memory='500Mi')| indent(2) }}
```
benbz: Does the exporter really need this much requests (i.e. the same as Postgres itself)?

charts/matrix-stack/source/common/sub_schema_values.yaml.j2
```diff
@@ -255,8 +255,8 @@ labels: {}
   # database:
 
   ## TLS settins to use for the PostgreSQL connection
-  sslMode: prefer
-{{ credential("PostgreSQL password", "password") | indent(2) }}
+  # sslMode: prefer
+{{ (credential("PostgreSQL password", "password") | indent(2)).split('\n') | join('\n#') }}
```
benbz: Might be better to add a `commented` flag to `credential` as this results in double-commenting the comments and the comment isn't indented. Neither is the end of the world but isn't pleasant either

charts/matrix-stack/source/postgresql.json
```diff
@@ -0,0 +1,61 @@
+{
+  "$id": "file://synapse",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "type": "object",
+  "additionalProperties": true,
```
benbz: Why do we need this

charts/matrix-stack/source/postgresql.json
```diff
@@ -0,0 +1,61 @@
...
+          "$ref": "file://common/resources.json"
+        }
+      }
+    },
+    "adminPassword": {
+      "$ref": "file://common/credential.json"
+    },
+    "essPassword": {
+      "$ref": "file://common/credential.json"
+    },
```
benbz: I thought we said the internal Postgres should require `initSecrets`, in which case these can disappear

charts/matrix-stack/templates/init-secrets/_helpers.tpl
```diff
@@ -17,6 +17,16 @@ app.kubernetes.io/version: {{ $root.Values.matrixTools.image.tag }}
 
 {{- define "element-io.init-secrets.generated-secrets" -}}
 {{- $root := .root -}}
+{{- with $root.Values.postgresql }}
+{{- if (include "element-io.postgresql.enabled" (dict "root" $root)) -}}
+{{- if not .essPassword }}
+- {{ (printf "%s-generated" $root.Release.Name) }}:POSTGRESQL_ESS_PASSWORD:rand32
```
benbz: I have a strong preference for per-application users & passwords. This protects us from misconfiguration in the Helm chart writing to the wrong DB or the app going a bit crazy trying to provision itself

charts/matrix-stack/ci/example-default-enabled-components-values.yaml
```diff
@@ -13,23 +13,11 @@ elementWeb:
...
     host: mas.ess.localhost
-  postgres:
-    host: postgres
-    user: mas
-    database: mas
-    password:
-      secret: ess-postgres
-      secretKey: password
+postgresql:
+  enabled: false
```
benbz: We've removed the test Postgres, so this should be enabled

charts/matrix-stack/ci/synapse-ingress-additional-paths-values.yaml
```diff
@@ -11,6 +11,8 @@ elementWeb:
   enabled: false
 matrixAuthenticationService:
   enabled: false
+postgresql:
+  enabled: false
```
benbz: Should be enabled as the test Postgres has been removed

charts/matrix-stack/configs/matrix-authentication-service/config.yaml.tpl
```diff
@@ -34,10 +34,17 @@ http:
...
+
 database:
+{{- if .postgres }}
+{{- with .postgres }}
   uri: "postgresql://{{ .user }}:${POSTGRES_PASSWORD}@{{ tpl .host $root }}:{{ .port }}/{{ .database }}?{{ with .sslMode }}sslmode={{ . }}&{{ end }}application_name=matrix-authentication-service"
 {{- end }}
+{{- else if $root.Values.postgresql.enabled }}
+  uri: "postgresql://ess_user:${POSTGRES_PASSWORD}@{{ $root.Release.Name }}-postgresql.{{ $root.Release.Namespace }}.svc.cluster.local:5432/mas?sslmode=prefer&application_name=matrix-authentication-service"
+{{ else }}
+  {{ fail "MAS requires .matrixAuthenticationService.postgres.* configured, or internal chart .postgresql to be enabled" }}
```
benbz:
```suggestion
  {{ fail "MAS requires matrixAuthenticationService.postgres.* configured, or internal chart Postgres to be enabled with postgres.enabled: true" }}
```

charts/matrix-stack/templates/synapse/synapse_statefulset.yaml
```diff
@@ -132,7 +132,11 @@ We have an init container to render & merge the config for several reasons:
         - "/matrix-tools"
         - tcpwait
         - -address
+{{- if .postgres }}
         - "{{ (tpl .postgres.host $) }}:{{ .postgres.port }}"
+{{- else }}
+        - "{{ $.Release.Name }}-postgresql.{{ $.Release.Namespace }}.svc.cluster.local:5432"
+{{- end }}
```
benbz: Extract to a helper given both Synapse and MAS have the same logic?

charts/matrix-stack/templates/synapse/_synapse_config.tpl
```diff
@@ -51,14 +51,25 @@ registration_shared_secret_path: /secrets/{{
...
+    sslmode: {{ .postgres.sslMode | default "prefer" }}
+{{- else if $root.Values.postgresql.enabled }}
+    user: "ess_user"
+    password: ${SYNAPSE_POSTGRES_PASSWORD}
+    database: "synapse"
+    host: "{{ $root.Release.Name }}-postgresql.{{ $root.Release.Namespace }}.svc.cluster.local"
+    port: 5432
+    sslmode: prefer
+{{ else }}
+  {{ fail "Synapse requires .synapse.postgres.* configured, or internal chart .postgresql to be enabled" }}
```
benbz:
```suggestion
  {{ fail "Synapse requires synapse.postgres.* configured, or internal chart Postgres to be enabled with postgres.enabled: true" }}
```

charts/matrix-stack/ci/fragments/synapse-test-postgres.yaml
benbz, line 1: File can go entirely?

charts/matrix-stack/ci/fragments/matrix-authentication-service-test-postgres.yaml
benbz, line 1: File can go entirely

charts/matrix-stack/templates/postgresql/statefulset.yaml
```diff
@@ -0,0 +1,161 @@
...
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  labels:
+    {{- include "element-io.postgresql.labels" (dict "root" $ "context" .) | nindent 4 }}
+  name: {{ $.Release.Name }}-postgresql
+  namespace: {{ $.Release.Namespace }}
+  annotations:
+    k8s.element.io/confighash: "{{ include (print $.Template.BasePath "/postgresql/secret.yaml") $ | sha1sum }}"
+    k8s.element.io/logconfighash: "{{ include (print $.Template.BasePath "/postgresql/configmap.yaml") $ | sha1sum }}"
```
benbz: Not logging config

charts/matrix-stack/templates/postgresql/statefulset.yaml
```diff
@@ -0,0 +1,161 @@
...
+{{- with .annotations }}
+    {{- toYaml . | nindent 4 }}
+{{- end }}
+spec:
+  serviceName: {{ $.Release.Name }}-postgresql
+  replicas: {{ .instances | default 1 }}
```
benbz: Must be hard-coded to 1. Is `instances` even a property?

tests/manifests/test_secrets_consistency.py
```diff
@@ -56,7 +56,7 @@ def get_volume_from_mount(template, volume_mount):
 
 def match_in_content(container_name, mounted_secret_keys, mount_path, match_in):
     found_mount = False
-    for match in re.findall(rf"{mount_path}/([^\s\n\");]+)", match_in):
+    for match in re.findall(rf"{mount_path}/([^\s\n\")`;]+)", match_in):
```
benbz: This definitely looks like an invalid path character, but what necessitated this change?

charts/matrix-stack/templates/postgresql/configmap.yaml
```diff
@@ -0,0 +1,47 @@
...
+      )
+    ) }}`;
+    (
+      (echo -n $POSTGRES_PASSWORD | psql -W -U postgres -tc "SELECT 1 FROM pg_catalog.pg_roles WHERE rolname = 'ess_user'" | grep -q 1) && \
+      (echo -n $POSTGRES_PASSWORD | psql -W -U postgres -c "ALTER USER ess_user PASSWORD '"$ESS_PASSWORD"'")
+    ) || \
+    (echo -n $POSTGRES_PASSWORD | psql -W -U postgres -c "CREATE ROLE ess_user LOGIN PASSWORD '"$ESS_PASSWORD"'");
+{{- range $dbname := list "synapse" "mas" }}
+    (echo -n $POSTGRES_PASSWORD | psql -W -U postgres -tc "SELECT 1 FROM pg_database WHERE datname = '{{ $dbname }}'" | grep -q 1) || \
+    (echo -n $POSTGRES_PASSWORD | createdb --encoding=UTF8 --locale=C --template=template0 --owner=ess_user {{ $dbname }} -U postgres)
```
benbz: I think we have to do this as a sidecar rather than one-time, esp if we're allowing non-generated secrets as the passwords could change

tests/manifests/test_secrets_consistency.py
gaelgatelement: The config map of postgresql uses backtick syntax : 

```
    export POSTGRES_PASSWORD=`cat /secrets/{{
        include "element-io.ess-library.init-secret-path" (
      dict "root" $ "context" (
        dict "secretProperty" .adminPassword
              "initSecretKey" "POSTGRESQL_ADMIN_PASSWORD"
              "defaultSecretName" (printf "%s-postgresql" $.Release.Name)
              "defaultSecretKey" "ADMIN_PASSWORD"
        )
    ) }}`;
```

charts/matrix-stack/templates/postgresql/configmap.yaml
gaelgatelement: I *think* init-dbs scripts runs on each pod startup ? But I should probably test it

charts/matrix-stack/ci/fragments/matrix-authentication-service-test-postgres.yaml
gaelgatelement: I need it to combine and verify both MAS postgres with external external and MAS postgres with in-helm secrets.

benbz: Sorry I'm failing to understand this comment

gaelgatelement: It should be clearer now, this file is gone, but we have 2 new files to test MAS+External postgres secrets.

charts/matrix-stack/source/postgres.json
```diff
@@ -0,0 +1,61 @@
+{
+  "$id": "file://synapse",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "type": "object",
+  "additionalProperties": true,
```
benbz: Reattaching the comment about why is this needed?

charts/matrix-stack/configs/matrix-authentication-service/config.yaml.tpl
```diff
@@ -34,10 +34,17 @@ http:
...
+
 database:
+{{- if .postgres }}
+{{- with .postgres }}
   uri: "postgresql://{{ .user }}:${POSTGRES_PASSWORD}@{{ tpl .host $root }}:{{ .port }}/{{ .database }}?{{ with .sslMode }}sslmode={{ . }}&{{ end }}application_name=matrix-authentication-service"
 {{- end }}
+{{- else if $root.Values.postgres.enabled }}
+  uri: "postgresql://ess_user:${POSTGRES_PASSWORD}@{{ $root.Release.Name }}-postgres.{{ $root.Release.Namespace }}.svc.cluster.local:5432/mas?sslmode=prefer&application_name=matrix-authentication-service"
+{{ else }}
+  {{ fail "MAS requires .matrixAuthenticationService.postgres.* configured, or internal chart .postgres to be enabled with postgres.enabled: true" }}
```
benbz:
```suggestion
  {{ fail "MAS requires matrixAuthenticationService.postgres.* to be configured, or the internal chart Postgres to be enabled with postgres.enabled: true" }}
```

Comment was only partially addressed and some more wording changes added too

charts/matrix-stack/templates/ess-library/_credentials.tpl
```diff
@@ -59,3 +59,24 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
...
+{{- define "element-io.ess-library.postgres-secret-path" -}}
+{{- $root := .root -}}
+{{- with required "element-io.ess-library.postgres-secret-path" .context -}}
+{{- $secretProperty := .secretProperty -}}
+{{- $defaultSecretName := required "element-io.ess-library.postgres-secret-path context missing defaultSecretName" .defaultSecretName -}}
+{{- $defaultSecretKey := required "element-io.ess-library.postgres-secret-path context missing defaultSecretKey" .defaultSecretKey -}}
+{{- if not $secretProperty -}}
+  {{- if not $root.Values.postgres.essPassword }}
+    {{- if $root.Values.initSecrets.enabled -}}
+    {{- printf "%s/%s" (printf "%s-generated" $root.Release.Name) "POSTGRESQL_ESS_PASSWORD" -}}
```
benbz: Reattaching the comment about the fact that we'd previously said the internal Postgres requires `initSecrets`

charts/matrix-stack/templates/matrix-authentication-service/deployment.yaml
```diff
@@ -176,7 +176,11 @@ spec:
         - "/matrix-tools"
         - tcpwait
         - -address
+{{- if .postgres }}
         - "{{ (tpl .postgres.host $) }}:{{ .postgres.port }}"
+{{- else }}
+        - "{{ $.Release.Name }}-postgres.{{ $.Release.Namespace }}.svc.cluster.local:5432"
+{{- end }}
```
benbz: Extract to a helper given used by Synapse + MAS + any future Postgres using apps

charts/matrix-stack/templates/synapse/_synapse_config.tpl
```diff
@@ -51,14 +51,25 @@ registration_shared_secret_path: /secrets/{{
...
+    sslmode: {{ .postgres.sslMode | default "prefer" }}
+{{- else if $root.Values.postgres.enabled }}
+    user: "ess_user"
+    password: ${SYNAPSE_POSTGRES_PASSWORD}
+    database: "synapse"
+    host: "{{ $root.Release.Name }}-postgres.{{ $root.Release.Namespace }}.svc.cluster.local"
+    port: 5432
+    sslmode: prefer
+{{ else }}
+  {{ fail "Synapse requires .synapse.postgres.* configured, or internal chart .postgres to be enabled with postgres.enabled: true" }}
```
benbz:
```suggestion
  {{ fail "Synapse requires synapse.postgres.* to be configured, or the internal chart Postgres to be enabled with postgres.enabled: true" }}
```

Comment was only partially addressed and some more wording changes added too

charts/matrix-stack/templates/synapse/synapse_statefulset.yaml
```diff
@@ -158,8 +162,10 @@ We have an init container to render & merge the config for several reasons:
         - >
           [ -e "/usr/lib/`uname -m`-linux-gnu/libjemalloc.so.2" ] && export LD_PRELOAD="/usr/lib/`uname -m`-linux-gnu/libjemalloc.so.2";
           python3 -m {{ include "element-io.synapse.process.app" (dict "root" $ "context" $processType) }} -c /conf/homeserver.yaml
+{{- with (include "element-io.synapse.env" (dict "root" $ "context" .)) }}
         env:
-          {{- include "element-io.synapse.env" (dict "root" $ "context" .) | nindent 10 }}
+          {{- . | nindent 10 }}
+{{- end }}
```
benbz, line 168: Any particular reason for this change, given I think we'll always have the hostname/application name env var?

charts/matrix-stack/source/postgres.json
gaelgatelement: URG I wondered why some obvious schema schema issues were not catched :facepalm:

charts/matrix-stack/templates/ess-library/_credentials.tpl
gaelgatelement: Mmmh true, I went a bit too far with this :roll_eyes: 

But now that this is here, maybe we can keep current mode. After all, it might happen that a customer would want to deploy the demo in a cluster without having permissions to create a role, and thus cannot rely on init Secrets.

charts/matrix-stack/templates/synapse/synapse_statefulset.yaml
gaelgatelement: > Any particular reason for this change, given I think we'll always have the hostname/application name env var?

This is true on the `matrix-tools.env` helper. But this is the synapse container which does not need these env vars. 

The `synapse.env` helper can generate empty dict. There was a mistake before where it would generate `PG_` variables for no reason.

charts/matrix-stack/templates/postgresql/configmap.yaml
gaelgatelement: This is now fixed and uses a sidecar.

gaelgatelement:
> * There's at least a couple values files that have `postgres.enabled: false` but `<component>.postgres` absent. I'd have thought that these should fail to render due to `fail` somewhere
> 
>     * The test Postgres should be ripped out of the `setup_test_cluster.sh` script

2. done in last commit

Regarding point 1. I tried to turn off chart postgres manually on minimal synapse & mas values and it failed as expected. Do you see a particular example ?

benbz:
> Regarding point 1. I tried to turn off chart postgres manually on minimal synapse & mas values and it failed as expected. Do you see a particular example ?

`charts/matrix-stack/ci/example-default-enabled-components-values.yaml` and on a prior review I think the Synapse additional paths one, though that looks fixed now

gaelgatelement:
> > Regarding point 1. I tried to turn off chart postgres manually on minimal synapse & mas values and it failed as expected. Do you see a particular example ?
> 
> `charts/matrix-stack/ci/example-default-enabled-components-values.yaml` and on a prior review I think the Synapse additional paths one, though that looks fixed now

This fails : 

```
helm template test-postgres-password -n test-ess --create-namespace -f charts/matrix-stack/ci/example-default-enabled-components-values.yaml ./charts/matrix-stack         
Error: execution error at (matrix-stack/templates/synapse/synapse_statefulset.yaml:23:38): Synapse requires synapse.postgres.* to be configured, or the internal chart Postgres to be enabled with postgres.enabled: true
```

But `ct lint` does not fail on required errors, it just prints it out  :/

benbz:
> But `ct lint` does not fail on required errors, it just prints it out :/

Urrrggh, that's going to be because `helm lint` doesn't fail even with `--strict`. I suspect we should run `ct lint` through `tee` and then manually check for `Fail` lines

charts/matrix-stack/templates/ess-library/_credentials.tpl
benbz: That's fair

charts/matrix-stack/templates/postgres/configmap.yaml
```diff
@@ -0,0 +1,47 @@
...
+            "initSecretKey" "POSTGRESQL_ESS_PASSWORD"
+            "defaultSecretName" (printf "%s-postgres" $.Release.Name)
+            "defaultSecretKey" "ESS_PASSWORD"
+      )
+    ) }}`;
+    (
+      (echo -n $POSTGRES_PASSWORD | psql -W -U postgres -tc "SELECT 1 FROM pg_catalog.pg_roles WHERE rolname = 'ess_user'" | grep -q 1) && \
+      (echo -n $POSTGRES_PASSWORD | psql -W -U postgres -c "ALTER USER ess_user PASSWORD '"$ESS_PASSWORD"'")
+    ) || \
+    (echo -n $POSTGRES_PASSWORD | psql -W -U postgres -c "CREATE ROLE ess_user LOGIN PASSWORD '"$ESS_PASSWORD"'");
```
benbz: Could be simplified by using `CREATE ROLE ... IF NOT EXISTS` and then unconditionally running the `ALTER USER ... PASSWORD`

gaelgatelement: `IF NOT EXISTS` does not exists for `CREATE ROLE` : 
 https://www.postgresql.org/docs/current/sql-createrole.html 
 https://stackoverflow.com/questions/8092086/create-postgresql-role-user-if-it-doesnt-exist

benbz: *siigh* I totally hallucinated this despite visiting that first link

charts/matrix-stack/templates/synapse/synapse_statefulset.yaml
```diff
@@ -45,7 +45,12 @@ spec:
 {{- range $index, $appservice := .appservices }}
         k8s.element.io/as-registration-{{ $index }}-hash: "{{ (lookup "v1" "ConfigMap" $.Release.Namespace $appservice.registrationFileConfigMap) | toJson | sha1sum }}"
 {{- end }}
-
+        {{ include "element-io.ess-library.postgres-label" (dict "root" $ "context" (dict
+                                                                "essPassword" "synapse"
+                                                                "postgresProperty" .postgres
+                                                                "defaultSecretName" (printf "%s-matrix-authentication-service" $.Release.Name)
```
benbz:
```suggestion
                                                                "defaultSecretName" (printf "%s-synapse" $.Release.Name)
```

Did a test catch this?

gaelgatelement: Turns out this value was ignored, but adding a test discovered a problem on MAS label :+1:

gaelgatelement:
> The new `<component>-postgres-secrets...-values.yaml` files should be added to the secrets manifests tests

I believe this is done ? 

```python

        if "postgres" in _component_details[component].setdefault("shared_components", []):
            _component_details[component]["secret_values_files"].append(
                f"{_component_details[component]['hyphened_name']}-postgres-secrets-in-helm-values.yaml"
            )
            _component_details[component]["secret_values_files"].append(
                f"{_component_details[component]['hyphened_name']}-postgres-secrets-externally-values.yaml"
            )
    return _component_details
```

charts/matrix-stack/ci/matrix-authentication-service-postgres-secrets-externally-values.yaml
```diff
@@ -0,0 +1,52 @@
...
+  adminPassword:
+    secret: "{{ $.Release.Name }}-pg-external"
+    secretKey: adminPasswordShared
+  essPasswords:
+    matrixAuthenticationService:
+      secret: "{{ $.Release.Name }}-pg-external"
+      secretKey: masPasswordShared
+    synapse:
+      secret: "{{ $.Release.Name }}-pg-external"
+      secretKey: synapsePasswordShared
```
benbz:
```suggestion
```

charts/matrix-stack/ci/matrix-authentication-service-postgres-secrets-in-helm-values.yaml
```diff
@@ -0,0 +1,49 @@
...
+      secretKey: keysEcdsaSecp384r1
+postgres:
+  enabled: true
+  adminPassword:
+    value: CHANGEME-phiaPh8iu9tiivaiWahquaeg8ohcub4a
+  essPasswords:
+    matrixAuthenticationService:
+      value: CHANGEME-aiT7eisheim3Ojo3rongikuo1eiV3Ooh
+    synapse:
+      value: CHANGEME-PahpeeMeesha5aileGhaebaiceabeime
```
benbz:
```suggestion
```

charts/matrix-stack/ci/synapse-postgres-secrets-externally-values.yaml
```diff
@@ -0,0 +1,40 @@
...
+  enabled: false
+postgres:
+  enabled: true
+  adminPassword:
+    secret: "{{ $.Release.Name }}-pg-external"
+    secretKey: adminPasswordShared
+  essPasswords:
+    matrixAuthenticationService:
+      secret: "{{ $.Release.Name }}-pg-external"
+      secretKey: masPasswordShared
```
benbz:
```suggestion
```

charts/matrix-stack/ci/synapse-postgres-secrets-in-helm-values.yaml
```diff
@@ -0,0 +1,34 @@
...
+  enabled: false
+matrixAuthenticationService:
+  enabled: false
+postgres:
+  enabled: true
+  adminPassword:
+    value: CHANGEME-phiaPh8iu9tiivaiWahquaeg8ohcub4a
+  essPasswords:
+    matrixAuthenticationService:
+      value: CHANGEME-aiT7eisheim3Ojo3rongikuo1eiV3Ooh
```
benbz:
```suggestion
```

tests/manifests/__init__.py
```diff
@@ -80,7 +90,7 @@ def _enrich_components_to_test(details) -> dict[str, Any]:
 values_files_to_components = {
     values_file: component
     for component, details in component_details.items()
-    for values_file in details["values_files"]
+    for values_file in (details["values_files"] + details["secret_values_files"])
```
benbz, line 93: Hmm this was deliberately omitted or we test all the secret value files in almost all tests. If there's a good reason to do this, we should remove `secrets_values_files_to_test` as all files will be in `values_files_to_test`

gaelgatelement: Mmmh this got missed when I did it properly after testing this

charts/matrix-stack/ci/synapse-postgres-secrets-in-helm-values.yaml
gaelgatelement: Well its part of the fragment `postgres-secrets-in-helm.yaml` . The only solution if we dont want really clean files is to have 1 `postgres-secrets-in-helm.yaml` per component using it.

benbz: Hmm and we can't put it in the per-component normal secrets file as that will enable external Postgres. Happy to keep it as it is now, but my worry is that we won't catch things where we use the wrong Secret

tests/manifests/__init__.py
gaelgatelement: Mmmh but this is `values_files_to_components`, not `values_files_to_test` ? This is to be able to resolve components when testing `secret_values_files` in the `component` fixture.

benbz: `values_files_to_test` is generated from `values_files_to_components`

gaelgatelement: Why dont you generate it from `for values_file in (details["values_files"]`  ?

benbz: So we've either got to unpick that link or provide a `secrets_values_files_to_components`

gaelgatelement: I suggest to do : 

```python

values_files_to_test = [values_file
                        for details in component_details.values()
                        for values_file in (details["values_files"])
                        ]
```

benbz: > Why dont you generate it from `for values_file in (details["values_files"]` ?

No reason, other than until now `values_files_to_components.keys()` was the same

charts/matrix-stack/ci/synapse-postgres-secrets-in-helm-values.yaml
gaelgatelement: Mmhm right. Let's do the effort then.

charts/matrix-stack/ci/fragments/postgres-synapse-secrets-in-helm.yaml
```diff
@@ -0,0 +1,8 @@
+# Copyright 2025 New Vector Ltd
+#
+# SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+
+postgres:
+    matrixAuthenticationService:
+      value: CHANGEME-aiT7eisheim3Ojo3rongikuo1eiV3Ooh
```
benbz:
```suggestion
postgres:
  essPasswords:
    synapse:
      value: CHANGEME-aiT7eisheim3Ojo3rongikuo1eiV3Ooh
```

PR #173, opened 2025-02-12
matrix-tools: add concat command

AndrewFerr:
Add a command that appends a list of files to a target file. This provides an alternative to update-ca-certificates for init containers to use with minimal container images.


AndrewFerr:
Not needed for FOSS.

PR #187, opened 2025-02-13
init-secret-path: fail if initSecret is disabled and secret is not provided

gaelgatelement:
The script `bash ct-lint.sh` will now make sure that nothing raises when its ran.

Example output : 

```


------------------------------------------------------------------------------------------------------------------------
 ✔︎ matrix-stack => (version: "0.5.1-dev", path: "charts/matrix-stack")
------------------------------------------------------------------------------------------------------------------------
All charts linted successfully
Errors were raised while running ct lint, exiting with error
------------------
engine.go:227: [INFO] Fail: initSecrets is disabled, but secret .synapse.signingKey is empty.
engine.go:227: [INFO] Fail: initSecrets is disabled, but secret .synapse.macaroon is empty.
engine.go:227: [INFO] Fail: initSecrets is disabled, but secret .synapse.registrationSharedSecret is empty.
engine.go:227: [INFO] Fail: initSecrets is disabled, but secret .synapse.signingKey is empty.
...
```

### DIFF
--- a/charts/matrix-stack/templates/synapse/_synapse_config.tpl
+++ b/charts/matrix-stack/templates/synapse/_synapse_config.tpl
@@ -22,7 +22,7 @@ signing_key_path: /secrets/{{
   include "element-io.ess-library.init-secret-path" (
     dict "root" $root "context" (
       dict "secretProperty" .signingKey
-            "secretPath" ".synapse.signingKey"
+           "secretPath" ".synapse.signingKey"
            "initSecretKey" "SYNAPSE_SIGNING_KEY"
            "defaultSecretName" (printf "%s-synapse" $root.Release.Name)
            "defaultSecretKey" "SIGNING_KEY"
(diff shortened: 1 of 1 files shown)


charts/matrix-stack/templates/synapse/_synapse_config.tpl
```diff
@@ -22,6 +22,7 @@ signing_key_path: /secrets/{{
   include "element-io.ess-library.init-secret-path" (
     dict "root" $root "context" (
       dict "secretProperty" .signingKey
+            "secretPath" ".synapse.signingKey"
```
benbz:
```suggestion
           "secretPath" ".synapse.signingKey"
```

So it lines up with the above

charts/matrix-stack/templates/synapse/_synapse_config.tpl
```diff
@@ -42,6 +44,7 @@ registration_shared_secret_path: /secrets/{{
   include "element-io.ess-library.init-secret-path" (
     dict "root" $root "context" (
       dict "secretProperty" .registrationSharedSecret
+          "secretPath" ".synapse.registrationSharedSecret"
```
benbz:
```suggestion
           "secretPath" ".synapse.registrationSharedSecret"
```

PR #190, opened 2025-02-14
pytest manifests: initSecrets is enabled by default

gaelgatelement:
The initSecrets, if not specified in the values file, should be considered enabled.


gaelgatelement:
Closing we actually want our CI Values file to explicitly set `true` when the job is expected, so that we can ensure the assertion : `AssertionError: initSecrets.enabled=true but no init-secrets job found to generate Secret for`

PR #191, opened 2025-02-14
CI: Use public runners

gaelgatelement:
In order to release public the community edition, we need to use public CI instead of our own runners. 

https://github.com/element-hq/serverproduct-internal/issues/1021

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -15,6 +15,12 @@ permissions:
   contents: read
   packages: read
 
+env:
+  DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
+  DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
+  GHCR_USERNAME: ${{ github.actor }}
+  GHCR_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
 jobs:
   pytest:
     runs-on: ubuntu-latest
--- a/tests/integration/fixtures/helm.py
+++ b/tests/integration/fixtures/helm.py
@@ -14,8 +14,8 @@ from lightkube.models.meta_v1 import ObjectMeta
 from lightkube.resources.core_v1 import Endpoints, Namespace, Secret, Service
 from lightkube.resources.networking_v1 import Ingress
 
-from ..lib.helpers import kubernetes_tls_secret
-from ..lib.utils import value_file_has
+from ..lib.helpers import kubernetes_docker_secret, kubernetes_tls_secret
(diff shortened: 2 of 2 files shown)


.github/workflows/build-test.yml
```diff
@@ -19,27 +19,15 @@ jobs:
   # We want to ensure that the commit of built changes does happen, so fail if building creates changes
   # If this gets problematic we change to not committing the built schemas/values to git
   no-changes-after-building:
-    runs-on: cpu-s
-    container:
-      image: ghcr.io/${{ github.repository }}/ci-runner
-      credentials:
-        username: ${{ github.actor }}
-        password: ${{ secrets.GITHUB_TOKEN }}
```
benbz, line 27: Can we remove `permissions.packages :read` now?

.github/workflows/build-test.yml
```diff
@@ -10,35 +10,23 @@ on:
...
-        key: venv-${{ hashFiles('poetry.lock') }}
+        python-version: "3.x"
+        poetry-version: "1.8.5"
 
-    - name: Install venv
+    - name: Load poetry path
       run: |
-        poetry env info
-        poetry install
         echo "$(poetry env info -p)/bin" >> "${GITHUB_PATH}"
```
benbz, line 30: Definitely still needed?

.github/workflows/pytest.yml
```diff
@@ -36,19 +37,41 @@ jobs:
...
+        name: "ess-helm"
+        skipClusterCreation: "true"
+        skipClusterLogsExport: "true"
+
+
+    - uses: azure/setup-helm@v4.2.0
 
     - name: Test with pytest
       run: |
-        cp tests/integration/fixtures/files/clusters/audit-policy.yml /dind-job-shared/audit-policy.yml
```
benbz, line 51: OOI what's the reason this isn't possible?

.github/workflows/build-test.yml
gaelgatelement: Yes, without it the poetry commands are not in the PATH. It's probably missing from the matrix-org action altogether?

.github/workflows/pytest.yml
gaelgatelement: It's still possible, but we can't rely on the dind shared volume. The runners are now VMs, so we just use the same method as our local dev envs.

.github/workflows/build-test.yml
benbz: Arguably yes it should be there but possibly we're a bit unusual in having apps installed by `poetry`

.github/workflows/pytest.yml
benbz: Ah right, audit policy isn't gone, just the CI fiddlyness

PR #193, opened 2025-02-17
ct-lint: fix errors were not catched

gaelgatelement:


scripts/ct-lint.sh
```diff
@@ -4,14 +4,17 @@
 #
 # SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 
+set -o pipefail
```
benbz: Do we want e and u (I think) as well?

gaelgatelement: From my local tests, it cannot hurt. Let's add it.

PR #207, opened 2025-02-18
Don't use bash to launch synapse process

AndrewFerr:
This is for using minimal container images of synapse, which won't have bash installed.


charts/matrix-stack/templates/synapse/_helpers.tpl
```diff
@@ -77,7 +77,7 @@ app.kubernetes.io/version: {{ .image.tag }}
 {{- define "element-io.synapse.env" }}
 {{- $root := .root -}}
 {{- with required "element-io.synapse.env missing context" .context -}}
-{{- $resultEnv := dict -}}
+{{- $resultEnv := dict "LD_PRELOAD" "libjemalloc.so.2" -}}
```
```suggestion
{{- $resultEnv := dict -}}
{{- $_ := set $resultEnv "LD_PRELOAD" "libjemalloc.so.2" -}}
```

So that future diffs are nicer

PR #208, opened 2025-02-19
postgres: fix duplicated port

gaelgatelement:


tests/manifests/test_pod_containers_ports.py
```diff
@@ -0,0 +1,21 @@
...
+@pytest.mark.parametrize("values_file", values_files_to_test)
+@pytest.mark.asyncio_cooperative
+async def test_sets_seccompProfile_in_pod_securityContext_by_default(templates):
+    for template in templates:
+        if template["kind"] in ["Deployment", "StatefulSet", "Job"]:
+            id = f"{template['kind']}/{template['metadata']['name']}"
+            containers = template["spec"]["template"]["spec"]["containers"] + \
+                template["spec"]["template"]["spec"].get("initContainers", [])
+            ports = []
+            for container in containers:
```
benbz:
```suggestion
            ports = []
            for container in template["spec"]["template"]["spec"]["containers"]:
```

As even if `initContainers` can have ports (not sure it makes sense) they're only run sequentially so can't clash with each other or `containers`

tests/manifests/test_pod_containers_ports.py
```diff
@@ -0,0 +1,21 @@
...
+# SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+
+import pytest
+
+from . import values_files_to_test
+
+
+@pytest.mark.parametrize("values_file", values_files_to_test)
+@pytest.mark.asyncio_cooperative
+async def test_sets_seccompProfile_in_pod_securityContext_by_default(templates):
```
benbz:
```suggestion
async def test_pods_dont_have_duplicate_ports(templates):
```

PR #213, opened 2025-02-19
artifact-hub: push artifacthub-repo.yaml

gaelgatelement:
Let's publish matrix-stack with proper metadata on artifacthub.

### DIFF
--- a/.github/workflows/artifact-hub.yml
+++ b/.github/workflows/artifact-hub.yml
@@ -18,10 +18,10 @@ jobs:
       packages: write
     runs-on: ubuntu-latest
     steps:
+    # This will push the OCI artifact only on merges
+    # As the checkout will push the PR Target commit hash
     - name: Checkout
       uses: actions/checkout@v4
-      with:
-        ref: ${{ github.event.pull_request.head.sha }}
 
     - uses: oras-project/setup-oras@v1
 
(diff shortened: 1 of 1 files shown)


.github/workflows/artifact-hub.yml
```diff
@@ -0,0 +1,40 @@
+# Copyright 2024 New Vector Ltd
```
benbz:
```suggestion
# Copyright 2025 New Vector Ltd
```

artifacthub-repo.yaml
```diff
@@ -0,0 +1,11 @@
+# Copyright 2024 New Vector Ltd
```
benbz:
```suggestion
# Copyright 2025 New Vector Ltd
```

artifacthub-repo.yaml
```diff
@@ -0,0 +1,11 @@
...
+#
+# SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+
+repositoryID: f900b890-eb67-4328-9100-336301f7cefe
+owners:
+  - name: gaelg
+    email: gaelg@element.io
```
benbz:
```suggestion
  - name: Element Server Products Team
    email: onprem-support@element.io
```

Or does this need to be an actual person?

gaelgatelement: That's a good point, I thought it had to be an actual account on artifacthub but I'm not sure actually.

gaelgatelement: I could not find the information. The field is optional, so I suppose it's not so important for the verification process. Let's start with this shared mail address for now.

PR #214, opened 2025-02-19
ci: fix poetry caching

gaelgatelement:
Let's  see if https://github.com/matrix-org/setup-python-poetry/pull/23/files fixes our caching problem.

### DIFF
--- a/.github/workflows/releasing.yml
+++ b/.github/workflows/releasing.yml
@@ -4,7 +4,7 @@
 
 name: Helm Chart packaging and releasing
 on:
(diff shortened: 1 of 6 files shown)


.github/workflows/releasing.yml
```diff
@@ -4,7 +4,7 @@
 
 name: Helm Chart packaging and releasing
 on:
-  pull_request_target:
+  pull_request:
```
benbz, line 7: This was outstanding on merge

PR #215, opened 2025-02-19
Configure RageShakes and Location Sharing in Element Web

benbz:
e.g. https://github.com/element-hq/element-docker-demo/blob/main/data-template/element-web/config.json configures location sharing.

We also had an old request to configure RageShakes on ESS deploys. This is definitely appropriate for FOSS

### DIFF
--- a/charts/matrix-stack/README.md
+++ b/charts/matrix-stack/README.md
@@ -170,7 +170,9 @@ elementWeb:
 ```
 
 If `serverName` is set this will be configured in Element Web. If Synapse is enabled
-this will be configured in Element Web as well.
+this will be configured in Element Web as well. The Element RageShakes endpoint is
+configured so that bug reports can be submitted. The Element Map Tiler endpoint is
+also configured so that location sharing shows maps; please reconfigure as appropriate.
 
 Additional Element Web configuration can be provided as arbitrary sub-properties with
 ```yaml


charts/matrix-stack/templates/element-web/_helpers.tpl
```diff
@@ -36,6 +36,8 @@ app.kubernetes.io/version: {{ .image.tag }}
 {{- end }}
 {{- $defaultServerConfig := dict "m.homeserver" $mHomeserver -}}
 {{- $_ := set $config "default_server_config" $defaultServerConfig -}}
+{{- $_ := set $config "bug_report_endpoint_url" "https://element.io/bugreports/submit" -}}
+{{- $_ := set $config "map_style_url" "https://api.maptiler.com/maps/streets/style.json?key=fU3vlMsMn4Jb6dnEIFsx" -}}
```
gaelgatelement, line 40: Should we advertise it in the README?

benbz: Location Sharing or RageShakes or both?

gaelgatelement: Maybe both actually

benbz: Wording added

PR #218, opened 2025-02-19
Limit where LD_PRELOAD environment variable is set

AndrewFerr:
Set it only for the main synapse container, not also for the matrix-tools container.

### DIFF
--- a/newsfragments/218.changed.md
+++ b/newsfragments/218.changed.md
@@ -1,0 +1,1 @@
+Set the LD_PRELOAD environment variable only in containers that run Synapse.


charts/matrix-stack/templates/synapse/_helpers.tpl
```diff
@@ -89,6 +88,14 @@ app.kubernetes.io/version: {{ .image.tag }}
 {{- end -}}
 {{- end -}}
 
+{{- define "element-io.synapse.pythonEnv" }}
+{{- $root := .root -}}
+{{- with required "element-io.synapse.pythonEnv missing context" .context -}}
```
AndrewFerr, line 93: This doesn't apply any variables from `.extraEnv` or something similar, because this is meant to be used alongside `"element-io.synapse.env"`.  This is analogous to `"element-io.synapse.matrixToolsEnv"` which doesn't do that either.

If this is appropriate, maybe this doesn't even need to enforce a context being passed in?

charts/matrix-stack/templates/synapse/synapse_statefulset.yaml
```diff
@@ -162,9 +162,9 @@ We have an init container to render & merge the config for several reasons:
         - {{ include "element-io.synapse.process.app" (dict "root" $ "context" $processType) }}
         - "-c"
         - /conf/homeserver.yaml
-{{- with (include "element-io.synapse.env" (dict "root" $ "context" .)) }}
         env:
-          {{- . | nindent 10 }}
+          {{- include "element-io.synapse.pythonEnv" (dict "root" $ "context" .) | nindent 10 }}
+          {{- include "element-io.synapse.env" (dict "root" $ "context" .) | nindent 10 }}
```
AndrewFerr, line 167: This is analogous to how containers running `matrix-tools` set their env.

If there were other containers that run Python on a Synapse module, they would also need this.

The intent is to set `LD_PRELOAD` only for containers that actually run Python on a Synapse module.

PR #219, opened 2025-02-20
element web: additional as string

gaelgatelement:
This is to have the same interface between all components, including the ones which might use other config formats than YAML/JSON.

### DIFF
--- a/charts/matrix-stack/ci/pytest-element-web-values.yaml
+++ b/charts/matrix-stack/ci/pytest-element-web-values.yaml
@@ -10,18 +10,19 @@ elementWeb:
     {
       "default_server_config": {
         "m.homeserver": {
-          "base_url": "https://synapse.{{ $.Values.serverName }}",
+          "base_url": "https://synapse.{{ $.Values.serverName }}"
         }
       },
-      "element_call":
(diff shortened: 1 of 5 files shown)


charts/matrix-stack/source/element-web.json
```diff
@@ -7,7 +7,7 @@
       "type": "boolean"
     },
     "additional": {
-      "type": "object",
+      "type": "string",
       "additionalProperties": true
```
benbz:
```suggestion
```

charts/matrix-stack/ci/pytest-element-web-values.yaml
```diff
@@ -6,18 +6,23 @@
...
+      "element_call":
+        {"url": "https://call.{{ $.Values.serverName }}", "use_exclusively": true},
+      "features":
+        {
+          "feature_video_rooms": true,
+          "feature_group_calls": true,
+          "feature_new_room_decoration_ui": true,
+          "feature_element_call_video_rooms": true,
+        },
+    }
```
benbz: Can this get formatted consistently in the fragment

PR #221, opened 2025-02-20
postgres & synapse: fix bootstrap failing and causing restarts

gaelgatelement:
2 problems are fixed here : 
 1. The DB was actually initialized by the container `postgres-ess-updater` after the restart.
 2. HAProxy needs to become ready only after Synapse Main has been ready itself. Workers not ready must fallback to main.

### DIFF
--- a/charts/matrix-stack/templates/synapse/_synapse_details.tpl
+++ b/charts/matrix-stack/templates/synapse/_synapse_details.tpl
@@ -45,8 +45,8 @@ isSingle
 {{- define "element-io.synapse.process.canFallbackToMain" -}}
 {{- $root := .root -}}
 {{- with required "element-io.synapse.process.canFallbackToMain missing context" .context -}}
-{{ $hasPersistenceCookie := (list "media-repository" "sso-login") }}
-{{- if not (has . $hasPersistenceCookie) -}}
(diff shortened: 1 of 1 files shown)


charts/matrix-stack/templates/synapse/_synapse_details.tpl
```diff
@@ -42,6 +42,16 @@ isSingle
 {{- end -}}
 {{- end }}
 
+{{- define "element-io.synapse.process.canFallbackToMain" -}}
+{{- $root := .root -}}
+{{- with required "element-io.synapse.process.canFallbackToMain missing context" .context -}}
+{{ $hasPersistenceCookie := (list "media-repository" "sso-login") }}
+{{- if not (has . $hasPersistenceCookie) -}}
```
benbz:
```suggestion
{{ $cantFallBackToMain := (list "media-repository" "sso-login") }}
{{- if not (has . $cantFallBackToMain) -}}
```

tests/integration/fixtures/helm.py
```diff
@@ -176,25 +176,6 @@ async def matrix_stack(
     )
     await asyncio.gather(revision, *helm_prerequisites)
```
benbz: I think we should await `revision` separately and check the status still

PR #224, opened 2025-02-20
HAProxy: On startup, wait for media-repository & sso-login if they are deployed

gaelgatelement:
To reduce test flakes and have a semantically more correct configuration.


charts/matrix-stack/configs/synapse/partial-haproxy.cfg.tpl
```diff
@@ -10,6 +10,13 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 frontend startup
    bind *:8406
    acl synapse_dead nbsrv(synapse-main) lt 1
+{{- if dig "media-repository" "enabled" false $root.Values.synapse.workers }}
+  acl synapse_dead nbsrv(synapse-media-repository) lt 1
+{{- end }}
+{{- if dig "sso-login" "enabled" false $root.Values.synapse.workers }}
+  acl synapse_dead nbsrv(synapse-sso-login) lt 1
+{{- end }}
+
```
benbz: Can we iterate over the configured worker types and use `element-io.synapse.process.canFallbackToMain` to do this dynamically and so only have this behaviour encoded in 1 place?

gaelgatelement:
> I believe this creates an `OR` on all conditions and so does the right thing if any are dead

I think so as well, considering the example of the docs : 

https://docs.haproxy.org/3.1/configuration.html#4.2-monitor%20fail

PR #226, opened 2025-02-21
tests: improve endpoints status verification

gaelgatelement:
1. We need to give a bit of time for endpoints to be ready, our CI tests will flake
2. Ingress_ready fixture should not wait for all endpoints to be ready. On a live cluster, it will need only at least 1 backend to start responding, so this is what we must test here

### DIFF
--- a/tests/integration/test_networking.py
+++ b/tests/integration/test_networking.py
@@ -60,15 +60,15 @@ async def test_services_have_endpoints(
     generated_data: ESSData,
 ):
     async def _wait_for_endpoint_ready(name):
-        # We wait maximum 3 minutes for the endpoints to be ready
+        await asyncio.to_thread(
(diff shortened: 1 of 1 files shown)


tests/integration/test_networking.py
```diff
@@ -58,17 +59,30 @@ async def test_services_have_endpoints(
     kube_client: AsyncClient,
     generated_data: ESSData,
 ):
+    async def _wait_for_endpoint_ready(name):
+        # We wait maximum 3 minutes for the endpoints to be ready
+        start_time = time.time()
+        while time.time() - start_time < 180:
```
benbz: 3 mins feels excessive, shall we start with 30s?

gaelgatelement: Let's try!

PR #231, opened 2025-02-24
well-known: redirect to elementWeb chat app unless its a well-known path

gaelgatelement:


charts/matrix-stack/configs/well-known/partial-haproxy.cfg.tpl
```diff
@@ -13,6 +13,12 @@ frontend well-known-in
   # same as http log, with %Th (handshake time)
   log-format "%ci:%cp [%tr] %ft %b/%s %Th/%TR/%Tw/%Tc/%Tr/%Ta %ST %B %CC %CS %tsc %ac/%fc/%bc/%sc/%rc %sq/%bq %hr %hs %{+Q}r"
 
+{{ if $root.Values.elementWeb.enabled }}
+{{- with $root.Values.elementWeb }}
+{{- $elementWebHttps := include "element-io.ess-library.ingress.tlsSecret" (dict "root" $root "context" (dict "hosts" (list (required "elementWeb.ingress.host is required" .ingress.host)) "tlsSecret" .ingress.tlsSecret "ingressName" "element-web")) }}
+  http-request redirect  code 301  location http{{ if $elementWebHttps }}s{{ end }}://{{ .ingress.host }} unless well-known
```
benbz: This isn't going to do anything at present as the WKD `Ingress` only looks at `/.well-known/{matrix,element}`. I guess the Q is whether we want to take over that whole domain

gaelgatelement: I can't think of a reason not to?

benbz: > I can't think of a reason not to?

We take over the whole of the customer's domain if their ingresss controller supports multiple ingresses with the same hostname.

I'd be somewhat tempted to gate the whole domain behaviour but default it on.

gaelgatelement: OK I implement a take over of the whole domain if either `elementWeb` is enabled or `baseDomainRedirectURL` is set. Let me know what you think of it.

benbz: Will comment on the gating in the ingress.

Does this need to be below the `acl well-known` lines below as at this point the `well-known` ACL doesn't exist?

charts/matrix-stack/templates/well-known/ingress.yaml
```diff
@@ -21,6 +21,15 @@ spec:
   - host: "{{ tpl ($.Values.wellKnownDelegation.ingress.host | default $.Values.serverName) $ }}"
     http:
       paths:
+{{- if or $.Values.elementWeb.enabled .baseDomainRedirectURL }}
```
benbz: This works with what's in the HAProxy. I do wonder whether we need to do something like
```suggestion
{{- if and .redirectOnBaseDomain (or $.Values.elementWeb.enabled .baseDomainRedirectURL) }}
```

Or simpler 
```suggestion
{{- if and .redirectOnBaseDomain $.Values.elementWeb.enabled }}
```

`redirectOnBaseDomain` can default to `true` but allows people to opt-out of this behaviour but still have EW deployed.

We could leave this for now though and add it later if someone asks for it

PR #238, opened 2025-02-26
synapse: add a config check as helm hook

gaelgatelement:
When enabled (default behaviour), this should prevent downtime due to synapse misconfiguration.

### DIFF
--- a/charts/matrix-stack/templates/synapse/synapse_secret_hook.yaml
+++ b/charts/matrix-stack/templates/synapse/synapse_secret_hook.yaml
@@ -8,6 +8,7 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 {{- if and .enabled .checkConfigHook.enabled -}}
 apiVersion: v1
 kind: Secret
+type: Opaque
 metadata:
   labels:
     {{- include "element-io.synapse-check-config-hook.labels" (dict "root" $ "context" .) | nindent 4 }}
--- a/charts/matrix-stack/templates/synapse/synapse_secret.yaml
+++ b/charts/matrix-stack/templates/synapse/synapse_secret.yaml
@@ -8,6 +8,7 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 {{- if .enabled -}}
 apiVersion: v1
 kind: Secret
+type: Opaque
 metadata:
   labels:
     {{- include "element-io.synapse.labels" (dict "root" $ "context" .) | nindent 4 }}
--- a/tests/manifests/test_pod_containers_ports.py
+++ b/tests/manifests/test_pod_containers_ports.py
@@ -36,6 +36,7 @@ async def test_ports_in_containers_are_named(templates):
             assert len(port_names) == len(set(port_names)), f"Port names are not unique: {id}, {port_names}"
 
 
+@pytest.mark.parametrize("values_file", values_files_to_test)
 @pytest.mark.asyncio_cooperative
 async def test_no_ports_in_jobs(templates):
     for template in templates:


charts/matrix-stack/templates/synapse/synapse_configmap.yaml
```diff
@@ -13,6 +13,11 @@ metadata:
     {{- include "element-io.synapse.labels" (dict "root" $ "context" .) | nindent 4 }}
   name: {{ $.Release.Name }}-synapse
   namespace: {{ $.Release.Namespace }}
+{{- if .checkConfigHook.enabled }}
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-delete-policy": before-hook-creation
```
benbz: I'm reading https://helm.sh/docs/topics/charts_hooks/#hook-resources-are-not-managed-with-corresponding-releases (and also https://helm.sh/docs/howto/charts_tips_and_tricks/#tell-helm-not-to-uninstall-a-resource) and am thinking this is somewhat undesirable. As it is, this will delete the old config prior to running the hook. If the hook fails (because the new config is invalid) we can't restart the old Pods as the old config is gone. We can't / shouldn't replace this annotation with `helm.sh/resource-policy: keep` as then this will be orphan after uninstalling the chart. 

I suspect the correct thing to do is to have a 1 or 2 item loop - normal Synapse config + optionally config check config. They both contain the same data but the 2nd one gets the Helm hook annotations which shouldn't need a `helm.sh/hook-delete-policy` or a `helm.sh/resource-policy` annotation then as the normal lifecycle of Hook resources is fine. I suspect we've problems with hook resources in other places too

charts/matrix-stack/templates/synapse/synapse_check_config_job.yaml
```diff
@@ -0,0 +1,189 @@
...
+  podReplacementPolicy: TerminatingOrFailed
+  template:
+    metadata:
+      annotations:
+{{- with .annotations }}
+        {{- toYaml . | nindent 8 }}
+{{- end }}
+      labels:
+        {{- include "element-io.init-secrets.labels" (dict "root" $ "context" .) | nindent 8 }}
+    spec:
```
benbz: How tricky would it be to refactor this out into a common Synapse Pod Spec template?

charts/matrix-stack/templates/synapse/synapse_serviceaccount.yaml
benbz, line 1: Rather than editing this, I think the config check should have its own `ServiceAccount`

charts/matrix-stack/templates/synapse/synapse_check_config_job.yaml
gaelgatelement: This is now done, that turned out to be quite involved, let me know how the code looks.

charts/matrix-stack/templates/synapse/synapse_configmap_hook.yaml
```diff
@@ -0,0 +1,21 @@
...
+{{- with .Values.synapse -}}
+{{- if and .enabled .checkConfigHook.enabled -}}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  labels:
+    {{- include "element-io.synapse-hook.labels" (dict "root" $ "context" .) | nindent 4 }}
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-delete-policy": before-hook-creation
```
benbz:
```suggestion
```

Don't think this is needed

charts/matrix-stack/templates/synapse/synapse_secret_hook.yaml
```diff
@@ -0,0 +1,21 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  labels:
+    {{- include "element-io.synapse-hook.labels" (dict "root" $ "context" .) | nindent 4 }}
+  name: {{ include "element-io.synapse.secret-name" (dict "root" $ "context" (dict "isHook" true)) }}
+  namespace: {{ $.Release.Namespace }}
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-delete-policy": before-hook-creation
```
benbz:
```suggestion
```

charts/matrix-stack/templates/synapse/synapse_serviceaccount_hook.yaml
```diff
@@ -0,0 +1,29 @@
...
+Copyright 2025 New Vector Ltd
+
+SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+*/ -}}
+
+{{- with .Values.synapse -}}
+{{- if .enabled }}
+{{- with .checkConfigHook }}
+{{ if and .enabled .serviceAccount.create -}}
+apiVersion: v1
```
benbz: What's the blocker to using `element-io.ess-library.serviceAccount`?

charts/matrix-stack/templates/synapse/synapse_check_config_job_hook.yaml
```diff
@@ -0,0 +1,34 @@
...
+{{- $processType := "check-config-hook" }}
+{{- $perProcessRoot := merge (dict "processType" $processType "isHook" true) ($.Values.synapse | deepCopy) (.checkConfigHook | deepCopy)}}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ $.Release.Name }}-synapse-check-config-hook
+  namespace: {{ $.Release.Namespace }}
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-delete-policy": before-hook-creation
```
benbz:
```suggestion
```

charts/matrix-stack/templates/synapse/_helpers.tpl
```diff
@@ -16,6 +16,18 @@ k8s.element.io/synapse-instance: {{ $root.Release.Name }}-synapse
 {{- end }}
 {{- end }}
 
+{{- define "element-io.synapse-hook.labels" -}}
```
```suggestion
{{- define "element-io.synapse-check-config-hook.labels" -}}
```

Possibly

tests/manifests/utils.py
```diff
@@ -159,22 +159,23 @@ async def _make_templates(values):
     return _make_templates
 
 
-def iterate_component_parts(component, values, setter, if_condition):
+def iterate_component_parts(component, values, setter, if_condition, force):
```
benbz: Can we have a more descriptive argument name than `force` as it doesn't make the caller in `tests/manifests/test_serviceaccounts.py` clear about why it is set

tests/manifests/test_service_monitors.py
```diff
@@ -11,7 +11,7 @@
 
 
 def selector_match(labels: dict[str, str], selector: dict[str, str]) -> bool:
-    return all(labels[key] == value for key, value in selector.items())
+    return all(labels.get(key, "") == value for key, value in selector.items())
```
benbz: Why's this changing?

charts/matrix-stack/templates/synapse/synapse_serviceaccount_hook.yaml
```diff
@@ -0,0 +1,15 @@
...
+Copyright 2025 New Vector Ltd
+
+SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+*/ -}}
+
+{{- with .Values.synapse -}}
+{{- if .enabled }}
+{{- with .checkConfigHook }}
+{{ if and .enabled -}}
+{{- include "element-io.ess-library.serviceAccount" (dict "root" $ "context" (dict "componentValues" . "key" "synapse-check-config-hook" "extraAnnotations" (dict "helm.sh/hook" "pre-install,pre-upgrade"))) }}
```
benbz: Needs the weight annotation?

charts/matrix-stack/templates/synapse/synapse_check_config_job_hook.yaml
```diff
@@ -0,0 +1,34 @@
...
+{{- $processType := "check-config-hook" }}
+{{- $perProcessRoot := merge (dict "processType" $processType "isHook" true) ($.Values.synapse | deepCopy) (.checkConfigHook | deepCopy)}}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ $.Release.Name }}-synapse-check-config-hook
+  namespace: {{ $.Release.Namespace }}
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+
```
benbz: Needs the weight annotation?

charts/matrix-stack/templates/synapse/synapse_serviceaccount_hook.yaml
gaelgatelement: I'm using weights for things that need to happen before the synapse check config job. The job has the default weight. Do you think we need something more here?

benbz: Both the config map and secret have the weight annotation. I'd prefer to see all resources for a given a hook at the same weight (or no weight). If they don't need the weight annotation, I'm happy for everything to be removed

gaelgatelement: They need to have a smaller weight than the job, or the job fails to start because its configmap & secrets are not existing. 

Actually, withouth the weight, the job fails to start and the configmap/secret never appear : 

```
│ Message:             MountVolume.SetUp failed for volume "secret-pytest-uvyx2moh-synapse-hook" : secret "pytest-uvyx2moh-synapse-hook" not found                                                                                    │
│ Message:             MountVolume.SetUp failed for volume "plain-config" : configmap "pytest-uvyx2moh-synapse-hook" not found                                                                                                        │
```

benbz: That feels strange
> The library sorts hooks by weight (assigning a weight of 0 by default), by resource kind and finally by name in ascending order.

I took the reference to sorting by kind to mean the standard Helm order of kinds and not alphabetical or anything sillier

gaelgatelement: I found a similar issue on SO, but I agree that the docs are rather confusing.. :/ https://stackoverflow.com/questions/59423084/kubernetes-use-secrets-on-pre-install-job 

On my tests, without weights, I cant get it to work.

benbz: :+1: let's set explicit weights then and comment why the Job has a higher weight than the CM/Secret

PR #254, opened 2025-03-05
element web: support multiple additional subproperties

gaelgatelement:
Let's use a common interface in all components.

### DIFF
--- a/charts/matrix-stack/templates/element-web/_helpers.tpl
+++ b/charts/matrix-stack/templates/element-web/_helpers.tpl
@@ -41,10 +41,10 @@ app.kubernetes.io/version: {{ .image.tag }}
 {{- with .additional }}
 {{- range $key := (. | keys | uniq | sortAlpha) }}
 {{- $prop := index $root.Values.elementWeb.additional $key }}
-{{- $_ := (merge $config ($prop | fromJson)) -}}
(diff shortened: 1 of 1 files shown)


charts/matrix-stack/templates/element-web/_helpers.tpl
```diff
@@ -38,7 +38,13 @@ app.kubernetes.io/version: {{ .image.tag }}
...
 {{- $_ := set $config "bug_report_endpoint_url" "https://element.io/bugreports/submit" -}}
 {{- $_ := set $config "map_style_url" "https://api.maptiler.com/maps/streets/style.json?key=fU3vlMsMn4Jb6dnEIFsx" -}}
-{{- tpl (toPrettyJson (merge $config (.additional | fromJson))) $root -}}
+{{- with .additional }}
+{{- range $key := (. | keys | uniq | sortAlpha) }}
+{{- $prop := index $root.Values.elementWeb.additional $key }}
+{{- $_ := (merge $config ($prop | fromJson)) -}}
+{{- end }}
+{{- end }}
+{{- tpl (toPrettyJson $config) $root -}}
```
benbz:
```suggestion
{{- $_ := (merge $config ((tpl $prop $root) | fromJson)) -}}
{{- end }}
{{- end }}
{{- toPrettyJson $config -}}
```

I wonder if we should do this instead incase multiple fragments merge into the same JSON keys

PR #260, opened 2025-03-07
tests: verify that mounted secrets are actually used explicitely

gaelgatelement:
Let's avoid to mount unused secrets in containers, to reduce leak potential.

### DIFF
--- a/tests/manifests/__init__.py
+++ b/tests/manifests/__init__.py
@@ -163,14 +163,14 @@ all_components_details = [
     ComponentDetails(
         name="postgres",
         has_ingress=False,
-        paths_consistency_noqa=("/docker-entrypoint-initdb.d",),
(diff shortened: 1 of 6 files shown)


tests/manifests/__init__.py
```diff
@@ -163,14 +163,14 @@ def should_visit_with_values(
     ComponentDetails(
         name="postgres",
         has_ingress=False,
-        paths_consistency_noqa=("/docker-entrypoint-initdb.d",),
+        paths_consistency_noqa=("/docker-entrypoint-initdb.d", "/docker-entrypoint-initdb.d/init-ess-dbs.sh"),
```
benbz: Both this addition and the one below are extensions of the original paths. Can we get rid of the original paths? Could we make usage of `paths_consistency_noqa` `startswith` rather than exact string match?

gaelgatelement: Indeed, we dont need to check for the parent path anymore.

PR #266, opened 2025-03-10
element web: disable sso immediate

gaelgatelement:
Feedback of install tests is that current behaviour is rather confusing because the MAS UI page does not have any logo or branding.


newsfragments/266.changed.md
```diff
@@ -0,0 +1 @@
+element-web & mas: disable `sso_immediate`.
```
benbz:
```suggestion
Disable immediate redirect to Matrix Authentication Service in Element Web.
```

PR #274, opened 2025-03-10
manifests tests: handle noqa at the mount key level.

gaelgatelement:
By mapping parents paths to mounted keys, we can check if we should noqa the parent based on the mounted key which it relates to.

### DIFF
--- a/tests/manifests/test_configs_and_mounts_consistency.py
+++ b/tests/manifests/test_configs_and_mounts_consistency.py
@@ -105,6 +105,7 @@ def get_mounts_part(secret_or_cm, volume_mount):
 
 def get_keys_from_container_using_rendered_config(template, templates, other_secrets):
     mounted_keys = []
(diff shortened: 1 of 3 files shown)


tests/manifests/test_configs_and_mounts_consistency.py
```diff
@@ -166,9 +169,9 @@ async def test_secrets_consistency(templates, other_secrets, template_to_deploya
         for container in containers:
             # Determine which secrets are mounted by this container
             mounted_keys = []
```
benbz, line 171: I don't think we need to track this separately as we can get it with `mounted_keys_to_parents.keys()`?

gaelgatelement: I track this separately so that we can run the assertion that mounted keys are unique. I could also do it by making sure the key does not exist when updating the dict, but I think the error message would be less useful.

PR #276, opened 2025-03-12
ci: fix pr release not suffixing with commit hash

gaelgatelement:
All PRs were overwriting the `-dev` artifact.


benbz:
I wonder if we should do both events given how often we've flipped backwards and forwards between the 2 triggers.

Edit I've just seen https://github.com/element-hq/ess-helm/pull/277 which does exactly this :p

PR #281, opened 2025-03-14
Ingress Controller Type automatic handling

gaelgatelement:
This adds ingress.controllerType to have automatic behaviour implemented depending on the ingress-controller used.

### DIFF
--- a/charts/matrix-stack/source/common/ingress.json
+++ b/charts/matrix-stack/source/common/ingress.json
@@ -22,7 +22,7 @@
     "controllerType": {
       "type": "string",
       "enum": [
-        "kubernetes-nginx"
+        "ingress-nginx"
       ]
     },
     "service": {
--- a/charts/matrix-stack/ci/fragments/well-known-pytest-extras.yaml
+++ b/charts/matrix-stack/ci/fragments/well-known-pytest-extras.yaml
@@ -9,7 +9,7 @@ global:
 serverName: "{{ $.Values.global.baseDomain }}"
 
 ingress:
-  controllerType: kubernetes-nginx
+  controllerType: ingress-nginx
 
 wellKnownDelegation:
   ingress:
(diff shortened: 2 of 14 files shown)


charts/matrix-stack/source/common/ingress.json
```diff
@@ -19,6 +19,12 @@
     "tlsSecret": {
       "type": "string"
     },
+    "type": {
+      "type": "string",
+      "enum": [
+        "kubernetes-nginx"
```
benbz: At the risk of bike-shedding can we use a different value here: `nginx` or `ingress-nginx` would be my suggestions

gaelgatelement: I fear it is confusing with https://docs.nginx.com/nginx-ingress-controller/

Whereas I think "kubernetes-nginx" points explicitely to https://github.com/kubernetes/ingress-nginx

benbz: Happy to rule out `nginx` on those grounds but https://github.com/kubernetes/ingress-nginx is explicitly named `ingress-nginx` and is even how the migration docs for the F5 Nginx Controller refers to it (https://docs.nginx.com/nginx-ingress-controller/installation/ingress-nginx/). If we need to support the other, I'd add it as either `nginx-ingress` or `f5-nginx-ingress`. I think the values file should probably link to the relevant project/docs page so that there's no ambiguity.

benbz: FTR both Gitlab (https://gitlab.com/gitlab-org/charts/gitlab/-/blob/master/charts/gitlab/charts/webservice/templates/_nginx.tpl#L23) and Posthog (https://github.com/adutchak/charts-clickhouse/blob/main/charts/posthog/templates/ingress.yaml#L21) use `nginx` for their controller type value. As said I'm happy to avoid using `nginx` ourselves but it does show there's clear mindshare for https://github.com/kubernetes/ingress-nginx over https://docs.nginx.com/nginx-ingress-controller/

PR #304, opened 2025-03-20
Update licenses to AGPLv3

gaelgatelement:


.github/workflows/matrix-tools.yml
```diff
@@ -87,9 +87,9 @@ jobs:
           type=raw,value=${{ env.VERSION }},enable=${{ github.ref_type == 'tag' }}
           type=sha,prefix=sha-,format=short
         labels: |
-          org.opencontainers.image.licenses=AGPL-3.0-only OR LicenseRef-Element-Commercial
+          org.opencontainers.image.licenses=AGPL-3.0-only
         annotations: |
-          org.opencontainers.image.licenses=AGPL-3.0-only OR LicenseRef-Element-Commercial
+          org.opencontainers.image.licenses=AGPL-3.0-only
```
benbz, line 92: These aren't needed anymore as GitHub will correctly set them. It was only for our commercial license that it didn't return the correct SPDX info

PR #307, opened 2025-03-20
MAS: drop init container

gaelgatelement:
@sandhose How does MAS handle IdP changes nowadays ? 

We used to keep this init container as so to have `prune` available. But it feels dangerous as in case of misconfiguration, all the oauth IdP links are lost.

What would be your recommandations nowadays?

### DIFF
--- a/charts/matrix-stack/templates/matrix-authentication-service/deployment.yaml
+++ b/charts/matrix-stack/templates/matrix-authentication-service/deployment.yaml
@@ -127,33 +127,6 @@ spec:
         resources:
           {{- toYaml . | nindent 10 }}
 {{- end }}
-{{- /* This init container job is kept despite https://github.com/matrix-org/matrix-authentication-service/pull/2432 so as to have --prune */}}
-      - name: config
-        args: ["config", "sync", "--prune"]
-{{- with .image -}}
-{{- if .digest }}
-        image: "{{ .registry }}/{{ .repository }}@{{ .digest }}"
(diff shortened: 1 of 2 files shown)


sandhose:
> How does MAS handle IdP changes nowadays ?

Nowadays we 'soft-delete' upstream IDPs, so they stay in the database, but get disabled

charts/matrix-stack/templates/matrix-authentication-service/deployment.yaml
```diff
@@ -127,33 +127,6 @@ spec:
         resources:
           {{- toYaml . | nindent 10 }}
 {{- end }}
-{{- /* This init container job is kept despite https://github.com/matrix-org/matrix-authentication-service/pull/2432 so as to have --prune */}}
-      - name: config
-        args: ["config", "sync", "--prune"]
```
jaywink, line 132: We don't need `config sync` anymore either?

sandhose: No this is done on startup, unless `--no-sync` is set

Although it could be interesting if we did a pre-upgrade Job which does a `config sync`, as it also migrates the database. This way, we do all the heavy database operations before rolling out the deployment?

gaelgatelement: No, see the pull request : this is now done on standard MAS container. We just kept this to be able to `--prune`.

PR #308, opened 2025-03-24
Don't allow matrix-org/element-hq floating tags for GitHub actions

benbz:
After #305 we have no more floating tags/refs so don't allow them at all


benbz:
https://github.com/element-hq/ess-helm/actions/runs/14036607440/job/39296015828?pr=308 (`pull_request`) is successful, removing the `DROPME` commit

PR #332, opened 2025-03-30
MAS: Fix EX login

gaelgatelement:
EX started failing to login against MAS due to recent OIDC MSC changes with error `denied by the policy: [Violation { msg: "missing contacts", field: None }]`.


gaelgatelement:
This is not necessary actually as ESS is properly on MAS version  0.13+.

PR #334, opened 2025-03-31
artifacthub: fix regex + package url / repo ID

gaelgatelement:
https://artifacthub.io/packages/helm/element-server-suite-community/matrix-stack

This fixes the exclusion of all dev versions : 

image

### DIFF
--- a/artifacthub-repo.yaml
+++ b/artifacthub-repo.yaml
@@ -2,10 +2,10 @@
 #
 # SPDX-License-Identifier: AGPL-3.0-only
 
-repositoryID: f900b890-eb67-4328-9100-336301f7cefe
+repositoryID: 6a910e71-6940-45a6-9d63-07ec0080aa55
 owners:
   - name: Element Server Products Team
     email: onprem-support@element.io
 ignore:
   - name: matrix-stack
-    version: "[0-9]+\\.[0-9]+\\.[0-9]+.+"
+    version: "[0-9]+\\.[0-9]+\\.[0-9]+-sha.+"
--- a/artifacthub-repo.yaml
+++ b/artifacthub-repo.yaml
@@ -8,4 +8,4 @@ owners:
     email: onprem-support@element.io
 ignore:
   - name: matrix-stack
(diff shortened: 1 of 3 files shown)


artifacthub-repo.yaml
```diff
@@ -8,4 +8,4 @@ owners:
     email: onprem-support@element.io
 ignore:
   - name: matrix-stack
-    version: "[0-9]+\\.[0-9]+\\.[0-9]+\\..+"
+    version: "[0-9]+\\.[0-9]+\\.[0-9]+\\-.+"
```
benbz: Does `-` need to be escaped here given it isn't include a character class? Should we also put the `sha` in the regex too?

gaelgatelement: I suppose I can simplify this and do `[0-9]+\\.[0-9]+\\.[0-9]+.+` to avoid any ambiguity

benbz: But this blocks us doing anything like `0.7.1.1` or `0.7.1-reissue1`. Also depending on greediness this may ignore double digit patch releases

gaelgatelement: OK, this may be something desirable indeed.

artifacthub-repo.yaml
```diff
@@ -2,10 +2,10 @@
...
 
-repositoryID: f900b890-eb67-4328-9100-336301f7cefe
+repositoryID: 6a910e71-6940-45a6-9d63-07ec0080aa55
 owners:
   - name: Element Server Products Team
     email: onprem-support@element.io
 ignore:
   - name: matrix-stack
-    version: "[0-9]+\\.[0-9]+\\.[0-9]+\\..+"
+    version: "[0-9]+\\.[0-9]+\\.[0-9]+-sha.+"
```
benbz:
```suggestion
    version: "[0-9]+\\.[0-9]+\\.[0-9]+-(sha.+|dev)"
```

We've got to hide the legacy `-dev` artifacts too yes?

benbz (approved):

PR #336, opened 2025-04-01
Fix MAS/Synapse when using something else than init secrets

gaelgatelement:
This also adds a test to anticipate such mishap in the future.

The mount path regex for reference does match like this : 
image

### DIFF
--- a/charts/matrix-stack/configs/haproxy/haproxy.cfg.tpl
+++ b/charts/matrix-stack/configs/haproxy/haproxy.cfg.tpl
@@ -64,7 +64,7 @@ defaults
   option redispatch
 
   compression algo gzip
-  compression type text/plain text/html text/xml application/json text/css  # noqa
+  compression type text/plain text/html text/xml application/json text/css
 
   # if we hit the maxconn on a server, and the queue timeout expires, we want
   # to avoid returning 503, since that will cause cloudflare to mark us down.
--- a/charts/matrix-stack/configs/haproxy/haproxy.cfg.tpl
+++ b/charts/matrix-stack/configs/haproxy/haproxy.cfg.tpl
@@ -64,7 +64,7 @@ defaults
   option redispatch
 
   compression algo gzip
-  compression type text/plain text/html text/xml application/json text/css
+  compression type text/plain text/html text/xml application/json text/css  # noqa
 
   # if we hit the maxconn on a server, and the queue timeout expires, we want
   # to avoid returning 503, since that will cause cloudflare to mark us down.
--- a/tests/manifests/__init__.py
+++ b/tests/manifests/__init__.py
@@ -163,7 +163,7 @@ all_components_details = [
     ComponentDetails(
         name="postgres",
         has_ingress=False,
-        paths_consistency_noqa=("/docker-entrypoint-initdb.d/init-ess-dbs.sh", "/var/lib/postgres/data/pgdata"),
+        paths_consistency_noqa=("/docker-entrypoint-initdb.d/init-ess-dbs.sh"),
         is_shared_component=True,
     ),
     ComponentDetails(
@@ -188,7 +188,6 @@ all_components_details = [
         additional_values_files=[
             "synapse-worker-example-values.yaml",
         ],
-        paths_consistency_noqa=["/media/media_store"],
         sub_components=[
             SubComponentDetails(
                 name="synapse-redis",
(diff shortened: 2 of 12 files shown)


tests/manifests/__init__.py
```diff
@@ -163,30 +163,32 @@ def should_visit_with_values(
     ComponentDetails(
         name="postgres",
         has_ingress=False,
-        paths_consistency_noqa=("/docker-entrypoint-initdb.d/init-ess-dbs.sh"),
+        paths_consistency_noqa=("/docker-entrypoint-initdb.d/init-ess-dbs.sh", "/var/lib/postgres/data/pgdata"),
```
benbz: Could we enhance the test_configs_and_mounts setup to be aware of PVCs?

tests/manifests/__init__.py
```diff
@@ -163,30 +163,32 @@ def should_visit_with_values(
...
-        ),
+        shared_component_names=("init-secrets", "postgres"),
     ),
     ComponentDetails(
         name="synapse",
         has_db=True,
         additional_values_files=[
             "synapse-worker-example-values.yaml",
         ],
+        paths_consistency_noqa=["/media/media_store"],
```
benbz: Could we enhance the test_configs_and_mounts_setup to be aware of PVCs + emptyDirs?

Edit: oh there appears to be emptyDir support, so just PVCs are outstanding?

tests/manifests/test_configs_and_mounts_consistency.py
```diff
@@ -54,6 +55,39 @@ def get_volume_from_mount(template, volume_mount):
...
+
+
+def find_paths_in_contents(container, mounted_config_maps):
+    paths_found = []
+    content_to_match = [e.get("value", "") for e in container.get("env", [])]
+    content_to_match += [c for c in container.get("command", [""])[1:] + container.get("args", [])]
+    for cm in mounted_config_maps:
+        for key, content in cm["data"].items():
+            if key.endswith(".yaml"):
+                content_to_match += get_values(yaml.safe_load(content))
```
benbz: How essential is being able to parse the CM files vs just treating the CM data as arbitrary strings?

tests/manifests/test_configs_and_mounts_consistency.py
```diff
@@ -54,6 +55,39 @@ def get_volume_from_mount(template, volume_mount):
...
+
+    for content in content_to_match:
+        assert type(content) is str, f"Content must be a string: {content}"
+        for match_in in content.split("\n"):
+            for exclude in ["://", "/bin/sh", "helm.sh/"]:
+                if exclude in match_in:
+                    break
+            else:
+                for match in re.findall(r"((?<![0-9:])/[^\s\n\")`:%;,/]+[^\s\n\")`:%;,]+(?!.*noqa))", match_in):
```
benbz: Could this have a comment as it looks like my cat walked over the keyboard in places :black_cat:

tests/manifests/test_configs_and_mounts_consistency.py
gaelgatelement: Treating the CM data as arbitrary strings works fine, but matches a lot of undesirable things, especially in haproxy config map where we have a lot of url paths due to url matching.

It would involve writing a lot of noqa.

benbz: Makes sense. How involved would it be to enhance the noqa setup to ignore lines matching regexes (`^\s*acl.*` would cover a lot of the HAProxy config I'd have thought)?

gaelgatelement: In the end, I ended up adding more noqa, and a new variable to skip a whole file.

charts/matrix-stack/configs/haproxy/haproxy.cfg.tpl
```diff
@@ -64,7 +64,7 @@ defaults
   option redispatch
 
   compression algo gzip
-  compression type text/plain text/html text/xml application/json text/css
+  compression type text/plain text/html text/xml application/json text/css  # noqa
```
benbz: I don't feel this should be a potential path as `[a-z]/` isn't an absolute path

gaelgatelement: It finds `/css` ... But yes I can enhance this regex.

benbz (approved):

PR #345, opened 2025-04-03
synapse: fix APPLICATION_NAME template

gaelgatelement:
Minor template issue but could case persisters to loop replicate requests to themselves.


benbz (approved):
This looks like a syntactically valid change. I don't understand the why though

gaelgatelement:
> This looks like a syntactically valid change. I don't understand the why though

In `instances_map` we have workers named as `{{ $processType }}-{{ $idx }}`. The worker name and the name in the instances map have to match.
The hostnames of the pods are `{{ .Release.Name }}-synapse-{{ $ProcessType }}-[0-9]+`. So we need to trim both `-synapse` and `{{ $.Release.Name}}` from their hostname to discover their proper worker name.

PR #347, opened 2025-04-03
Simple install script for ESS CE.

rxl881:
A simple install script for ESS community edition. Although it works for my use-case, it will need further work for general use. For example it currently only supports installation using wildcard certs (but the other cases can be simply added to the switch statement).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -6,10 +6,12 @@
 
 # This script installs the Element Server Suite (ESS) Community Edition.
 # It depends on the following tools:
-# - kubectl: Command-line tool for interacting with Kubernetes clusters.
 # - curl: Command-line tool for transferring data with URLs.
 # - dig: DNS lookup utility for querying DNS records.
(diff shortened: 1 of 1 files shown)


benbz:
Hi Rick, thanks for providing this initial PR of your install script. As you say further work would be required to get this merged for general use. (Not a full review) Some initial things that that stand out to me as being needed before we can thank about merging it:
* Added to the documentation
* Moved into the `scripts` folder and renamed to distinguish it from `scripts/setup_test_cluster.sh` (`scripts/setup_k3s_and_install.sh`?)
* Needs to handle `matrixRTC` given that's now in the quick setup values files since this PR was created
* How safe is the script to re-run, appreciating that while you've got `y/N` around various steps, people will still blindly hit `y`?
* Implementing the individual certs case would be good
  * This might be easier using `yq` rather than `sed`
* I don't think there's anything much to do for the external reverse proxy setup so not bailing out would be good

PR #352, opened 2025-04-05
Add script to extract Synapse configuration values

### LINKED ISSUE #340: Add Migration Guide for different kinds of legacy matrix installations
Hi again,

as user of some kind of custom docker / compose based matrix install, I would be interested in some specific and generic migration guides that help me and others to switch from i variety of installations to this one...

sincerely,
rico

hanthor:
Fixes #340 

Will need to make the migration docs split out for MDAD specific or more generic isntructions

This only supports migrating Synapse it pulls all values from homeserver.yaml

### DIFF
--- a/scripts/extract-synapse-values.sh
+++ b/scripts/extract-synapse-values.sh
@@ -1,0 +1,108 @@
+#!/usr/bin/env bash
+
+# Copyright 2025 New Vector Ltd
+#
+# SPDX-License-Identifier: AGPL-3.0-only
+
+set -xe
+
+
+targetDirectory="$1"
+homeserverconfig="$2"
+
+
+if [ -z "$targetDirectory" ]; then
+  echo "Please provide a valid path for the ESS-Helm values directory"
+  echo "Usage: $0 <target_directory> <homeserver.yaml>"
+  echo "Example: $0 ~/ess-helm-values /matrx/synapse/config/homeserver.yaml"
+  exit 1
+fi
+
+if [ -z "$homeserverconfig" ]; then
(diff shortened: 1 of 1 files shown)


scripts/extract-synapse-values.sh
```diff
@@ -0,0 +1,108 @@
...
+if [ ! -d "$targetDirectory" ]; then
+  mkdir -p "$targetDirectory"
+fi
+
+touch "$targetDirectory/hostnames.yaml"
+touch "$targetDirectory/secrets.yaml"
+
+servername=$(yq '.servername' "$homeserverconfig")
+synapseHost=$(yq '(.public_baseurl | match "https://(.+)/").captures[0].string' "$homeserverconfig")
+macaroon=$(yq '.macaroon_secret_key' "$homeserverconfig")
```
gaelgatelement, line 59: Synapse supports  `_path` suffixes for all secrets nowadays https://element-hq.github.io/synapse/latest/usage/configuration/config_documentation.html?highlight=macaroon#macaroon_secret_key_path . It would be great if the script could handle those automatically.

scripts/extract-synapse-values.sh
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env bash
+
+# Copyright 2025 New Vector Ltd
+#
+# SPDX-License-Identifier: AGPL-3.0-only
+
+set -xe
+
+
+targetDirectory="$1"
```
gaelgatelement, line 10: Can you update the migration docs with your script usage please?

PR #359, opened 2025-04-08
Fix checkov run

gaelgatelement:
Because of https://github.com/bridgecrewio/checkov/issues/5643

Checkov is skipping resources which have '{{'. This is our case when we rely on `render-config` templating capacities.

Let's instead generate templates locally, and drop the '{{' syntax automatically.


.github/workflows/build-test.yml
```diff
@@ -4,7 +4,7 @@
 
 name: Helm Chart Building tests
 on:
-  pull_request_target:
+  pull_request:
```

.github/workflows/build-test.yml
```diff
@@ -94,7 +94,12 @@ jobs:
         export HELM_NAMESPACE=ess
         for checkov_values in charts/matrix-stack/ci/*checkov*values.yaml; do
           echo "Testing matrix-stack with $checkov_values";
-          checkov -d "charts/matrix-stack" --framework helm --var-file "$checkov_values" --quiet
+          mkdir checkov-tmp
+          cd checkov-tmp
+          helm template checkov -f "../$checkov_values" ../charts/matrix-stack | sed 's/{{//' | yq -s '.kind + "-" + .metadata.name'
+          checkov -d .  --skip-check CKV_SECRET_6 --quiet
```
benbz: We've got suppression annotations in various files, do we also need `--skip-check` here.

.github/workflows/build-test.yml
```diff
@@ -94,7 +94,12 @@ jobs:
         export HELM_NAMESPACE=ess
         for checkov_values in charts/matrix-stack/ci/*checkov*values.yaml; do
           echo "Testing matrix-stack with $checkov_values";
-          checkov -d "charts/matrix-stack" --framework helm --var-file "$checkov_values" --quiet
+          mkdir checkov-tmp
```
benbz: Add to `.gitignore` or otherwise I think we should use `mktemp -d` so that we don't get a folder in this directory

charts/matrix-stack/source/matrixAuthenticationService.yaml.j2
```diff
@@ -25,7 +25,7 @@ privateKeys:
 {{ sub_schema_values.credential("ECDSA Secp384r1 Private Key", "ecdsaSecp384r1") | indent(2) }}
 
 {{ sub_schema_values.ingress() }}
-
+{{- sub_schema_values.resources(requests_memory='50Mi', requests_cpu='50m', limits_memory='350Mi') }}
```
benbz: All the other sub-schemas here are missing `-`. Add to all or remove `-` from here

.github/workflows/build-test.yml
gaelgatelement: `CKV_SECRET_6` applies to all resources (including ConfigMaps, for which we dont have annotations)

I'm fine in keeping skip-check annotations per-workload, but we can also skip the check globally. In which case we probably don't need checkov values files anymore. What do you think?

benbz: Let's comment here why we're skipping globally and then remove the individual skips from the values files.

However all these changes make it harder to test locally as we can't just run `checkov` anymore. I suspect we should extract a script that takes a values file path and does all this

benbz: Sorry I think there's been some confusion, I was only meaning remove the `checkov.io/skip4: CKV_SECRET_6=Postgres annotations hash are triggering CKV_SECRET_6.` skips as everything else can be done on a targetted per-workload basis.

scripts/checkov.sh
```diff
@@ -0,0 +1,24 @@
...
+#
+# SPDX-License-Identifier: AGPL-3.0-only
+
+set -euo pipefail
+
+checkov --version
+export HELM_NAMESPACE=ess
+workdir="$PWD"
+matrix_stack_path="$workdir/charts/matrix-stack"
+for checkov_values in charts/matrix-stack/ci/*checkov*values.yaml; do
```
benbz: I think this script shouldn't do the looping but take a specific values file and test that

.github/workflows/build-test.yml
```diff
@@ -94,7 +94,11 @@ jobs:
         export HELM_NAMESPACE=ess
         for checkov_values in charts/matrix-stack/ci/*checkov*values.yaml; do
           echo "Testing matrix-stack with $checkov_values";
-          checkov -d "charts/matrix-stack" --framework helm --var-file "$checkov_values" --quiet
+          tmpdir=$(mktemp -d)
+          cd "$tmpdir"
+          helm template checkov -f "../$checkov_values" ../charts/matrix-stack | sed 's/{{//' | yq -s '.kind + "-" + .metadata.name'
+          checkov -d .  --skip-check CKV_SECRET_6 --quiet
```
benbz: We should be using the script here rather than implementing things twice

gaelgatelement: :facepalm:

benbz (approved):

PR #371, opened 2025-04-15
MAS: Allow to setup without enabling auth delegation

gaelgatelement:
This will be helpful in the context of Synapse to MAS migrations.


charts/matrix-stack/ci/matrix-authentication-service-keep-auth-in-synapse.yaml
```diff
@@ -0,0 +1,70 @@
...
+#
+# source_fragments: matrix-authentication-service-minimal.yaml matrix-authentication-service-auth-synapse.yaml matrix-authentication-service-test-postgres.yaml synapse-minimal.yaml synapse-test-postgres.yaml
+# DO NOT EDIT DIRECTLY. Edit the fragment files to add / modify / remove values
+
+# initSecrets, postgres don't have any required properties to be set and defaults to enabled
+elementWeb:
+  enabled: false
+matrixAuthenticationService:
+  encryptionSecret:
+    secret: '{{ $.Release.Name }}-mas-external'
```
benbz: This doesn't look up to date with `source_fragments` to me. Slightly concerned CI isn't catching this

gaelgatelement: That's because the file does not end with values...

benbz: CI job that checks all the filenames in `charts/matrix-stack/ci` itself end in `-values.yaml`?

PR #373, opened 2025-04-16
Test Chart Version Change and pods idempotency

gaelgatelement:
We want to avoid causing restarts on version changes.


tests/manifests/test_pod_idempotency.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+@pytest.mark.parametrize("values_file", ["element-web-minimal-values.yaml"])
+@pytest.mark.asyncio_cooperative
+async def test_values_file_renders_idempotent_pods(release_name, values, helm_client, temp_chart):
+    async def _patch_version_chart():
+        with open(f"{temp_chart}/Chart.yaml") as f:
+            chart = yaml.safe_load(f)
+        with open(f"{temp_chart}/Chart.yaml", "w") as f:
+            chart["version"] = f"{random.randint(0, 100)}.{random.randint(0, 100)}.0"
```
benbz: There's a chance this could match what's already there. Add a suffix or interpret the chart version and explicitly increment?

tests/manifests/test_pod_idempotency.py
```diff
@@ -0,0 +1,42 @@
...
+            yaml.dump(chart, f)
+        return await helm_client.get_chart(temp_chart)
+
+    first_render = {}
+    second_render = {}
+    for template in await helm_template(
+        (await _patch_version_chart()), release_name, values, has_service_monitor_crd=True, skip_cache=True
+    ):
+        if template["kind"] in ["Deployment", "StatefulSet", "Job"]:
```
benbz: I think we should be testing everything. We should test that all manifests aren't equal by default and that if we remove `metadata.labels['helm.sh/chart']` all manifests are now equal. That proves that the label isn't on the Pod spec and it also proves that all hashes (be it on the Pod spec or the parent) are avoiding including the full manifest

charts/matrix-stack/templates/element-web/deployment.yaml
```diff
@@ -29,9 +29,11 @@ spec:
...
     metadata:
       labels:
-        {{- include "element-io.element-web.labels" (dict "root" $ "context" .) | nindent 8 }}
-        k8s.element.io/confighash: "{{ include (print $.Template.BasePath "/element-web/configmap.yaml") $ | sha1sum }}"
-        k8s.element.io/nginxhash: "{{ include (print $.Template.BasePath "/element-web/nginx_configmap.yaml") $ | sha1sum }}"
+        {{- include "element-io.element-web.labels" (dict "root" $ "context" (dict "image" .image "labels" .labels "withChartVersion" false)) | nindent 8 }}
+        k8s.element.io/confighash: "{{( tpl ($.Files.Get "configs/element-web/config.json.tpl") (dict "root" $ "context" .)) | sha1sum }}"
+        k8s.element.io/nginxhash: "{{ (cat ($.Files.Get "configs/element-web/default.conf")
+                                            ($.Files.Get "configs/element-web/http_customisations.conf")
+                                            ($.Files.Get "configs/element-web/security_headers.conf"))  | sha1sum }}"
```
benbz: I think we should have a named template (e.g. `element-io.synapse.secret-data` exists) that does all these `$.Files.Get`. That gives us:
* Ensures if we add a new file to the `Secret`/`ConfigMap` these hash labels will get it without us having to remember
* We rename the file in the `Secret`/`ConfigMap` that also changes the hash label

charts/matrix-stack/templates/element-web/_helpers.tpl
```diff
@@ -56,23 +78,25 @@ app.kubernetes.io/version: {{ .image.tag }}
...
+# the server block iff there's no add_header directives in the location block
+security_headers.conf: |-
+  {{- ($root.Files.Get "configs/element-web/security_headers.conf") | nindent 2 }}
+{{- end }}
+
+{{- define "element-io.element-web.configmap-data" -}}
+{{- $root := .root }}
+{{- with required "element-io.element-web.configmap-data missing context" .context -}}
+config.json: |-
+  {{- include "element-io.element-web.config" (dict "root" $root "context" .) | nindent 2 }}
```
benbz: I think this named template should be converted into a tpl file for inclusion

charts/matrix-stack/templates/haproxy/_helpers.tpl
```diff
@@ -7,10 +7,28 @@ SPDX-License-Identifier: AGPL-3.0-only
...
 app.kubernetes.io/version: {{ .image.tag }}
 {{- end }}
 {{- end }}
+
+{{- define "element-io.haproxy.configmap-data" }}
+{{- $root := .root -}}
+{{- with required "element-io.haproxy.configmap-data missing context" .context -}}
+haproxy.cfg: |
+  {{- tpl ($root.Files.Get "configs/haproxy/haproxy.cfg.tpl") (dict "root" $root "context" .) | nindent 2 }}
+429.http: |
```
benbz, line 23: To be included via a file

charts/matrix-stack/templates/postgres/_helpers.tpl
```diff
@@ -121,3 +121,43 @@ true
 {{- end -}}
 {{- end -}}
 {{- end -}}
+
+
+{{- define "element-io.postgres.configmap-data" -}}
+{{- $root := .root -}}
+{{- with required "element-io.postgres.configmap-data" .context -}}
+configure-dbs.sh: |
```
benbz, line 129: Should come from an included file

charts/matrix-stack/templates/matrix-rtc/_helpers.tpl
```diff
@@ -121,10 +121,40 @@ app.kubernetes.io/version: {{ .image.tag }}
...
 {{- $root := .root -}}
 {{- with required "element-io.matrix-rtc-sfu.config missing context" .context -}}
 {{- $config := (tpl ($root.Files.Get "configs/matrix-rtc/sfu/config.yaml.tpl") (dict "root" $root "context" .)) | fromYaml }}
-{{- toYaml (merge (.additional | fromYaml) $config) }}
+config.yaml: |
+  {{- toYaml (merge (.additional | fromYaml) $config) | nindent 2 }}
+{{- if not ($root.Values.matrixRTC.livekitAuth).keysYaml }}
+keys-template.yaml: |
+  ${LIVEKIT_KEY}: ${LIVEKIT_SECRET}
+{{- end -}}
```
benbz: These 2 feel so simple that it isn't worth putting in an included file. But consistency?

charts/matrix-stack/templates/synapse/_synapse_pod.tpl
```diff
@@ -14,12 +14,12 @@ template:
   metadata:
     labels:
 {{- if $isHook }}
-      {{- include "element-io.synapse-check-config-hook.labels" (dict "root" $root "context" .) | nindent 6 }}
+      {{- include "element-io.synapse-check-config-hook.labels" (dict "root" $root "context" (dict "image" .image "labels" .labels "withChartVersion" false "processType" $processType)) | nindent 6 }}
```
benbz: Why `image`? I see `$root.Values.synapse.image.tag` being used in this helper

charts/matrix-stack/templates/synapse/_synapse_pod.tpl
```diff
@@ -14,12 +14,12 @@ template:
...
 {{- else }}
-      {{- include "element-io.synapse.process.labels" (dict "root" $root "context" .) | nindent 6 }}
+      {{- include "element-io.synapse.process.labels" (dict "root" $root "context" (dict "image" .image "labels" .labels "withChartVersion" false "isHook" $isHook "processType" $processType)) | nindent 6 }}
```
benbz, line 19: Same query about `image`

tests/integration/test_networking.py
```diff
@@ -23,6 +23,7 @@ async def test_services_have_matching_labels(
 ):
     ignored_labels = [
         "app.kubernetes.io/managed-by",
+        "helm.sh/chart",
```
benbz, line 26: Because the Pod isn't going to have this label anymore but the Ingress will?

charts/matrix-stack/templates/haproxy/_helpers.tpl
gaelgatelement: > To be included via a file

For this one its a bit painful as the file cannot have a license in there. I either have to add a `.license` alongside, or add it to reuse toml file. WDYT?

benbz: Let's do `REUSE.toml`. There's nothing else in the `charts` tree that uses a `.license` file, but there are files listed in `REUSE.toml`

tests/integration/test_networking.py
gaelgatelement: Yes

.github/workflows/build-test.yml
```diff
@@ -4,7 +4,7 @@
 
 name: Helm Chart Building tests
 on:
-  pull_request_target:
+  pull_request:
```

REUSE.toml
```diff
@@ -29,3 +29,13 @@ SPDX-License-Identifier = "AGPL-3.0-only"
...
 SPDX-FileCopyrightText = "Copyright 2025 New Vector Ltd"
 SPDX-License-Identifier = "AGPL-3.0-only"
+
+[[annotations]]
+path = "charts/matrix-stack/configs/haproxy/429.http"
+SPDX-FileCopyrightText = "Copyright 2025 New Vector Ltd"
+SPDX-License-Identifier = "AGPL-3.0-only"
+
+[[annotations]]
+path = "charts/matrix-stack/configs/matrix-rtc/sfu/keys-template.yaml"
```
benbz: Why can't this have the copyright headers?

PR #381, opened 2025-04-17
Extract Synapse config into template files

benbz:
Follow on from #373

PR #387, opened 2025-04-24
synapse & element web: make sure serverName can use helm templates

gaelgatelement:
Minor regression from https://github.com/element-hq/ess-helm/pull/381


benbz (commented):
And `charts/matrix-stack/configs/element-web/config.json.tpl` ?

benbz (approved):

PR #390, opened 2025-04-25
synapse: allow to tpl postgres host

gaelgatelement:


benbz (approved):
LGTM. The MAS reference to Postgres host is `tpl`'ed and so are other Synapse usages AFAICS

PR #391, opened 2025-04-25
Add test for pod containers env values.

gaelgatelement:


tests/manifests/test_pod_env.py
```diff
@@ -0,0 +1,31 @@
...
+        if template["kind"] in ["Deployment", "StatefulSet"]:
+            id = f"{template['kind']}/{template['metadata']['name']}"
+            for container in template["spec"]["template"]["spec"]["containers"]:
+                env_keys = [env["name"] for env in container.get("env", [])]
+                assert len(env_keys) == len(set(env_keys)), f"Env keys are not unique: {id}, {env_keys}"
+
+
+@pytest.mark.parametrize("values_file", values_files_to_test)
+@pytest.mark.asyncio_cooperative
+async def test_unique_env_values_are_quoted_in_containers(templates):
```
benbz:
```suggestion
async def test_env_values_are_quoted_in_containers(templates):
```

benbz: I think we actually need to set some `extraEnv` in this test, not all components will have env vars by default IIRC

tests/manifests/test_pod_env.py
```diff
@@ -0,0 +1,31 @@
...
+import pytest
+
+from . import values_files_to_test
+
+
+@pytest.mark.parametrize("values_file", values_files_to_test)
+@pytest.mark.asyncio_cooperative
+async def test_unique_env_name_in_containers(templates):
+    for template in templates:
+        if template["kind"] in ["Deployment", "StatefulSet"]:
```
benbz: Also `Job`

tests/manifests/test_pod_env.py
benbz: Ideally the `value` set in `extraEnv` should be one that requires quoting when rendered

benbz (approved):

PR #395, opened 2025-04-25
Replace `merge` with `mergeOverwrite` to properly merge booleans.

gaelgatelement:
`merge` has some strange behaviour regarding boolean, and `true` will always win over `false`, even if it's not the dict with precedence.

See https://github.com/helm/helm/issues/5238

### DIFF
--- a/charts/matrix-stack/templates/ess-library/_credentials.tpl
+++ b/charts/matrix-stack/templates/ess-library/_credentials.tpl
@@ -9,7 +9,7 @@ SPDX-License-Identifier: AGPL-3.0-only
 {{- $root := .root -}}
 {{- with required "element-io.ess-library.value-from-values-path missing context" .context -}}
 {{- $path := . -}}
-{{- $navigatedToPart := merge (dict) (mustDeepCopy $root.Values) -}}
+{{- $navigatedToPart := mergeOverwrite (mustDeepCopy $root.Values) (dict) -}}
 {{- range (mustRegexSplit "\\." $path -1) -}}
 {{- if $navigatedToPart -}}
 {{- $navigatedToPart = dig . nil $navigatedToPart -}}
--- a/charts/matrix-stack/templates/ess-library/_credentials.tpl
+++ b/charts/matrix-stack/templates/ess-library/_credentials.tpl
@@ -9,7 +9,7 @@ SPDX-License-Identifier: AGPL-3.0-only
 {{- $root := .root -}}
 {{- with required "element-io.ess-library.value-from-values-path missing context" .context -}}
 {{- $path := . -}}
-{{- $navigatedToPart := mergeOverwrite (mustDeepCopy $root.Values) (dict) -}}
+{{- $navigatedToPart := mustMergeOverwrite (dict) (mustDeepCopy $root.Values) -}}
 {{- range (mustRegexSplit "\\." $path -1) -}}
 {{- if $navigatedToPart -}}
 {{- $navigatedToPart = dig . nil $navigatedToPart -}}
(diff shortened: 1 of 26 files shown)


benbz (commented):
* Do we want to prefer `mustMergeOverwrite` so that failures to merge are errors?
* We should have manifest tests asserting the merge precedence for EW, WKD and MatrixRTC. Esp as they're different.
  * Happy for this to be a separate MR though

charts/matrix-stack/templates/ess-library/_credentials.tpl
```diff
@@ -9,7 +9,7 @@ SPDX-License-Identifier: AGPL-3.0-only
 {{- $root := .root -}}
 {{- with required "element-io.ess-library.value-from-values-path missing context" .context -}}
 {{- $path := . -}}
-{{- $navigatedToPart := merge (dict) (mustDeepCopy $root.Values) -}}
+{{- $navigatedToPart := mergeOverwrite (mustDeepCopy $root.Values) (dict) -}}
```
gaelgatelement: Unsure what that was supposed to do @benbz

benbz: IIRC `$root.Values` isn't a standard map so it couldn't be traversed. The merge into a `dict` meant that it did become a standard map

benbz (approved):

benbz:
>     * We should have manifest tests asserting the merge precedence for EW, WKD and MatrixRTC. Esp as they're different.
>       
>       * Happy for this to be a separate MR though

And the Synapse logging levels

tests/manifests/test_matrix_rtc.py
```diff
@@ -0,0 +1,49 @@
...
+            assert tcp_port == 30881
+            break
+    else:
+        raise RuntimeError("Could not find config.yaml")
+
+    values["matrixRTC"].setdefault("sfu", {})["additional"] = yaml.dump(
+        {
+            "rtc": {
+                "use_external_ip": False,
+                "tcp_port": 32001,
```
benbz, line 31: I suspect we want to revisit the merge order for Matrix RTC to disallow this but this is correct for now and this can happen in a separate PR

tests/manifests/test_well_known_delegation.py
```diff
@@ -31,10 +40,15 @@ async def assert_well_known_files(
     if expected_client is None:
         expected_client = {}
 
-    client_config = {"testclientkey": {"testsubket": "testvalue"}}
-    server_config = {"testserverkey": {"testsubket": "testvalue"}}
-    element_config = {"testelementkey": {"testsubket": "testvalue"}}
-    support_config = {"testsupportkey": {"testsubket": "testvalue"}}
+    if client_config is None:
+        client_config = {"testclientkey": {"testsubket": "testvalue"}}
```
benbz:
```suggestion
        client_config = {"testclientkey": {"testsubkey": "testvalue"}}
```

And below

charts/matrix-stack/templates/postgres/_helpers.tpl
```diff
@@ -114,7 +114,7 @@ true
                         "LC_COLLATE" "C"
                         "LC_CTYPE" "C"
 -}}
-{{- $resultEnv := merge $resultEnv $overrideEnv -}}
+{{- $resultEnv := mustMergeOverwrite $overrideEnv $resultEnv -}}
```
benbz: I don't think this is particularly worth having a test for. Agreed?

gaelgatelement: Well actually it was buggy : 
 - We did not have `extraEnv` on postgres
 - The override was wrongly ordered

benbz (approved):

PR #427, opened 2025-05-06
Synapse v1.129.0, MAS v0.16.0

AndrewFerr:


benbz (commented):
LGTM modulo changelog entry

newsfragments/427.changed.md
benbz, line 1: Does this sensibly render? If you repeatedly `towncrier create 427.changed` it will construct distinct files that definitely get rendered sensibly

AndrewFerr: It indeed does not render properly; thanks for the tip!  Fixed in the latest commit.

PR #430, opened 2025-05-09
Make probe thresholds and frequencies configurable

benbz:
Make the thresholds and frequencies for all probes configurable.

Draft as I want to split this up a bit


tests/manifests/test_pod_probes.py
```diff
@@ -0,0 +1,280 @@
...
+    iterate_deployables_workload_parts(deployables_details, set_probe_details)
+    return deployable_details_to_probe_details
+
+
+def set_synapse_probe_details(all_values, probe_type):
+    worker_to_probe_details = {}
+    counter = 100
+
+    def set_probe_details(worker_name, values):
+        nonlocal counter
```
gaelgatelement, line 107: Minor comment : For future us, the counter mechanism would deserve a comment explaining the reasoning of the test. It feels a bit counterintuitive at first sight to test the probes value with a counter.

benbz: Added to both `set_(synapse_)?probe_details methods

PR #431, opened 2025-05-09
Add a job to compare generated templates between source and target branches

benbz:
It is often unclear what the impact of changes are. Run through all the CI files generating all the templates for both source and target branches and compare the 2.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -94,3 +94,107 @@ jobs:
         for checkov_values in charts/matrix-stack/ci/*checkov*values.yaml; do
           scripts/checkov.sh "$checkov_values"
         done
+
+  template-dyff:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write # required to post a comment to a pull request
+    steps:
+
+    - name: Checkout PR
+      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+
+    # helm template doesn't reliably order manifests within the same kind, so use yq to do it for us
+    - name: Generate manifests for PR
+      run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/build-test.yml
```diff
@@ -94,3 +94,59 @@ jobs:
...
+
+    - name: dyff old and new manifests
+      run: |
+        old_files=$(find /tmp/old -name '*values.yaml' -maxdepth 1 | sed 's|/tmp/old/||' | sort)
+        new_files=$(find /tmp/new -name '*values.yaml' -maxdepth 1 | sed 's|/tmp/new/||' | sort)
+        echo "The following CI values files have been removed in this PR: $(comm -23 <(echo "$old_files") <(echo "$new_files"))"
+        echo "The following CI values files have been added in this PR: $(comm -13 <(echo "$old_files") <(echo "$new_files"))"
+
+        for values in $(comm -12 <(echo "$old_files") <(echo "$new_files")); do
+          dyff between "/tmp/old/$values" "/tmp/new/$values";
```
gaelgatelement: Could it be worth to post a comment on the PR when the dyff is not empty ? 

For example : https://github.com/mikekantzer-drizly/dyff-action

benbz: That action doesn't actually use `dyff` AFAICT - https://github.com/mikekantzer-drizly/dyff-action/blob/main/src/diff.ts#L12.

Will investigate further

benbz: https://github.com/int128/diff-action is where that action was forked from. It doesn't do `dyff` either but I wonder whether I could set `dyff` up as the tool behind `git diff` to make this work

gaelgatelement: An alternative is to do it ourselves with : https://github.com/marketplace/actions/create-or-update-comment

gaelgatelement (approved):

PR #432, opened 2025-05-12
MAS: remove default keys from method calls

gaelgatelement:


charts/matrix-stack/source/matrixAuthenticationService.yaml.j2
```diff
@@ -35,10 +35,10 @@ privateKeys:
 {{ sub_schema_values.labels() }}
 {{ sub_schema_values.serviceAccount() }}
 {{ sub_schema_values.nodeSelector() }}
-{{ sub_schema_values.tolerations(global=false, key='tolerations') }}
-{{ sub_schema_values.topologySpreadConstraints(global=true, key= 'topologySpreadConstraints') }}
+{{ sub_schema_values.tolerations() }}
+{{ sub_schema_values.topologySpreadConstraints(global=true) }}
```
benbz: Surely this shouldn't be `global=true`?

PR #442, opened 2025-05-13
pytest: fix CA cache was regenerated every run

gaelgatelement:
Testing in browser now works fine when importing `~/.cache/pytest-ess/ess-ca.key`

### DIFF
--- a/newsfragments/442.internal.md
+++ b/newsfragments/442.internal.md
@@ -1,0 +1,1 @@
+Fix pytest CA was re-generated on every pytest run, preventing tests in local browsers.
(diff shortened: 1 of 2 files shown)


benbz (approved):
LGTM but do we want to use `.ca/ca.{crt,pem}` as the root CA if it exists so that locally we only need to trust 1 CA cert

gaelgatelement:
> LGTM but do we want to use `.ca/ca.{crt,pem}` as the root CA if it exists so that locally we only need to trust 1 CA cert

I think the local `pytest` cache is better in this regards. With 2 projects cloned locally of ESS Helm, you need to trust 2 CA, while pytest cache allows you to trust only 1 CA.

benbz:
> > LGTM but do we want to use `.ca/ca.{crt,pem}` as the root CA if it exists so that locally we only need to trust 1 CA cert
> 
> I think the local `pytest` cache is better in this regards. With 2 projects cloned locally of ESS Helm, you need to trust 2 CA, while pytest cache allows you to trust only 1 CA.

Or you just copy the contents of `.ca` from 1 checkout to another :p

Regardless that's a separate change and so still approved

PR #449, opened 2025-05-15
Check copyright dates in diffs

gaelgatelement:
This will help us keeping clean copyright headers dates.

### DIFF
--- a/.github/workflows/licensing.yml
+++ b/.github/workflows/licensing.yml
@@ -19,6 +19,7 @@ jobs:
       uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4
       with:
         ref: ${{ github.event.pull_request.head.sha }}
+        fetch-depth: 0
 
     - uses: matrix-org/setup-python-poetry@5bbf6603c5c930615ec8a29f1b5d7d258d905aa4  # v2
       with:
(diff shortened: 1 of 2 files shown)


.github/workflows/licensing.yml
```diff
@@ -40,3 +40,54 @@ jobs:
...
+      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4
+      with:
+        ref: ${{ github.event.pull_request.base.sha }}
+
+# REUSE-IgnoreStart
+    - name: Copyright dates
+      env:
+        BASE_REF: ${{ github.event.pull_request.base.sha }}
+      run: |
+        echo "Comparing $BASE_REF against current branch HEAD $(git rev-parse HEAD)"
```
benbz: $HEAD and $BASE_REF are the same commit due to `Checkout target` above

.github/workflows/licensing.yml
```diff
@@ -40,3 +40,54 @@ jobs:
...
+
+    - name: Checkout target
+      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4
+      with:
+        ref: ${{ github.event.pull_request.base.sha }}
+
+# REUSE-IgnoreStart
+    - name: Copyright dates
+      env:
+        BASE_REF: ${{ github.event.pull_request.base.sha }}
```
benbz, line 48: Is this going to have the same problem as `templates-dyff` in that this commit is the current HEAD of the target branch rather than most-recent common ancestor of the PR branch and the target branch?

benbz: No. Because `git diff A...B` is the same as `git diff $(git merge-base A B) B` and `git merge-base A B` returns the most recent common ancestor for A & B.

#450 uses `git merge-base` to fix the `template-dyff` problem

benbz (approved):

PR #450, opened 2025-05-15
Use most-recent common ancestor as base for `template-dyff` job

benbz:
We don't want changes on the target branch since the PR branched off to contribute to the `dyff`


.github/workflows/build-test.yml
```diff
@@ -4,7 +4,7 @@
 
 name: Helm Chart Building tests
 on:
-  pull_request_target:
```
benbz: Actually remember to drop this commit before merging this time

gaelgatelement (approved):

PR #454, opened 2025-05-16
Implement Synapse to MAS migration

gaelgatelement:
WIP.

Options below are the one we envisioned. In the end we are going for the option B.2.

**Option A: From Synapse to MAS-Enabled in 1 values file change.**.  This would allow the user to enable matrixAuthenticationService and run the migrate job at the same time. 2 ways to approach this : 
   1. With Syn2MAS as a pre-upgrade job. This does not work as the postgres database might not exist at this step : 
    - We could run a pre-upgrade hook to initialize it before the syn2mas job runs. This breaks if MAS is using the in-cluster postgres DB, and it was not initialized as no component was using it before
    - The strategy could be to initialize a statefulset during the pre-upgrade hook. It would create the PVC during this step as well. But this would conflict with any currently running postgres. 
   2. With Syn2MAS as a post-upgrade job, Synapse is reconfigured to delegate auth to MAS before the job is run. We could configure Synapse so that it does not delegate authentication if Syn2Mas is running, but then we need need a 2-steps helm upgrade process.

 **Option B: From Synapse to MAS-Enabled in 2-steps values file change**.
   1. Handle Synapse replicas in the chart depending on syn2mas value.
       1. Enabling the migration:
       -  Synapse replicas are set to 0 as `matrixAuthenticationService.migrate.enabled` is true
       - matrixAuthenticationService is not deployed
       - syn2mas post-install/post-upgrade hook
      2. Migration ran properly, disable `matrixAuthenticationService.migrate.enabled`
       - Synapse replicas are set back
       - matrixAuthenticationService is deployed

  2. Handle Synapse replicas from the migration job.
       1. Enabling MAS, enable migration in dryRun mode
         -  MAS is deployed together with its DB.
         - Synapse does not delegate auth to MAS.
         - MAS is in read-only mode OR its ingress is not enabled.
         - The migration dry runs as a post-upgrade hook job.
        2. Enabling the actual migration
         - The migration runs as a pre-hook upgrade.
         - This pre-hook actually scales down synapse during the migration run.
         - If it fails or if it succeeds, in any case it scales back synapse.
         - matrixAuthenticationService read-only is disabled OR its ingress is enabled
         - Synapse delegated its authentication to MAS

### DIFF
--- a/charts/matrix-stack/source/common/sub_schema_values.yaml.j2
+++ b/charts/matrix-stack/source/common/sub_schema_values.yaml.j2
@@ -13,7 +13,7 @@ SPDX-License-Identifier: AGPL-3.0-only
 
 ## The matrix-tools image, used in multiple components
 matrixTools:
-{{ image(registry="ghcr.io", repository="element-hq/ess-helm/matrix-tools", tag="0.4.0") | indent(2) }}
+{{ image(registry="ghcr.io", repository="element-hq/ess-helm/matrix-tools", tag="0.5.0") | indent(2) }}
 
 ## CertManager Issuer to configure by default automatically on all ingresses
 ## If configured, the chart will automatically generate the tlsSecret name for all ingresses
--- a/charts/matrix-stack/values.yaml
+++ b/charts/matrix-stack/values.yaml
@@ -20,7 +20,7 @@ matrixTools:
 
     ## The tag of the container image to use.
     ## Defaults to the Chart's appVersion if not set
-    tag: "0.4.0"
+    tag: "0.5.0"
 
     ## Container digest to use. Used to pull the image instead of the image tag / Chart appVersion if set
     # digest:
--- a/newsfragments/454.changed.md
+++ b/newsfragments/454.changed.md
@@ -1,0 +1,1 @@
+matrix-tools: Update to 0.5.0 to support syn2mas migration command.
--- a/tests/integration/env/matrix-authentication-service-syn2mas.rc
+++ b/tests/integration/env/matrix-authentication-service-syn2mas.rc
@@ -3,4 +3,3 @@
 # SPDX-License-Identifier: AGPL-3.0-only
 
 export TEST_VALUES_FILE=charts/matrix-stack/ci/pytest-matrix-authentication-service-syn2mas-values.yaml
-export BUILD_MATRIX_TOOLS=1


benbz:
My thoughts:
* We must do a 2 stage deploy due to the Postgres issue
  * We may not have had a Postgres running previously if Synapse used an external Postgres and MAS is to use the in-cluster Postgres
  * We can't spin up one in a `pre-install`/`pre-upgrade` hook as we may have an in-cluster Postgres currently and the PVC can't be shared
* We can't do the migration in a `post-install`/`post-upgrade` hook as the Synapse config will be wrong at some point during the deploy
  * If we update it for MAS then there's a period of time before the `post-install`/`post-upgrade` hook runs that Synapse is pointing at MAS that hasn't had data migrated to it
  * If we don't update it for MAS then after the `post-install`/`post-upgrade` hook runs the config isn't setup for MAS
    * To say nothing for what happens if a subsequent `helm install -U`/`helm upgrade` occurs
* We should have "modes" for migration rather than a set of boolean flags so that we can eliminate bad states and make the `helm upgrade --reuse-values` incantations as simple as possible
* By default the chart should store markers in-cluster as to the state of the migration.
  * I don't think adding a `syn2mas` marker is sufficient as I think we also want a marker for `Synapse has been run with MAS, a migration will be required`
  * We'll also want `MAS prepared but not migrated to` & `MAS deployed and used for auth` (be it migrated to or from the beginning.
  * The `MAS prepared but not migrated to` marker must be removable in-case the admin redeploys the chart with different MAS DB config

gaelgatelement:
> We should have "modes" for migration rather than a set of boolean flags so that we can eliminate bad states and make the helm upgrade --reuse-values incantations as simple as possible

With current booleans it feels simple : 

```
helm upgrade --reuse-values --set matrixAuthenticationService.enabled=true --set matrixAuthenticationService.syn2mas=true
```

On success, apply the migration : 

```
helm upgrade --reuse-values --set matrixAuthenticationService.syn2mas.dryRun=false
```

I'm not sure a "mode" would make it simpler here.

Regarding markers, I'm not sure to see in which context they are required. I'm not a huge fan of storing state when we can just avoid it. What type of critical error scenario those states would prevent ?

benbz:
> > We should have "modes" for migration rather than a set of boolean flags so that we can eliminate bad states and make the helm upgrade --reuse-values incantations as simple as possible
> 
> With current booleans it feels simple :
> 
> ```
> helm upgrade --reuse-values --set matrixAuthenticationService.enabled=true --set matrixAuthenticationService.syn2mas=true
> ```
> 
> On success, apply the migration :
> 
> ```
> helm upgrade --reuse-values --set matrixAuthenticationService.syn2mas.dryRun=false
> ```
> 
> I'm not sure a "mode" would make it simpler here.

```sh
# Edit values.yaml to enable MAS and configure matrixAuthenticationService.migrationState=preparation
$ helm upgrade -f values.yaml
# Move through the migration steps
$ helm upgrade --reuse-values --set matrixAuthenticationService.migrationState=dryRun
$ helm upgrade --reuse-values --set matrixAuthenticationService.migrationState=migrate
# Edit values.yaml to remove matrixAuthenticationService.migrationState from values.yaml
```

If we've only got `matrixAuthenticationService.syn2mas.{enabled,dryRun}` it probably isn't necessary but if we need the preparation step (i.e. what `preMigrationSynapseHandlesAuth` did) then a single mode variable is better than booleans.

> Regarding markers, I'm not sure to see in which context they are required. I'm not a huge fan of storing state when we can just avoid it. What type of critical error scenario those states would prevent ?

* Prevents upgrading the chart with MAS enabled but no migration flags/mode if Synapse has been deployed before
* Prevents upgrading the chart with MAS disabled if MAS has been migrated to / was initially deployed
* Prevents upgrading the chart with MAS migration turned on but has already been completed

gaelgatelement:
I think we can somehow skip doing deployment markers, as they will conflict with the "all components are optional" philosophy.

For example I would argue the following points:

>    Prevents upgrading the chart with MAS enabled but no migration flags/mode if Synapse has been deployed before

This could still result in a valid deployment as long as no user registered on Synapse. So a 2 step deployment where Synapse is enabled first, and then the admin adds MAS right away would work.

>    Prevents upgrading the chart with MAS disabled if MAS has been migrated to / was initially deployed

This could be a check that we'd enforce for a lot of components. Does that actually make sense ? Let's imagine that one wants to migrate to MAS external to the chart, that should remain possible.

>    Prevents upgrading the chart with MAS migration turned on but has already been completed

I think this is fine as it will run as a pre-upgrade hook job, so it will fail without doing any harm.

benbz:
> I think we can somehow skip doing deployment markers, as they will conflict with the "all components are optional" philosophy.
> 
> For example I would argue the following points:
> 
> > Prevents upgrading the chart with MAS enabled but no migration flags/mode if Synapse has been deployed before
> 
> This could still result in a valid deployment as long as no user registered on Synapse. So a 2 step deployment where Synapse is enabled first, and then the admin adds MAS right away would work.

Yes no-one might have registered on Synapse but that feels like an edge case given MAS is default enabled. Given how wrong things will go if MAS is enabled but not migrated to, in ways that will take DB wrangling to untangle, I think we want to prevent this.

This is the scenario I'm most worried about given if we tell people to deploy with MAS off to enable some incompatible component and then we make that component compatible, people will shoot themselves in the foot turning MAS on.

> > Prevents upgrading the chart with MAS disabled if MAS has been migrated to / was initially deployed
> 
> This could be a check that we'd enforce for a lot of components. Does that actually make sense ? Let's imagine that one wants to migrate to MAS external to the chart, that should remain possible.

If the admin wants to move to an external MAS they can remove the marker. Again removing MAS and falling back to Synapse auth will make things go very wrong in DB wrangling required ways to fix. I don't think this particularly conflicts with the everything is optional approach - MAS is optional until the point you use it and there's an escape hatch to move to an external instance. MAS is fundamentally different to bridges, clients, etc. Disabling Synapse after initial deploy would require significant reconfiguration too.

This is the least likely scenario but still worth protecting IMO.

> > Prevents upgrading the chart with MAS migration turned on but has already been completed
> 
> I think this is fine as it will run as a pre-upgrade hook job, so it will fail without doing any harm.

If it is a pre-upgrade hook then yes `syn2mas` will fail but if we're adding markers for the above 2 cases we might be able to make a cleaner/more obvious error.

This scenario will happen often and so if we can improve the UX of it, great.

My thinking was that we'd mark these markers part of `initSecrets`/check config. Or do something very similar. If people have RBAC, etc concerns then they turn those off and don't get this protection.

charts/matrix-stack/source/matrixAuthenticationService.yaml.j2
```diff
@@ -45,3 +40,28 @@ privateKeys:
...
+  {{- sub_schema_values.nodeSelector() | indent(2) -}}
+  {{- sub_schema_values.podSecurityContext(user_id='10005', group_id='10005') | indent(2) -}}
+  {{- sub_schema_values.resources(requests_memory='50Mi', requests_cpu='50m', limits_memory='350Mi') | indent(2) -}}
+  {{- sub_schema_values.serviceAccount() | indent(2) -}}
+  {{- sub_schema_values.tolerations() | indent(2) }}
+
+  ## Runs the syn2mas process in dryRun mode.
+  ## Force the authentication to happen with legacy authentication.
+  ## This can be used to deploy Matrix Authentication Service and keeping auth on Synapse.
+  ## This should be set back to false and never switch again after the migration to MAS has been run.
```
benbz: Can we validate this in the chart or with the markers?

charts/matrix-stack/templates/ess-library/_render_config.tpl
```diff
@@ -9,6 +9,7 @@ SPDX-License-Identifier: AGPL-3.0-only
 {{- with required "element-io.ess-library.render-config-container missing context" .context -}}
 {{- $context := . -}}
 {{- $nameSuffix := required "element-io.ess-library.render-config-container missing context.nameSuffix" .nameSuffix -}}
+{{- $containerName := .containerName | default "render-config" -}}
```
benbz: Why do we need this to be configurable?

charts/matrix-stack/templates/ess-library/_render_config.tpl
```diff
@@ -9,6 +9,7 @@ SPDX-License-Identifier: AGPL-3.0-only
 {{- with required "element-io.ess-library.render-config-container missing context" .context -}}
 {{- $context := . -}}
 {{- $nameSuffix := required "element-io.ess-library.render-config-container missing context.nameSuffix" .nameSuffix -}}
+{{- $containerName := (.containerName | default "render-config") -}}
```
benbz, line 12: Why do we need this?

charts/matrix-stack/ci/fragments/matrix-authentication-service-syn2mas-migrate.yaml
benbz, line 1: The password scheme stuff should be here either as well as or instead of in the dry-run file

charts/matrix-stack/templates/matrix-authentication-service/deployment.yaml
```diff
@@ -32,7 +32,7 @@ spec:
   strategy:
     type: RollingUpdate
     rollingUpdate:
-      maxUnavailable: 1
+      maxUnavailable: {{ min (max 0 (sub .replicas 1)) 1 }}
```
benbz: Can you document why this is here

charts/matrix-stack/templates/matrix-authentication-service/syn2mas_role.yaml
```diff
@@ -0,0 +1,44 @@
...
+*/}}
+- apiGroups: ["apps"]
+  resources: ["statefulsets"]
+  verbs: ["list"]
+- apiGroups: ["apps"]
+  resources: ["statefulsets"]
+  resourceNames:
+{{- $enabledWorkers := (include "element-io.synapse.enabledWorkers" (dict "root" $)) | fromJson }}
+{{- range $processType, $unmergedProcessDetails := mustMergeOverwrite (dict "main" dict) $enabledWorkers }}
+  - "{{ $.Release.Name }}-synapse-{{ $processType }}"
```
benbz: `$processType` will be wrong for renamed workers. Can we use wildcards here or update to use the renamed value

charts/matrix-stack/templates/matrix-authentication-service/syn2mas_job.yaml
```diff
@@ -0,0 +1,258 @@
...
+{{- if .dryRun }}
+    "helm.sh/hook": post-install,post-upgrade
+{{- else }}
+    "helm.sh/hook": pre-install,pre-upgrade
+{{- /*
+Hook Weights are
+- -10 : The initSecret hook generating secrets used by the check config job
+- -5 : The synapse secret & configMap for the hook, so that they are created before the job
+- 0 : The job itself, so that it is run after the secrets and configs are created &
+*/}}
```
benbz: Docs still seem to be for check-config hook

charts/matrix-stack/templates/matrix-authentication-service/syn2mas_role.yaml
```diff
@@ -0,0 +1,44 @@
...
+  namespace: {{ $.Release.Namespace }}
+  labels:
+    {{- include "element-io.matrix-authentication-service-syn2mas.labels" (dict "root" $ "context" .) | nindent 4 }}
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+{{- /*
+Hook Weights are
+- -10 : The initSecret hook generating secrets used by the check config job
+- -5 : The synapse secret & configMap for the hook, so that they are created before the job
+- 0 : The job itself, so that it is run after the secrets and configs are created &
```
benbz: Docs still seem to be for check-config hook

charts/matrix-stack/templates/matrix-authentication-service/syn2mas_rolebinding.yaml
```diff
@@ -0,0 +1,36 @@
...
+  namespace: {{ $.Release.Namespace }}
+  labels:
+    {{- include "element-io.matrix-authentication-service-syn2mas.labels" (dict "root" $ "context" .) | nindent 4 }}
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+{{- /*
+Hook Weights are
+- -10 : The initSecret hook generating secrets used by the check config job
+- -5 : The synapse secret & configMap for the hook, so that they are created before the job
+- 0 : The job itself, so that it is run after the secrets and configs are created &
```
benbz: Docs still seem to be for check-config hook

charts/matrix-stack/templates/matrix-authentication-service/syn2mas_job.yaml
```diff
@@ -0,0 +1,258 @@
...
+          name: tmp-mas-cli
+      {{- include "element-io.matrix-authentication-service.render-config-container" (dict "root" $ "context" $masContext) | nindent 6 }}
+      {{- include "element-io.synapse.render-config-container" (dict "root" $ "context" $synapseContext) | nindent 6 }}
+      - name: db-wait-mas
+{{- with $.Values.matrixTools.image -}}
+{{- if .digest }}
+        image: "{{ .registry }}/{{ .repository }}@{{ .digest }}"
+        imagePullPolicy: {{ .pullPolicy | default "IfNotPresent" }}
+{{- else }}
+        image: "{{ .registry }}/{{ .repository }}:{{ required "matrixTools.image.tag is required if no digest" .tag }}"
```
benbz:
```suggestion
        image: "{{ .registry }}/{{ .repository }}:{{ .tag }}"
```

charts/matrix-stack/templates/matrix-authentication-service/syn2mas_job.yaml
```diff
@@ -0,0 +1,258 @@
...
+      restartPolicy: Never
+{{- include "element-io.ess-library.pods.commonSpec" (dict "root" $ "context" (dict "componentValues" . "instanceSuffix" "matrix-authentication-service-syn2mas" "deployment" false "usesMatrixTools" true "mountServiceAccountToken" true)) | nindent 6 }}
+      initContainers:
+      - name: copy-mas-cli
+{{- with .image -}}
+{{- if .digest }}
+        image: "{{ .registry }}/{{ .repository }}@{{ .digest }}"
+        imagePullPolicy: {{ .pullPolicy | default "IfNotPresent" }}
+{{- else }}
+        image: "{{ .registry }}/{{ .repository }}:{{ required "matrixAuthenticationService.syn2mas.image.tag is required if no digest" .tag }}"
```
benbz:
```suggestion
        image: "{{ .registry }}/{{ .repository }}:{{ .tag }}"
```

charts/matrix-stack/templates/matrix-authentication-service/syn2mas_job.yaml
```diff
@@ -0,0 +1,258 @@
...
+        resources:
+          {{- toYaml . | nindent 10 }}
+{{- end }}
+      - name: db-wait-syn
+{{- with $.Values.matrixTools.image -}}
+{{- if .digest }}
+        image: "{{ .registry }}/{{ .repository }}@{{ .digest }}"
+        imagePullPolicy: {{ .pullPolicy | default "IfNotPresent" }}
+{{- else }}
+        image: "{{ .registry }}/{{ .repository }}:{{ required "matrixTools.image.tag is required if no digest" .tag }}"
```
benbz:
```suggestion
        image: "{{ .registry }}/{{ .repository }}:{{ .tag }}"
```

charts/matrix-stack/templates/matrix-authentication-service/syn2mas_job.yaml
```diff
@@ -0,0 +1,258 @@
...
+          {{- toYaml . | nindent 10 }}
+{{- end }}
+      - name: syn2mas-check
+        args: ["syn2mas", "check", "--config", "/conf/config.yaml", "--synapse-config", "/conf/homeserver.yaml"]
+{{- with .image -}}
+{{- if .digest }}
+        image: "{{ .registry }}/{{ .repository }}@{{ .digest }}"
+        imagePullPolicy: {{ .pullPolicy | default "IfNotPresent" }}
+{{- else }}
+        image: "{{ .registry }}/{{ .repository }}:{{ required "matrixAuthenticationService.image.tag is required if no digest" .tag }}"
```
benbz:
```suggestion
        image: "{{ .registry }}/{{ .repository }}:{{ .tag }}"
```

charts/matrix-stack/templates/matrix-authentication-service/syn2mas_job.yaml
```diff
@@ -0,0 +1,258 @@
...
+      containers:
+      - name: syn2mas-migrate
+{{- if .dryRun }}
+        args: ["syn2mas", "migrate", "--config", "/conf/config.yaml", "--synapse-config", "/conf/homeserver.yaml", "--dry-run"]
+    {{- with .image -}}
+    {{- if .digest }}
+        image: "{{ .registry }}/{{ .repository }}@{{ .digest }}"
+        imagePullPolicy: {{ .pullPolicy | default "IfNotPresent" }}
+    {{- else }}
+        image: "{{ .registry }}/{{ .repository }}:{{ required "matrixAuthenticationService.image.tag is required if no digest" .tag }}"
```
benbz:
```suggestion
        image: "{{ .registry }}/{{ .repository }}:{{ .tag }}"
```

charts/matrix-stack/templates/matrix-authentication-service/syn2mas_job.yaml
```diff
@@ -0,0 +1,258 @@
...
+    {{- end }}
+    {{- end }}
+{{- else }}
+        command: ["/matrix-tools", "syn2mas", "--config", "/conf/config.yaml", "--synapse-config", "/conf/homeserver.yaml"]
+    {{- with $.Values.matrixTools.image -}}
+    {{- if .digest }}
+        image: "{{ .registry }}/{{ .repository }}@{{ .digest }}"
+        imagePullPolicy: {{ .pullPolicy | default "IfNotPresent" }}
+    {{- else }}
+        image: "{{ .registry }}/{{ .repository }}:{{ required "matrixTools.image.tag is required if no digest" .tag }}"
```
benbz:
```suggestion
        image: "{{ .registry }}/{{ .repository }}:{{ .tag }}"
```

charts/matrix-stack/templates/synapse/_helpers.tpl
```diff
@@ -236,3 +236,23 @@ path_map_file: |
...
+
+{{- define "element-io.synapse.render-config-container" -}}
+{{- $root := .root -}}
+{{- with required "element-io.synapse.render-config-container missing context" .context }}
+{{- $processType := required "element-io.synapse.render-config-container context required processType" .processType -}}
+{{- $isHook := required "element-io.synapse.render-config-container context required isHook" .isHook -}}
+{{- include "element-io.ess-library.render-config-container" (dict "root" $root "context"
+            (dict "additionalPath" "synapse.additional"
+                  "nameSuffix" "synapse"
+                  "containerName" (.containerName | default "render-config")
```
benbz, line 248: Doesn't seem to be overridden AFAICT

matrix-tools/internal/pkg/args/args.go
benbz, line 1: Needs tests

tests/manifests/__init__.py
```diff
@@ -561,5 +591,9 @@ def _get_all_deployables_details() -> set[DeployableDetails]:
     sum([component_details.values_files for component_details in all_components_details], tuple())
 ) | set(_extra_values_files_to_test)
 
+workloads_values_files_to_test = set(
```
benbz: Why do we need this extra set vs `values_files_to_test` which gets `_extra_values_files_to_test`?

tests/integration/test_matrix_authentication_service.py
```diff
@@ -11,6 +11,7 @@
 
 
 @pytest.mark.skipif(value_file_has("matrixAuthenticationService.enabled", False), reason="MAS not deployed")
+@pytest.mark.skipif(value_file_has("matrixAuthenticationService.syn2mas.enabled", False), reason="Syn2Mas not deployed")
```
benbz: This feels odd. We skip the graph endpoint test if `syn2mas` is not deployed, which will be most of the time...

tests/integration/fixtures/syn2mas.py
benbz, line 1: Whole file unused?

tests/integration/test_syn2mas.py
```diff
@@ -0,0 +1,76 @@
...
+from .fixtures import ESSData
+from .lib.utils import aiohttp_post_json, value_file_has
+from .test_matrix_authentication_service import test_matrix_authentication_service_graphql_endpoint
+
+
+@pytest.mark.skipif(value_file_has("synapse.enabled", False), reason="Synapse not deployed")
+@pytest.mark.skipif(value_file_has("matrixAuthenticationService.syn2mas.enabled", False), reason="Syn2Mas not deployed")
+@pytest.mark.parametrize("users", [("syn2mas-user",)], indirect=True)
+@pytest.mark.asyncio_cooperative
+async def test_run_syn2mas_upgrade(
```
benbz, line 22: Do we need a skip if MAS is enabled from the start?

tests/integration/test_syn2mas.py
```diff
@@ -0,0 +1,76 @@
...
+    # Install or upgrade a release
+    revision = await helm_client.install_or_upgrade_release(
+        generated_data.release_name,
+        chart,
+        values,
+        namespace=generated_data.ess_namespace,
+        atomic="CI" not in os.environ,
+        wait=True,
+    )
+    assert revision.status == pyhelm3.ReleaseRevisionStatus.DEPLOYED
```
benbz, line 38: Should we also check that the job has run correctly?

tests/integration/test_syn2mas.py
```diff
@@ -0,0 +1,76 @@
...
+    values["matrixAuthenticationService"]["enabled"] = True
+    chart = await helm_client.get_chart("charts/matrix-stack")
+    # Install or upgrade a release
+    revision = await helm_client.install_or_upgrade_release(
+        generated_data.release_name,
+        chart,
+        values,
+        namespace=generated_data.ess_namespace,
+        atomic="CI" not in os.environ,
+        wait=True,
```
benbz: And `wait_for_jobs=True`?

tests/integration/test_syn2mas.py
benbz, line 1: Add marker checks?

charts/matrix-stack/templates/matrix-authentication-service/deployment.yaml
gaelgatelement: I've implemented it to pass the tests, but I think it might deserve to be implemented globally. 

The issue is that the deployment with one replica is instant ready when the pod is created, even if the pod is not ready, as maxUnavailable is 1. In the tests it's causing helm not to wait properly, but I suppose it might be causing other issues.

WDYT?

charts/matrix-stack/templates/ess-library/_render_config.tpl
gaelgatelement: Because the `syn2mas` job has 2 containers rendering the config, one for MAS and one for Synapse. So this is naming them distinctively.

tests/manifests/__init__.py
gaelgatelement: This is because MAS ingress is not deployed until it is ready to handle auth, which has after syn2mas has been run successfully. So this breaks the following tests : 

```
FAILED tests/manifests/test_ingresses.py::test_ingress_is_expected_host[matrix-authentication-service-synapse-syn2mas-dry-run-secrets-externally-values.yaml] - AssertionError: assert {'synapse.ess.localhost'} == {'mas.ess.loc...ss.localhost'}
FAILED tests/manifests/test_ingresses.py::test_ingress_is_expected_host[matrix-authentication-service-synapse-syn2mas-dry-run-secrets-in-helm-values.yaml] - AssertionError: assert {'synapse.ess.localhost'} == {'mas.ess.loc...ss.localhost'}
```

charts/matrix-stack/templates/ess-library/_render_config.tpl
benbz: I haven't seen anything explictly set `containerName`

charts/matrix-stack/ci/fragments/matrix-authentication-service-syn2mas-migrate.yaml
gaelgatelement: The `syn2mas-migrate` values file is actually using both : 

```
matrix-authentication-service-syn2mas-dryrun.yaml matrix-authentication-service-syn2mas-migrate.yaml
```

So that it follows the upgrade flow (first do the dryRun, then switch dryRun off)

tests/manifests/__init__.py
benbz: Right. I think I'd prefer to attempt to skip these 2 values files in the test itself given it is only a single test rather than introduce a whole other set of values files to reason about. We can also have a specific manifest test to assert the ingress situation with syn2mas as well

charts/matrix-stack/ci/fragments/matrix-authentication-service-syn2mas-migrate.yaml
benbz: We sometimes point people at the fragments. Having the password scheme stuff missing from the actual migration fragment feels dangerous in this regards

charts/matrix-stack/ci/fragments/matrix-authentication-service-syn2mas-dryrun.yaml
```diff
@@ -0,0 +1,16 @@
...
+matrixAuthenticationService:
+  syn2mas:
+    enabled: true
+    dryRun: true
+
+  additional:
+    password-scheme.yml:
+      config: |
+        passwords:
+          schemes:
```
benbz, line 14: Should we also have `version: 2` with the MAS default algorithm as well

charts/matrix-stack/templates/matrix-authentication-service/deployment.yaml
benbz: Makes sense. In which case I'd suggest adding it globally using a helper, with manifest tests, in a separate PR.

tests/integration/test_syn2mas.py
gaelgatelement: > Do we need a skip if MAS is enabled from the start?

This is equivalent to the following : 

```
@pytest.mark.skipif(value_file_has("matrixAuthenticationService.syn2mas.enabled", False), reason="Syn2Mas not deployed")
```

tests/integration/test_syn2mas.py
gaelgatelement: The pyhelm3 library does not support it. AFAICT, it is not required for the test to run properly.

charts/matrix-stack/templates/ess-library/_render_config.tpl
gaelgatelement: Because it was buggy but not anymore :smile:

charts/matrix-stack/templates/matrix-rtc/sfu_jwt_deployment.yaml
```diff
@@ -81,16 +81,20 @@ spec:
           {{- toYaml . | nindent 10 }}
 {{- end }}
         volumeMounts:
-{{- range $idx, $secret := include "element-io.matrix-rtc-authorisation-service.configSecrets" (dict "root" $ "context" $.Values.matrixRTC) | fromJsonArray }}
-        - mountPath: /secrets/{{ tpl $secret $ }}
-          name: "secret-{{ $idx }}"
+{{- range $_, $secret := include "element-io.matrix-rtc-authorisation-service.configSecrets" (dict "root" $ "context" $.Values.matrixRTC) | fromJsonArray }}
```
benbz: Can we split these mount changes out into their own PR for ease of reviewing them and and syn2mas?

charts/matrix-stack/source/matrixAuthenticationService.yaml.j2
benbz: If I understand the change in `charts/matrix-stack/templates/deployment-markers/_helpers.tpl` this has now been done. I think this comment just needs to be updated

tests/integration/test_syn2mas.py
gaelgatelement: > Should we also check that the job has run correctly?

I'm not sure to follow here, how could the job be failed if the release revision is deployed?

benbz: Given the lack of `wait_for_jobs` the job may still be running and so doesn't contribute to the status. Or is that not true with Hooks?

gaelgatelement: It looks like Hooks are treated differently yes : 

https://helm.sh/docs/topics/charts_hooks/
> What does it mean to wait until a hook is ready? This depends on the resource declared in the hook. If the resource is a Job or Pod kind, Helm will wait until it successfully runs to completion. And if the hook fails, the release will fail. This is a blocking operation, so the Helm client will pause while the Job is run.

charts/matrix-stack/source/matrixAuthenticationService.yaml.j2
```diff
@@ -47,3 +42,34 @@ privateKeys:
...
+  {{- sub_schema_values.nodeSelector() | indent(2) -}}
+  {{- sub_schema_values.podSecurityContext(user_id='10005', group_id='10005') | indent(2) -}}
+  {{- sub_schema_values.resources(requests_memory='50Mi', requests_cpu='50m', limits_memory='350Mi') | indent(2) -}}
+  {{- sub_schema_values.serviceAccount() | indent(2) -}}
+  {{- sub_schema_values.tolerations() | indent(2) }}
+
+  ## Runs the syn2mas process in dryRun mode.
+  ## Force the authentication to happen with legacy authentication.
+  ## This can be used to deploy Matrix Authentication Service and keeping auth on Synapse.
+  ## This should be set to false and never switch again after the migration to MAS has been run.
```
benbz:
```suggestion
  ## This must be set to false and never switch again after the migration to MAS has been run or the deployment markers hooks will prevent redeploying.
```

charts/matrix-stack/templates/matrix-authentication-service/syn2mas_job.yaml
```diff
@@ -0,0 +1,270 @@
...
+*/}}
+    "helm.sh/hook": post-install,post-upgrade
+    "helm.sh/hook-weight": "-25"
+{{- else }}
+    "helm.sh/hook": pre-install,pre-upgrade
+{{- /*
+Hook Weights are
+- -10 : The initSecret hook generating secrets used by the syn2mas job
+- -5 : The MAS & synapse secret & configMap for the hook, so that they are created before the job
+- 0 : The job itself, so that it is run after the secrets and configs are created &
```
benbz: Trailing `&`. Is there something missing or the `&` needs to go?

charts/matrix-stack/templates/matrix-authentication-service/syn2mas_role.yaml
```diff
@@ -0,0 +1,44 @@
...
+  namespace: {{ $.Release.Namespace }}
+  labels:
+    {{- include "element-io.syn2mas.labels" (dict "root" $ "context" .) | nindent 4 }}
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+{{- /*
+Hook Weights are
+- -10 : The initSecret hook generating secrets used by the syn2mas job
+- -5 : The MAS & synapse secret & configMap for the hook, so that they are created before the job
+- 0 : The job itself, so that it is run after the secrets and configs are created &
```
benbz: Trailing `&`

charts/matrix-stack/templates/matrix-authentication-service/syn2mas_role.yaml
```diff
@@ -0,0 +1,44 @@
+{{- /*
+Copyright 2025 New Vector Ltd
+
+SPDX-License-Identifier: AGPL-3.0-only
+*/ -}}
+{{- with .Values.matrixAuthenticationService -}}
+{{- if and .enabled .syn2mas.enabled -}}
```
benbz:
```suggestion
{{- if and .enabled .syn2mas.enabled (not .syn2mas.dryRun) -}}
```

IIUC this isn't needed until we're doing this in anger

charts/matrix-stack/templates/matrix-authentication-service/syn2mas_rolebinding.yaml
```diff
@@ -0,0 +1,36 @@
+{{- /*
+Copyright 2025 New Vector Ltd
+
+SPDX-License-Identifier: AGPL-3.0-only
+*/ -}}
+
+{{- with .Values.matrixAuthenticationService -}}
+{{- if and .enabled .syn2mas.enabled -}}
```
benbz:
```suggestion
{{- if and .enabled .syn2mas.enabled (not .syn2mas.dryRun) -}}
```

charts/matrix-stack/templates/matrix-authentication-service/syn2mas_rolebinding.yaml
```diff
@@ -0,0 +1,36 @@
...
+  namespace: {{ $.Release.Namespace }}
+  labels:
+    {{- include "element-io.syn2mas.labels" (dict "root" $ "context" .) | nindent 4 }}
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+{{- /*
+Hook Weights are
+- -10 : The initSecret hook generating secrets used by the syn2mas job
+- -5 : The MAS & synapse secret & configMap for the hook, so that they are created before the job
+- 0 : The job itself, so that it is run after the secrets and configs are created &
```
benbz: Trailing `&`

tests/integration/lib/helpers.py
```diff
@@ -59,3 +61,43 @@ async def wait_for_endpoint_ready(name, namespace, cluster, kube_client):
         else:
             break
     return endpoint
+
+
+async def deploy_with_values_patch(
+    generated_data, helm_client: pyhelm3.Client, values_patch: dict, timeout="600s"
+) -> pyhelm3.ReleaseRevision:
+    # After the base chart is setup, we enable MAS to run the syn2mas dry run job
```
benbz: Comment isn't necessarily relevant here

tests/integration/test_syn2mas.py
```diff
@@ -0,0 +1,120 @@
...
+    # We should still be able to reach synapse ingress
+    await ingress_ready("synapse")
+
+    # Auth metadata endpoint should not be reachable
+    with pytest.raises(aiohttp.ClientResponseError):
+        await aiohttp_get_json(
+            f"https://synapse.{generated_data.server_name}/_matrix/client/unstable/org.matrix.msc2965/auth_metadata",
+            ssl_context,
+        )
+
```
benbz:
```suggestion
```

Dupe of lines 38-47 and nothing has changed inbetween

matrix-tools/internal/pkg/args/args.go
benbz, line 1: Reattaching the needs tests comment

docs/syn2mas.md
```diff
@@ -0,0 +1,103 @@
...
+   - At the end of the migration, Synapse will be scaled back up to its original number of replicas.
+   - Matrix Authentication Service restarts and is ready to serve the delegated authentication
+   - The `MATRIX_STACK_MSC3861` marker is updated to reflect the `syn2mas_migrated` state.
+
+Your users are now able to login using the delegated authentication.
+
+
+### Step 3: Disable syn2mas
+
+While syn2mas is enabled, everytime you will run `helm upgrade`, it will downscale Synapse and its workers to 0 replicas.
```
benbz: Doesn't the actual `syn2mas` job fail blocking the deployment?

docs/syn2mas.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+3. This step will deploy the following resources :
+   - The migration job is executed as a pre-upgrade hook.
+   - The job will downscale Synapse and its workers to 0 replicas while the migration runs.
+   - At the end of the migration, Synapse will be scaled back up to its original number of replicas.
+   - Matrix Authentication Service restarts and is ready to serve the delegated authentication
+   - The `MATRIX_STACK_MSC3861` marker is updated to reflect the `syn2mas_migrated` state.
+
+Your users are now able to login using the delegated authentication.
```
benbz: Also not possible to rollback

docs/syn2mas.md
```diff
@@ -0,0 +1,103 @@
...
+3. **After Step 2** – `syn2mas_migrated` (migration completed)
+4. **After Step 3** – `delegated_auth` (migration finalized)
+
+> ⚠️ **Note:** The `MATRIX_STACK_MSC3861` marker will **prevent downgrading** from `syn2mas_migrated` or `delegated_auth` back to `legacy_auth`.
+
+
+## Important Notes
+
+- Please make sure to backup the synapse database before running the migration.
+- The migration is a **one-way process**. Once the system is in the `delegated_auth` state, it cannot be rolled back to `legacy_auth`.
```
benbz:
```suggestion
- The migration is a **one-way process**. Once the system is in the `syn2mas_migrated` or `delegated_auth` states, it cannot be rolled back to `legacy_auth`.
```

benbz (approved):

PR #455, opened 2025-05-16
ess-helm new versioning scheme

gaelgatelement:
Describe the new versioning scheme.


benbz (approved):
Anything in the workflows that expects a `x.y.z+` rather than `xx.yy.zz`?

gaelgatelement:
> Anything in the workflows that expects a `x.y.z+` rather than `xx.yy.zz`?

I could not find anything that would break with `xx.yy.zz` by grepping `[0-9]`

PR #466, opened 2025-05-20
Have a distinct copyright dates job

benbz:
So that failures are clearly identified in emails.

Also harmonise on `sh` for string comparisons rather than mixing `bash` and `sh`


.github/workflows/licensing.yml
```diff
@@ -5,7 +5,7 @@
 name: Licensing checks
 
 on:
-  pull_request_target:
+  pull_request:
```
benbz: Dropme before merging

gaelgatelement (approved):

PR #473, opened 2025-05-22
Add a job to preview the changelog for the next release

benbz:
So that we can easily see what would be generated


.github/workflows/changelog_check.yml
```diff
@@ -3,7 +3,9 @@
 # SPDX-License-Identifier: AGPL-3.0-only
 
 name: Changelog
-on: [pull_request]
+on:
+  pull_request:
```
benbz: Remove me before merging

gaelgatelement (approved):

PR #477, opened 2025-05-22
README tweaks

AndrewFerr:


AndrewFerr:
IMO this doesn't need a changelog entry.

PR #487, opened 2025-05-26
matrix-tools: implement deployment-markers command

gaelgatelement:
This will add a safeguard for future migrations.

This will be used for the following scenarios : 

- Synapse is deployed without MAS : The marker will make sure that helm upgrrades pass only if MAS is not deployed
- Synapse is deployed with MAS : The marker will make sure that helm upgrades pass only if MAS is not disabled

In the future, the syn2mas PR will transition the marker value from `legacy_auth` to `syn2mas_migrated`.
- MAS with `syn2mas` enabled will be allowed if the value is `legacy_auth`
- MAS without `syn2mas` enabled will only be able to be enabled if the value is `syn2mas_migrated` or `delegated_auth`.

### DIFF
--- a/charts/matrix-stack/source/common/sub_schema_values.yaml.j2
+++ b/charts/matrix-stack/source/common/sub_schema_values.yaml.j2
@@ -13,7 +13,7 @@ SPDX-License-Identifier: AGPL-3.0-only
 
 ## The matrix-tools image, used in multiple components
 matrixTools:
-{{ image(registry="ghcr.io", repository="element-hq/ess-helm/matrix-tools", tag="0.3.5") | indent(2) }}
+{{ image(registry="ghcr.io", repository="element-hq/ess-helm/matrix-tools", tag="0.4.0") | indent(2) }}
 
 ## CertManager Issuer to configure by default automatically on all ingresses
 ## If configured, the chart will automatically generate the tlsSecret name for all ingresses
--- a/charts/matrix-stack/values.yaml
+++ b/charts/matrix-stack/values.yaml
@@ -20,7 +20,7 @@ matrixTools:
 
     ## The tag of the container image to use.
     ## Defaults to the Chart's appVersion if not set
-    tag: "0.3.5"
+    tag: "0.4.0"
 
     ## Container digest to use. Used to pull the image instead of the image tag / Chart appVersion if set
     # digest:


charts/matrix-stack/source/deployment-markers.yaml.j2
benbz, line 1: `extraEnv` in schema but not here

charts/matrix-stack/templates/deployment-markers/job.yaml
```diff
@@ -0,0 +1,72 @@
...
+      restartPolicy: OnFailure
+{{- include "element-io.ess-library.pods.commonSpec" (dict "root" $ "context" (dict "componentValues" . "instanceSuffix" (printf "deployment-markers-%s" $step) "deployment" false "usesMatrixTools" true "mountServiceAccountToken" true)) | nindent 6 }}
+      containers:
+      - name: deployment-markers
+{{- with $.Values.matrixTools.image -}}
+{{- if .digest }}
+        image: "{{ .registry }}/{{ .repository }}@{{ .digest }}"
+        imagePullPolicy: {{ .pullPolicy | default "IfNotPresent" }}
+{{- else }}
+        image: "{{ .registry }}/{{ .repository }}:{{ required "matrixTools.image.tag is required if no digest" .tag }}"
```
benbz:
```suggestion
        image: "{{ .registry }}/{{ .repository }}:{{ .tag }}"
```

Given #484 will validate this

charts/matrix-stack/templates/deployment-markers/role.yaml
```diff
@@ -0,0 +1,35 @@
...
+    "helm.sh/hook": {{ $step }}-install,{{ $step }}-upgrade
+    "helm.sh/hook-weight": "-20"
+rules:
+{{/*
+  https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-resources
+  You cannot restrict create or deletecollection requests by resourceName.
+  For create, this limitation is because the object name is not known at authorization time.
+*/}}
+- apiGroups: [""]
+  resources: ["secrets"]
```
benbz: Is the reason to use `Secrets` so that we can reuse the existing `matrix-tools` infra for writing to `Secrets`? Because there's nothing secret/credentially about these markers, so would naively have put them in `ConfigMap`

tests/integration/env/matrix-authentication-service.rc
```diff
@@ -3,3 +3,4 @@
 # SPDX-License-Identifier: AGPL-3.0-only
 
 export TEST_VALUES_FILE=charts/matrix-stack/ci/pytest-matrix-authentication-service-values.yaml
+export BUILD_MATRIX_TOOLS=1
```
benbz: To be reverted before merging

tests/integration/env/matrix-rtc-synapse-wellknown.rc
```diff
@@ -3,3 +3,4 @@
 # SPDX-License-Identifier: AGPL-3.0-only
 
 export TEST_VALUES_FILE=charts/matrix-stack/ci/pytest-matrix-rtc-synapse-wellknown-values.yaml
+export BUILD_MATRIX_TOOLS=1
```
benbz: To be reverted before merging

tests/integration/env/synapse.rc
```diff
@@ -1,5 +1,6 @@
-# Copyright 2024 New Vector Ltd
+# Copyright 2024-2025 New Vector Ltd
 #
 # SPDX-License-Identifier: AGPL-3.0-only
 
 export TEST_VALUES_FILE=charts/matrix-stack/ci/pytest-synapse-values.yaml
+export BUILD_MATRIX_TOOLS=1
```
benbz: To be reverted before merging

benbz (commented):
Bit confused as to the `nothing-enabled-values.yaml` change in `templates-dyff`, suspect extra whitespace? Maybe the `{{ range $step:= ... }}` needs to be `{{- range ... }}`?

charts/matrix-stack/templates/deployment-markers/role.yaml
gaelgatelement: Yes it's mainly the reason. But I thought about it last night and I think I will use another command, as to make sure we do not confuse both tools (potentially we could generate a marker in the `init-secrets` container and it just feels wrong)

matrix-tools/internal/pkg/marker/marker.go
```diff
@@ -0,0 +1,84 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to initialize secret: %w", err)
+		}
+	} else {
+		if managedBy, ok := existingConfigMap.Labels["app.kubernetes.io/managed-by"]; ok {
+			if managedBy != "matrix-tools-deployment-markers" {
+				return fmt.Errorf("secret %s/%s is not managed by this matrix-tools-deployment-markers", namespace, name)
+			}
+		} else {
+			return fmt.Errorf("secret %s/%s is not managed by this matrix-tools-deployment-markers", namespace, name)
```
benbz:
```suggestion
			return fmt.Errorf("configmap %s/%s is not managed by this matrix-tools-deployment-markers", namespace, name)
```

matrix-tools/internal/pkg/marker/marker.go
```diff
@@ -0,0 +1,84 @@
...
+		existingConfigMap, err = configMapsClient.Create(ctx, &corev1.ConfigMap{
+			ObjectMeta: configMapMeta, Data: nil}, metav1.CreateOptions{},
+		)
+		if err != nil {
+			return fmt.Errorf("failed to initialize secret: %w", err)
+		}
+	} else {
+		if managedBy, ok := existingConfigMap.Labels["app.kubernetes.io/managed-by"]; ok {
+			if managedBy != "matrix-tools-deployment-markers" {
+				return fmt.Errorf("secret %s/%s is not managed by this matrix-tools-deployment-markers", namespace, name)
```
benbz:
```suggestion
				return fmt.Errorf("configmap %s/%s is not managed by this matrix-tools-deployment-markers", namespace, name)
```

matrix-tools/internal/pkg/marker/marker.go
```diff
@@ -0,0 +1,84 @@
...
+		Labels:    labels,
+	}
+	// Fetch the existing secret or initialize an empty one
+	existingConfigMap, err := configMapsClient.Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		existingConfigMap, err = configMapsClient.Create(ctx, &corev1.ConfigMap{
+			ObjectMeta: configMapMeta, Data: nil}, metav1.CreateOptions{},
+		)
+		if err != nil {
+			return fmt.Errorf("failed to initialize secret: %w", err)
```
benbz:
```suggestion
			return fmt.Errorf("failed to initialize configmap: %w", err)
```

matrix-tools/internal/pkg/marker/marker.go
```diff
@@ -0,0 +1,84 @@
...
+func GenerateConfigMap(client kubernetes.Interface, labels map[string]string, namespace string, name string, key string, step string, newValue string, allowedValues []string) error {
+	ctx := context.Background()
+
+	configMapsClient := client.CoreV1().ConfigMaps(namespace)
+	configMapMeta := metav1.ObjectMeta{
+		Name:      name,
+		Namespace: namespace,
+		Labels:    labels,
+	}
+	// Fetch the existing secret or initialize an empty one
```
benbz:
```suggestion
	// Fetch the existing ConfigMap or initialize an empty one
```

matrix-tools/internal/pkg/args/args.go
```diff
@@ -110,26 +127,52 @@ func ParseArgs(args []string) (*Options, error) {
...
 		options.Command = GenerateSecrets
+	case "deployment-markers":
+		err := deploymentMarkersSet.Parse(args[2:])
+		if err != nil {
+			return nil, err
+		}
+		for _, deploymentMarkerArg := range strings.Split(*deploymentMarkers, ",") {
+			parsedValue := strings.Split(deploymentMarkerArg, ":")
+			if len(parsedValue) < 4 {
+				return nil, fmt.Errorf("invalid deployment marker format, expect <name:key:step:newValue:[allowedValues:..]>: %s", deploymentMarkerArg)
```
benbz:
```suggestion
				return nil, fmt.Errorf("invalid deployment marker format, expect <name:key:step:newValue:[allowedValues,..]>: %s", deploymentMarkerArg)
```

I think doing the `allowedValues` as a CSV would be better than a varargs

charts/matrix-stack/source/deployment-markers.yaml.j2
benbz, line 1: Needs `extraEnv`

matrix-tools/internal/pkg/args/args.go
gaelgatelement: I will go for the semicolon then, as the comma is already used to separate markers.

tests/integration/test_matrix_authentication_service.py
```diff
@@ -22,3 +24,41 @@ async def test_matrix_authentication_service_graphql_endpoint(ingress_ready, gen
...
     # When not authenticated, the userByUsername will return an empty result whatever the username queried
     assert json_content["data"] == {"userByUsername": None}
+
+
+@pytest.mark.skipif(value_file_has("matrixAuthenticationService.enabled", False), reason="MAS not deployed")
+@pytest.mark.asyncio_cooperative
+async def test_matrix_authentication_service_marker_delegated_auth(
+    kube_client, helm_client: pyhelm3.Client, ingress_ready, generated_data: ESSData, ssl_context
+):
+    secret = await kube_client.get(
```
benbz:
```suggestion
    configmap = await kube_client.get(
```

benbz:
To call out here as well, there's something odd with `templates-dyff` - Synapse markers manifests aren't currently showing up, on top of `nothing-enabled-values.yaml` that's already been mentioned. This could be a job problem or an actual issue but I'd like us to get to the bottom of it incase it is a real issue with these changes

.github/workflows/build-test.yml
```diff
@@ -187,6 +202,14 @@ jobs:
 
         echo -e "$header$comment_body" | tee "$RUNNER_TEMP/dyff-output"
 
+    - name: Upload generated manifests
+      if: ${{ failure() }}
```
benbz: Let's always upload them so that we can inspect newly added files if we desire. Also useful for big `dyffs` which exceed the comment limit

.github/workflows/build-test.yml
```diff
@@ -4,7 +4,7 @@
 
 name: Helm Chart Building tests
 on:
-  pull_request_target:
+  pull_request:
```

.github/workflows/build-test.yml
```diff
@@ -114,11 +114,13 @@ jobs:
         mkdir -p "$RUNNER_TEMP/new"
         for values in charts/matrix-stack/ci/*values.yaml; do
           echo "Generating new templates with $values";
+          mkdir -p "$RUNNER_TEMP/new/$(basename "$values" ".yaml")"
           helm template \
               -n ess-ci \
               -a monitoring.coreos.com/v1/ServiceMonitor \
               -f "$values" charts/matrix-stack | \
-          yq ea '[.] | sort_by(.kind, .metadata.name) | .[] | splitDoc' > "$RUNNER_TEMP/new/$(basename "$values")"
+          yq ea '[. | select(. | has("kind"))] | sort_by(.kind, .metadata.name) | .[] | splitDoc' | \
```
benbz: What doesn't have `kind` in it?

We can probably also simplify this down to something like `.[] | splitDoc` as we don't care about sorting within the file anymore

.github/workflows/build-test.yml
```diff
@@ -156,11 +160,18 @@ jobs:
...
         header="# dyff of changes in rendered templates of CI manifests\n\n"
         comment_body=""
         while read -r templates_file; do
+          if [[ "$templates_file" == ".yaml" ]]; then
+            echo "skipping empty file"
+            continue
+          fi
+
+          echo "$templates_file"
           if [ ! -f "$RUNNER_TEMP/old/$templates_file" ]; then
```
benbz, line 173: I'm wondering if we do want to include the body of added/removed templates now given it is going to be a lot more common.

.github/workflows/build-test.yml
```diff
@@ -156,11 +160,18 @@ jobs:
...
       shell: bash
       run: |
-        templates_files=$(find "$RUNNER_TEMP/old" "$RUNNER_TEMP/new" -maxdepth 1 -name '*values.yaml' | sed -E 's|'"$RUNNER_TEMP"'/(old\|new)/||' | sort | uniq)
+        echo "output_dir=$RUNNER_TEMP" | tee -a "$GITHUB_OUTPUT"
+        templates_files=$(find "$RUNNER_TEMP/old" "$RUNNER_TEMP/new" -maxdepth 2 -name '*.yaml' | sed -E 's|'"$RUNNER_TEMP"'/(old\|new)/||' | sort | uniq)
 
         header="# dyff of changes in rendered templates of CI manifests\n\n"
         comment_body=""
         while read -r templates_file; do
+          if [[ "$templates_file" == ".yaml" ]]; then
```
benbz: How does this happen?

gaelgatelement: `nothing-enabled` values file generated an empty `.yaml` file.

gaelgatelement: I've been trying to filter it out by `select(. | has("kind")` but it does not seem to work.

benbz: Ack. In which case let's remove this and find the correct way to remove empty docs from the above I think

charts/matrix-stack/templates/deployment-markers/rolebinding.yaml
```diff
@@ -0,0 +1,31 @@
...
+SPDX-License-Identifier: AGPL-3.0-only
+*/ -}}
+
+{{ range $step := list "pre" "post" -}}
+{{- with $.Values.deploymentMarkers -}}
+{{- if and .enabled .rbac.create (include "element-io.deployment-markers.markers" (dict "root" $)) }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: {{ $.Release.Name }}-deployment-markers-{{ .step }}
```
benbz:
```suggestion
  name: {{ $.Release.Name }}-deployment-markers-{{ $step }}
```

.github/workflows/build-test.yml
```diff
@@ -156,13 +160,23 @@ jobs:
...
+        echo "output_dir=$RUNNER_TEMP" | tee -a "$GITHUB_OUTPUT"
+        templates_files=$(find "$RUNNER_TEMP/old" "$RUNNER_TEMP/new" -maxdepth 2 -name '*.yaml' | sed -E 's|'"$RUNNER_TEMP"'/(old\|new)/||' | sort | uniq)
 
         header="# dyff of changes in rendered templates of CI manifests\n\n"
         comment_body=""
         while read -r templates_file; do
+          if [[ "$(basename "$templates_file")" == ".yaml" ]]; then
+            comment_body+="**$templates_file** (empty)\n\n"
+            continue
+          fi
```
benbz: This will result in `nothing-enabled-values/.yaml (empty)` being output for every PR. Maybe the thing to do here is to silently skip that one filename IIF it is also empty. Any other filename gets either an `(added)` line if new or a non-empty `nothing-enabled-values/.yaml` would get printed with `dyff` as per normal.

.github/workflows/build-test.yml
```diff
@@ -156,13 +160,23 @@ jobs:
...
+
           if [ ! -f "$RUNNER_TEMP/old/$templates_file" ]; then
-            comment_body+="**$templates_file** (added)\n\n"
+            file_content=$(cat "$RUNNER_TEMP/new/$templates_file")
+            comment_body+="<details><summary><b>$templates_file</b> (added)</summary>\n"
+            comment_body+='\n```diff\n'
+            comment_body+=$file_content
```
benbz, line 177: So this isn't going to result in a highlighted diff (as no header, no leading `+`, etc). Probably not worth worrying about but just to call it out

benbz (approved):

PR #490, opened 2025-05-27
add a symbolic link to example-values

gaelgatelement:


gaelgatelement:
We drop this idea, as the symbolic link is not very helpful in the end.

PR #494, opened 2025-05-27
Correct initial-synchrotron routing

benbz:
The service was previously `<release_name>-synapse-initial-synchrotron` but HAProxy was expecting `<release_name>-synapse-initial-sync` as per the migrated config.

This revisited the max name length restrictions around `StatefulSets` vs other kinds.

Validation in the chart itself on the max length of `Release.Name` will come in a separate PR.


charts/matrix-stack/templates/synapse/_synapse_details.tpl
```diff
@@ -78,8 +78,24 @@ canFallback
...
+{{- else if eq . "federation-inbound" -}}
+fed-innbound
+{{- else if eq . "federation-reader" -}}
+fed-reader
+{{- else if eq . "federation-sender" -}}
+fed-sender
+{{- else if eq . "initial-synchrotron" -}}
 initial-sync
+{{- else if eq . "media-repository" -}}
+media-repo
```
gaelgatelement, line 92: Shouldnt this change in `charts/matrix-stack/configs/synapse/synapse-04-homeserver-overrides.yaml.tpl` 

```
media_instance_running_background_jobs: "{{ $root.Release.Name }}-synapse-media-repository-0"
```
to

```
media_instance_running_background_jobs: "{{ $root.Release.Name }}-synapse-media-repo-0"
```

?

benbz: Hmm true as that's going to come from `hostname` eventually. The stream writers/instance map stuff needs adjusting too I think

gaelgatelement (approved):

PR #510, opened 2025-06-03
Fix dyff job

gaelgatelement:


.github/workflows/build-test.yml
```diff
@@ -4,7 +4,7 @@
 
 name: Helm Chart Building tests
 on:
-  pull_request_target:
+  pull_request:
```

benbz (approved):

PR #513, opened 2025-06-04
synapse: configure matrixRTC values

gaelgatelement:
According to https://github.com/element-hq/element-call/blob/livekit/docs/self-hosting.md

### DIFF
--- a/newsfragments/513.changed.md
+++ b/newsfragments/513.changed.md
@@ -1,0 +1,1 @@
+Configure Synapse appropriately for Element Call when matrixRTC is enabled.
--- a/charts/matrix-stack/configs/synapse/synapse-01-shared-underrides.yaml.tpl
+++ b/charts/matrix-stack/configs/synapse/synapse-01-shared-underrides.yaml.tpl
@@ -8,3 +8,21 @@ SPDX-License-Identifier: AGPL-3.0-only
 report_stats: false
 
 require_auth_for_profile_requests: true
+
+{{- if $root.Values.matrixRTC.enabled }}
+# The maximum allowed duration by which sent events can be delayed, as
+# per MSC4140.
+max_event_delay_duration: 24h
+
+rc_message:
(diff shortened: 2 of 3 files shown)


newsfragments/513.changed.md
```diff
@@ -0,0 +1 @@
+Synapse: Automatically configure MSC3266, MSC4222 and set duration and rate limits according to MSC4140 when matrixRTC is enabled.
```
benbz:
```suggestion
Configure Synapse appropriately for Element Call when matrixRTC is enabled.
```

I don't think we need to go into too much detail on the specifics, the outcome is what we care about

benbz (approved):
Minor changelog comment.

I wonder if we also need to be doing things with the ingress (depending on the controller type) for the SFU websocket

gaelgatelement:
> Minor changelog comment.
> 
> I wonder if we also need to be doing things with the ingress (depending on the controller type) for the SFU websocket

As we cannot set nginx annotations for subpaths only, how would you handle this ?

Regarding the values, I'd set automatically for nginx only the timeouts : 

- `proxy_send_timeout 120` and `proxy_read_timeout 120` 
- `proxy_buffering off` is [already disabled by default](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#proxy-buffering)

benbz:
> > Minor changelog comment.
> > I wonder if we also need to be doing things with the ingress (depending on the controller type) for the SFU websocket
> 
> As we cannot set nginx annotations for subpaths only, how would you handle this ?
> 
> Regarding the values, I'd set automatically for nginx only the timeouts :
> 
>     * `proxy_send_timeout 120` and `proxy_read_timeout 120`
> 
>     * `proxy_buffering off` is [already disabled by default](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#proxy-buffering)

https://www.civo.com/learn/using-websockets-with-ingress-controller we could use a server snippet :see_no_evil: and include a `location` block

PR #514, opened 2025-06-04
SFU: Fix ingress needs to set proxy-timeouts and disable buffering

gaelgatelement:
According to https://github.com/element-hq/element-call/blob/livekit/docs/self-hosting.md#matrix-site-endpoint-routing

### DIFF
--- a/charts/matrix-stack/templates/matrix-rtc/ingress.yaml
+++ b/charts/matrix-stack/templates/matrix-rtc/ingress.yaml
@@ -1,5 +1,5 @@
 {{- /*
-Copyright 2024 New Vector Ltd
+Copyright 2024-2025 New Vector Ltd
 
 SPDX-License-Identifier: AGPL-3.0-only
 */ -}}
@@ -8,7 +8,15 @@ SPDX-License-Identifier: AGPL-3.0-only
 apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
-{{- include "element-io.ess-library.ingress.annotations" (dict "root" $ "context" (dict "ingress" .ingress)) | nindent 2 }}
+{{- $extraAnnotations := dict }}
+{{- if .sfu.enabled }}
+{{- if eq "ingress-nginx" (include "element-io.ess-library.ingress-controller-type" (dict "root" $ "context" .ingress.controllerType)) }}
+{{- $_ := set $extraAnnotations "nginx.ingress.kubernetes.io/proxy-send-timeout" "120" }}
+{{- $_ := set $extraAnnotations "nginx.ingress.kubernetes.io/proxy-read-timeout" "120" }}
+{{- $_ := set $extraAnnotations "nginx.ingress.kubernetes.io/proxy-buffering" "off" }}
+{{- end }}
+{{- end }}
+{{- include "element-io.ess-library.ingress.annotations" (dict "root" $ "context" (dict "ingress" .ingress "extraAnnotations" $extraAnnotations)) | nindent 2 }}
   labels:
     {{- include "element-io.matrix-rtc-ingress.labels" (dict "root" $ "context" .) | nindent 4 }}
   name: {{ $.Release.Name }}-matrix-rtc
--- a/newsfragments/514.fixed.md
+++ b/newsfragments/514.fixed.md
@@ -1,0 +1,1 @@
+Matrix RTC: Set proxy timeout and enforce disabled buffering `nginx-ingress` `controllerType` annotations if SFU is enabled.


README.md
```diff
@@ -378,7 +378,40 @@ server {
...
+
+    server_name mrtc.example.com;
+
+    location / {
+        proxy_pass http://127.0.0.1:8080;
+        proxy_set_header X-Forwarded-For $remote_addr;
+        proxy_set_header X-Forwarded-Proto $scheme;
+        proxy_set_header Host $host;
+
         proxy_http_version 1.1;
```
benbz: These lines should only be on the SFU path, where as the lines above should be for all paths

charts/matrix-stack/source/common/sub_schema_values.yaml.j2
```diff
@@ -169,7 +169,9 @@ imagePullSecrets: []
   service: {}
 {%- endif %}
   ## If set, some tweaks will be applied automatically to ingresses based on the controller type here.
-  ## This can be set to `ingress-nginx`.
+  ## This can be set to `ingress-nginx` or `ingress-nginx-with-snippets`.
+  ## `ingress-nginx` relies on default ingress-nginx configuration
+  ## `ingress-nginx-with-snippets` relies on ingress-nginx configuration with `allow-snippet-annotations` enabled.
```
benbz:
```suggestion
  ## `ingress-nginx-with-snippets` relies on ingress-nginx configuration with `allow-snippet-annotations` which is not enabled by default.
```

charts/matrix-stack/configs/matrix-rtc/sfu/ingress-nginx-server-snippet.conf.tpl
```diff
@@ -0,0 +1,18 @@
+{{- /*
+Copyright 2025 New Vector Ltd
+
+SPDX-License-Identifier: AGPL-3.0-only
+*/ -}}
+
+location / {
```
benbz: This will also match `/sfu/get`. Given longest match wins I think we should add a separate `location /sfu/get` block that does the (probably not-necessary) `proxy_set_header X-Forwarded...` lines so that we don't get the upgrade lines

charts/matrix-stack/configs/matrix-rtc/sfu/ingress-nginx-server-snippet.conf.tpl
```diff
@@ -0,0 +1,18 @@
...
+    proxy_set_header Upgrade $http_upgrade;
+    proxy_http_version 1.1;
```
benbz: Not part of https://github.com/element-hq/element-call/blob/livekit/docs/self-hosting.md#matrix-site-endpoint-routing ?

charts/matrix-stack/configs/matrix-rtc/sfu/ingress-nginx-server-snippet.conf.tpl
```diff
@@ -0,0 +1,18 @@
...
+    proxy_set_header X-Forwarded-Host $http_host;
+    proxy_set_header X-Forwarded-Proto $scheme;
+    proxy_set_header X-Forwarded-For $remote_addr;
+    proxy_set_header Host $host;
+
+    proxy_set_header Accept-Encoding gzip;
+    proxy_set_header Connection "upgrade";
+    proxy_cache_bypass $http_upgrade;
```
benbz: Not part of https://github.com/element-hq/element-call/blob/livekit/docs/self-hosting.md#matrix-site-endpoint-routing ?

charts/matrix-stack/templates/matrix-rtc/ingress.yaml
```diff
@@ -8,7 +8,16 @@ SPDX-License-Identifier: AGPL-3.0-only
...
-{{- include "element-io.ess-library.ingress.annotations" (dict "root" $ "context" (dict "ingress" .ingress)) | nindent 2 }}
+{{- $extraAnnotations := dict }}
+{{- if hasPrefix "ingress-nginx" (include "element-io.ess-library.ingress-controller-type" (dict "root" $ "context" .ingress.controllerType)) }}
+{{- $_ := set $extraAnnotations "nginx.ingress.kubernetes.io/proxy-send-timeout" "120" }}
+{{- $_ := set $extraAnnotations "nginx.ingress.kubernetes.io/proxy-read-timeout" "120" }}
+{{- $_ := set $extraAnnotations "nginx.ingress.kubernetes.io/proxy-buffering" "off" }}
+{{- end }}
+{{- if eq (include "element-io.ess-library.ingress-controller-type" (dict "root" $ "context" .ingress.controllerType)) "ingress-nginx-with-snippets" }}
+{{- $_ := set $extraAnnotations "nginx.ingress.kubernetes.io/server-snippets" (tpl ($.Files.Get "configs/matrix-rtc/sfu/ingress-nginx-server-snippet.conf.tpl") $) }}
+{{- end }}
```
benbz: This should all be gated on the SFU being enabled. It isn't required for just the authoriser

charts/matrix-stack/configs/matrix-rtc/sfu/ingress-nginx-server-snippet.conf.tpl
gaelgatelement: But doesn't the `ingress-controller` already generates `location /sfu/get` due to v ?

```yml
      paths:
      - path: /sfu/get
``` 

I suppose I need to look into what is actually rendered in the pytest env.

benbz: There's 2 paths on the ingress though `/` and `/sfu/get`. So I'd already expect 2x `location` blocks and for the ingress controller to merge them

gaelgatelement: It turns out the ingress-controller already has the proper values on all locations :

```

                        # Allow websocket connections
                        proxy_set_header                        Upgrade           $http_upgrade;

                        proxy_set_header                        Connection        $connection_upgrade;


```

PR #515, opened 2025-06-04
ess-library: handle replicas & maxUnavailable using a helper

gaelgatelement:


charts/matrix-stack/source/matrixAuthenticationService.json
```diff
@@ -44,6 +44,9 @@
         }
       }
     },
+    "replicas": {
+      "type": "number"
```
benbz:
```suggestion
      "type": "integer"
```

benbz (commented):
It would be good to get a manifest test of `replicas` being settable and the `maxUnavailable` behaviour. I won't block this PR on that though

benbz (approved):

tests/manifests/__init__.py
```diff
@@ -343,29 +348,29 @@ def make_synapse_worker_sub_component(worker_name: str) -> SubComponentDetails:
...
-        "sso-login",
-        "synchrotron",
-        "typing-persister",
-        "user-dir",
-    ]
+    make_synapse_worker_sub_component(worker_name, worker_type)
+    for worker_name, worker_type in {
+        "appservice": "single",
+        "background": "single",
+        "client-reader": "static",
```
benbz:
```suggestion
        "client-reader": "scalable",
```

And on all others

tests/manifests/test_deployments.py
```diff
@@ -0,0 +1,68 @@
...
+@pytest.mark.asyncio_cooperative
+async def test_all_deployments_set_replicas(values, make_templates):
+    deployments = []
+    for template in await make_templates(values):
+        if template["kind"] == "Deployment":
+            deployments.append(template)
+
+    for deployment in deployments:
+        id = deployment["metadata"]["name"]
+        assert "replicas" in deployment["spec"], f"{id} does not specify replicas"
```
benbz:
```suggestion
async def test_all_deployments_set_replicas(templates):
    for template in templates:
        if template["kind"] in ["Deployment", "StatefulSet"]:
          assert "replicas" in template["spec"], f"{template_id(template)} does not specify replicas"
```

As this is a static test that isn't manipulating values and because `StatefulSets` should have this property too

tests/manifests/test_deployments.py
```diff
@@ -0,0 +1,68 @@
...
+
+def assert_matching_replicas(template, values):
+    deployable_details = template_to_deployable_details(template)
+    replicas = template["spec"]["replicas"]
+    max_unavailable = template["spec"]["strategy"]["rollingUpdate"]["maxUnavailable"]
+    if (
+        not deployable_details.values_file_path_overrides
+        or deployable_details.values_file_path_overrides
+        and deployable_details.values_file_path_overrides[PropertyType.Replicas] != ValuesFilePath.not_supported()
+    ):
```
benbz: I think all this is telling you that there should be an `is_scalable`/`has_replicas` flag on `DeployableDetails` rather than inferring it from the `values_file_path_overrides` which we don't do anywhere else

tests/manifests/test_deployments.py
benbz, line 1: I think there should also be a test setting `replicas: 1` for everything so that we know that `maxUnavailable: 0` is set for everything configurable and not just for things that don't have configurable `replicas`

tests/manifests/test_statefulset.py
```diff
@@ -7,6 +7,19 @@
...
+@pytest.mark.asyncio_cooperative
+async def test_all_statefulsets_set_replicas(templates):
+    statefulsets = []
+    for template in templates:
+        if template["kind"] in ["StatefulSet"]:
+            statefulsets.append(template)
+
+    for sts in statefulsets:
+        id = sts["metadata"]["name"]
+        assert "replicas" in sts["spec"], f"{id} does not specify replicas"
```
benbz:
```suggestion
    for template in templates:
        if template["kind"] in ["StatefulSet"]:
           assert "replicas" in template["spec"], f"{template_id(template)} does not specify replicas"
```

tests/manifests/test_deployments.py
```diff
@@ -0,0 +1,85 @@
...
+@pytest.mark.asyncio_cooperative
+async def test_all_deployments_set_replicas(templates):
+    deployments = []
+    for template in templates:
+        if template["kind"] in ["Deployment"]:
+            deployments.append(template)
+
+    for deployment in deployments:
+        id = deployment["metadata"]["name"]
+        assert "replicas" in deployment["spec"], f"{id} does not specify replicas"
```
benbz:
```suggestion
    for template in templates:
        if template["kind"] in ["Deployment"]:
            assert "replicas" in template["spec"], f"{template_id(template)} does not specify replicas"
```

tests/manifests/__init__.py
```diff
@@ -111,6 +113,12 @@ def __post_init__(self):
             self.has_service_monitor = self.has_workloads
         if self.has_topology_spread_constraints is None:
             self.has_topology_spread_constraints = self.has_workloads
+        if self.has_replicas is None:
+            self.has_replicas = self.has_workloads and (
+                not self.values_file_path_overrides
+                or self.values_file_path_overrides
+                and self.values_file_path_overrides.get(PropertyType.Replicas) != ValuesFilePath.not_supported()
+            )
```
benbz:
```suggestion
            self.has_replicas = self.has_workloads
```

And then set `has_replicas` explicitly on sidecars, jobs and things that don't support it. You shouldn't need to interact with `values_file_path_overrides` at all

tests/manifests/test_deployments.py
```diff
@@ -0,0 +1,82 @@
...
+        if template["kind"] in ["Deployment"]:
+            assert_matching_replicas(template, values)
+
+
+@pytest.mark.parametrize("values_file", values_files_to_test)
+@pytest.mark.asyncio_cooperative
+async def test_max_unaivalable_single_replicas(values, make_templates):
+    iterate_deployables_workload_parts(
+        lambda deployable_details: deployable_details.set_helm_values(values, PropertyType.Replicas, 1)
+    )
```
benbz:
```suggestion
    iterate_deployables_parts(
        lambda deployable_details: deployable_details.set_helm_values(values, PropertyType.Replicas, 1),
        lambda deployable_details: deployable_details.has_replicas
    )
```

So that you don't need to specify `values_file_path_overrides` for anything not supporting it

tests/manifests/test_deployments.py
```diff
@@ -0,0 +1,82 @@
...
+    # That way we can assert a) the correct value is going into the correct field and
+    # b) that the correct part of the values file is being used
+    counter = 100
+
+    def set_replicas_details(deployable_details: DeployableDetails):
+        nonlocal counter
+        counter += 1
+        deployable_details.set_helm_values(values, PropertyType.Replicas, counter)
+
+    iterate_deployables_workload_parts(set_replicas_details)
```
benbz:
```suggestion
    iterate_deployables_parts(set_replicas_details, lambda deployable_details: deployable_details.has_replicas)
```

tests/manifests/__init__.py
```diff
@@ -111,6 +113,11 @@ def __post_init__(self):
             self.has_service_monitor = self.has_workloads
         if self.has_topology_spread_constraints is None:
             self.has_topology_spread_constraints = self.has_workloads
+        self.has_replicas = self.has_workloads and (
+            not self.values_file_path_overrides
+            or self.values_file_path_overrides
+            and self.values_file_path_overrides.get(PropertyType.Replicas) != ValuesFilePath.not_supported()
+        )
```
benbz:
```suggestion
        self.has_replicas = self.has_workloads 
```

And then set `has_replicas = False` on sidecars, jobs and anything that legitimately doesn't support it. You won't need to specify `values_file_path_overrides` then.

benbz (approved):

PR #523, opened 2025-06-06
Dyff comment : truncate added files

gaelgatelement:


.github/workflows/build-test.yml
```diff
@@ -4,7 +4,7 @@
 
 name: Helm Chart Building tests
 on:
-  pull_request_target:
+  pull_request:
```

charts/matrix-stack/templates/matrix-authentication-service/configmap_added.yaml
benbz, line 1: To be deleted before merging

PR #524, opened 2025-06-06
Pytest: Test helm upgrades

gaelgatelement:


.github/workflows/pytest.yml
```diff
@@ -5,7 +5,7 @@
 name: Run PyTest tests - Integration and Manifests
 
 on:
-  pull_request_target:
+  pull_request:
```

tests/integration/artifacts/certs.py
```diff
@@ -59,6 +59,17 @@ def key_as_pem(self):
             encryption_algorithm=serialization.NoEncryption(),
         ).decode("utf-8")
 
+    @classmethod
+    def from_dict(cls, kv):
+        return CertKey(
+            ca=CertKey.from_dict(kv["ca"]) if kv["ca"] else None,
```
benbz, line 66: We always expect `cert` and `key` to be present. Why can't we expect `ca` to be present?

.github/workflows/pytest.yml
```diff
@@ -79,13 +91,34 @@ jobs:
         skipClusterCreation: "true"
         skipClusterLogsExport: "true"
 
-
     - uses: azure/setup-helm@b9e51907a09c216f16ebe8536097933489208112  # v4.3.0
 
-    - name: Test with pytest
+    - name: Checkout values files from previous ref
+      run: |
+        git checkout "${MATRIX_TEST_FROM_REF}" -- charts/matrix-stack/ci tests/integration/env/
```
benbz: Hmm so this is going to test the old values with the new chart templates & new test python. That may/may not work. Should we not test old values + old chart templates + old test python and then test new values + new chart templates + new test python instead?

tests/integration/artifacts/certs.py
gaelgatelement: > We always expect cert and key to be present. Why can't we expect ca to be present

The Root CA is self-signed and does not have a CA.

.github/workflows/pytest.yml
gaelgatelement: I thought about it, but it will cause us to re-install the poetry dependencies between the two pytest calls. This might cause caching issues.

I figured testing the old chart with the new python should work as the dependency between helm and python is very light. It's mostly treated as an external binary call, so we should always be able to figure out a way to make the old chart work with the new python if required.

tests/integration/artifacts/certs.py
benbz: Makes sense. Why was the type being non-nullable before acceptable?

.github/workflows/pytest.yml
benbz: I think at the very least we've got to test with the old chart templates too

tests/integration/artifacts/certs.py
gaelgatelement: I think it was just wrongly typed as the actual logic expected it to be `None` : 

```

    def cert_bundle_as_pem(self):
        bundle = []
        bundle.append(self.cert.public_bytes(encoding=serialization.Encoding.PEM).decode("utf-8"))
        if self.ca is not None:
            bundle.append(self.ca.cert_bundle_as_pem())
        return "".join(bundle)
```

.github/workflows/pytest.yml
gaelgatelement: > I think at the very least we've got to test with the old chart templates too

Uh :facepalm:  yeah

tests/integration/artifacts/certs.py
benbz: I guess we don't run `mypy` against `test/integrations` yet

benbz (approved):

PR #525, opened 2025-06-06
Run mypy on integration tests

gaelgatelement:


.github/workflows/scripts-linting.yml
```diff
@@ -5,7 +5,7 @@
 name: Scripts linting
 
 on:
-  pull_request_target:
+  pull_request:
```
benbz: To be reverted before merging

tests/integration/fixtures/__init__.py
```diff
@@ -10,22 +10,22 @@
...
-    ingress_ready,
-    kube_client,
-    loaded_matrix_tools,
-    matrix_stack,
-    prometheus_operator_crds,
-    registry,
-    secrets_generated,
-    ssl_context,
-    users,
+    "build_matrix_tools",
```
benbz, line 13: /me makes unhappy face given this was deliberately changed away from strings https://github.com/element-hq/ess-helm/commit/083474b10c1d21fdc3deff0da3eca5809644172f

gaelgatelement: Yeah but :(

`tests/integration/fixtures/__init__.py:12: error: Type of __all__ must be "Sequence[str]", not "list[object]"  [misc]`

I could also ignore it if we prefer to.

gaelgatelement: But well python says it must be strings : https://docs.python.org/3/tutorial/modules.html#importing-from-a-package

benbz: No let's leave it given https://stackoverflow.com/questions/48755622/can-all-be-list-of-objects-rather-than-strings makes it sound like strings are indeed correct

PR #526, opened 2025-06-06
syn2mas: fail if pods are still up after statefulset are all down

gaelgatelement:
If for any reason kubernetes is misbehaving, we might have unsynced issue which might turn wrong. Let's just be extra safe with the process.

### DIFF
--- a/matrix-tools/internal/pkg/syn2mas/syn2mas.go
+++ b/matrix-tools/internal/pkg/syn2mas/syn2mas.go
@@ -59,6 +59,19 @@ func scaleDownSynapse(client kubernetes.Interface, namespace string) map[string]
 			break
 		}
 	}
+	podsClient := client.CoreV1().Pods(namespace)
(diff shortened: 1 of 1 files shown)


tests/integration/env/matrix-authentication-service-syn2mas.rc
```diff
@@ -2,4 +2,5 @@
 #
 # SPDX-License-Identifier: AGPL-3.0-only
 
-export TEST_VALUES_FILE=charts/matrix-stack/ci/pytest-matrix-authentication-service-syn2mas-values.yaml
\ No newline at end of file
+export TEST_VALUES_FILE=charts/matrix-stack/ci/pytest-matrix-authentication-service-syn2mas-values.yaml
+export BUILD_MATRIX_TOOLS=1
```
benbz: Needs reverting before merging

PR #527, opened 2025-06-06
syn2mas: the configmap hook is not used in dryRun mode

gaelgatelement:


gaelgatelement:
> Also `charts/matrix-stack/templates/matrix-authentication-service/mas_secret_hook.yaml`?

The secret is required for Synapse Check Config though!

gaelgatelement:
> > Also `charts/matrix-stack/templates/matrix-authentication-service/mas_secret_hook.yaml`?
> 
> The secret is required for Synapse Check Config though!

Oh but it's not used before MAS actually migrates !

charts/matrix-stack/templates/matrix-authentication-service/secret_hook.yaml
```diff
@@ -9,7 +9,7 @@ SPDX-License-Identifier: AGPL-3.0-only
   (or (and $.Values.synapse.checkConfigHook.enabled
         (include "element-io.matrix-authentication-service.synapse-secret-data" (dict "root" $ "context" .))
         )
-      .syn2mas.enabled
+       .syn2mas.enabled (not .syn2mas.dryRun)
```
benbz:
```suggestion
       (and .syn2mas.enabled (not .syn2mas.dryRun))
```

I think as we're inside the `or` only at this point

newsfragments/527.internal.md
```diff
@@ -0,0 +1 @@
+Make sure MAS Secret & ConfigMap hooks are created only when required.
```
benbz: Given `syn2mas` hasn't been in a release yet, let's just c/p that changelog fragment here as this is a fix-up to that PR. Then both PRs will be linked to the changelog entry

PR #530, opened 2025-06-09
On chart version upgrade, expect 429s to occur

gaelgatelement:


.github/workflows/pytest.yml
```diff
@@ -5,7 +5,7 @@
 name: Run PyTest tests - Integration and Manifests
 
 on:
-  pull_request_target:
+  pull_request:
```

benbz (approved):

PR #533, opened 2025-06-09
Rename Matrix RTC Authorisation Service templates

benbz:
Though `matrix-rtc/authorisation_service_service.yaml` is a bit of an annoying name. Could go with `s/authorisation_service/authoriser/`?


gaelgatelement:
> Though `matrix-rtc/authorisation_service_service.yaml` is a bit of an annoying name. Could go with `s/authorisation_service/authoriser/`?

Authoriser feels wrong as we explicitly went away from naming it like this.

I'd be fine with `mrtc_authorisation(s)?`

benbz:
> > Though `matrix-rtc/authorisation_service_service.yaml` is a bit of an annoying name. Could go with `s/authorisation_service/authoriser/`?
> 
> Authoriser feels wrong as we explicitly went away from naming it like this.

We went away from it for anything public, this is just internal. 

> I'd be fine with `mrtc_authorisation(s)?`

Not a huge fan of `mrtc_` given we're in the `matrix-rtc` folder. How about just `authorisation`?

gaelgatelement:
> > > Though `matrix-rtc/authorisation_service_service.yaml` is a bit of an annoying name. Could go with `s/authorisation_service/authoriser/`?
> > 
> > 
> > Authoriser feels wrong as we explicitly went away from naming it like this.
> 
> We went away from it for anything public, this is just internal.
> 
> > I'd be fine with `mrtc_authorisation(s)?`
> 
> Not a huge fan of `mrtc_` given we're in the `matrix-rtc` folder. How about just `authorisation`?

Let's do `authorisation`.

gaelgatelement (approved):

PR #540, opened 2025-06-10
update ci to check anti-affinity against kubeconform

gaelgatelement:


charts/matrix-stack/ci/pytest-synapse-values.yaml
benbz, line 1: Let's do the workers in `charts/matrix-stack/ci/fragments/synapse-pytest-self-extras.yaml` as well. They don't currently have antiAffinity but could get it in future

PR #546, opened 2025-06-10
matrix-tools syn2mas: filter completed pods

gaelgatelement:
When trying out the migration on my own HS, I had to delete a job created in an older released, which was suffixed with `-hook`.

Filtering out completed and failed pods should be more reliable.


tests/integration/env/matrix-authentication-service-syn2mas.rc
```diff
@@ -2,4 +2,5 @@
 #
 # SPDX-License-Identifier: AGPL-3.0-only
 
-export TEST_VALUES_FILE=charts/matrix-stack/ci/pytest-matrix-authentication-service-syn2mas-values.yaml
\ No newline at end of file
+export TEST_VALUES_FILE=charts/matrix-stack/ci/pytest-matrix-authentication-service-syn2mas-values.yaml
+exprort BUILD_MATRIX_TOOLS=1
```
benbz: To be reverted before merging

benbz (approved):

PR #547, opened 2025-06-10
Improve upgrade test tag detection

benbz:
We may not want to test against the most recently created tag if that isn't in the history for a given commit. We want the nearest reachable tag.

Does `--match` so that we exclude anything that doesn't look like a release or is `matrix-tools-<version>`


.github/workflows/pytest.yml
```diff
@@ -5,7 +5,7 @@
 name: Run PyTest tests - Integration and Manifests
 
 on:
-  pull_request_target:
```

Issue #556, opened 2025-06-12
RTCSession Error if synapse workers enabled


watermelon0339:
**Describe the bug**
After enabling Synapse workers, Element Call stops working and reports an RTCSession error.

**To Reproduce**
1. Edit `hostnames.yaml` to enable synapse workers:
   ```yml
      synapse:
        ingress:
          host: matrix.xxx.xxx
        resources:
          limits:
            memory: 12Gi
        media:
          maxUploadSize: 50M
        workers:
          background:
            enabled: true
          client-reader:
            enabled: true
            replicas: 1
          encryption:
            enabled: true
          event-creator:
            enabled: true
            replicas: 1
          event-persister:
            enabled: true
            replicas: 2
          initial-synchrotron:
            enabled: true
            replicas: 2
          media-repository:
            enabled: true
          presence-writer:
            enabled: true
          push-rules:
            enabled: true
          pusher:
            enabled: true
            replicas: 2
          receipts-account:
            enabled: true
          sliding-sync:
            enabled: true
            replicas: 2
          synchrotron:
            enabled: true
            replicas: 1
          typing-persister:
            enabled: true
          user-dir:
            enabled: true
   ```
2. Re-deploy helm chart `helm upgrade ......`
3. Make a Element Call by ElementX App
4. Error occured:
   ```
    RTCSession Error
    The MembershipManager shut down because of the end condition: Error: Could not restart delayed event, even though delayed events are supported. M_UNRECOGNIZED: MatrixError: [404] Unrecognized request (https://matrix.mtx01.cc/_matrix/client/unstable/org.matrix.msc4140/delayed_events/syd_aRnWkIJhnZGrpKFjazgR)
   ```

**Expected behavior**
Element Call should work.

**Screenshots**

![Image](https://github.com/user-attachments/assets/2a53e2e5-4742-48a3-8473-bb1e32c7e9ed)

**Desktop (please complete the following information):**
 - App: Element X 25.06.02
 - Browser: edge

**Smartphone (please complete the following information):**
 - Device: Android
 - OS: android 11
 - App: Element X 25.06.02

**Additional context**
 - ess-helm version: 25.6.1

watermelon0339:
```
POST /_matrix/client/unstable/org.matrix.msc4140/delayed_events/syd_PEDCpFvsRFtxThipJQGV

{
  "errcode": "M_UNRECOGNIZED",
  "error": "Unrecognized request"
}
```

PR #557, opened 2025-06-13
matrixrtc: fix error when workers are enabled

gaelgatelement:
Fixes https://github.com/element-hq/ess-helm/issues/556


newsfragments/557.fixed.md
```diff
@@ -0,0 +1 @@
+Fix MatrixRTC RTCSession Error if synapse workers enabled.
```
benbz:
```suggestion
Fix MatrixRTC RTCSession Error if a `push-rules` Synapse worker is enabled.
```

PR #569, opened 2025-06-23
matrix-rtc: fix services monitors

gaelgatelement:
In the room, it was raised that matrix-rtc services monitors were wrong : https://matrix.to/#/!RkXKfjpGXIVObpCqOI:element.io/$DM2GN8exEvvq69WPVVVBMJf-1G_E0_35G7pSB0nkhCw?via=matrix.org&via=element.io&via=k8ekat.dev

1.  This drops the Authoriser Service Monitor as it does not have any metrics
2. This fixes the port of t he SFU Service Monitor


benbz (approved):
This all looks plausible but there's no automated way of knowing whether this is correct.

How tricky would it be to have an integration test that looks at all `ServiceMonitors`, looks at all matching `Services`, poke `/metrics` on all `Endpoints` on said `Services` and assert all have successful responses with Prometheus/OpenTelemetry looking responses

gaelgatelement:
> This all looks plausible but there's no automated way of knowing whether this is correct.
> 
> How tricky would it be to have an integration test that looks at all `ServiceMonitors`, looks at all matching `Services`, poke `/metrics` on all `Endpoints` on said `Services` and assert all have successful responses with Prometheus/OpenTelemetry looking responses

The difficult part is that we have to do some `ports-forward` tricks as we cant access those endpoints from the host network.

benbz:
> > This all looks plausible but there's no automated way of knowing whether this is correct.
> > How tricky would it be to have an integration test that looks at all `ServiceMonitors`, looks at all matching `Services`, poke `/metrics` on all `Endpoints` on said `Services` and assert all have successful responses with Prometheus/OpenTelemetry looking responses
> 
> The difficult part is that we have to do some `ports-forward` tricks as we cant access those endpoints from the host network.

Makes sense. Or `kubectl exec` a cURL command would also work I suppose

tests/integration/test_networking.py
```diff
@@ -162,3 +163,58 @@ async def test_pods_monitored(
...
+            assert service.spec.ports, f"Ecountered a service without port : {service}"
+            assert service.spec.selector, f"Ecountered a service without selectors : {service}"
+
+            for endpoint in service_monitor["spec"]["endpoints"]:
+                service_port_names = [port.name for port in service.spec.ports if port.name]
+                if endpoint["port"] in service_port_names:
+                    break
+            # This Service does not have the named port. Potentially there's another Service that covers it
+            else:
+                continue
```
benbz: OOI do we ever actually hit this?

charts/matrix-stack/configs/matrix-rtc/sfu/config-overrides.yaml.tpl
```diff
@@ -9,8 +9,7 @@ SPDX-License-Identifier: AGPL-3.0-only
 
 port: 7880
 
-prometheus:
-  port: 6789
+prometheus_port: 6789
```
benbz: This is deprecated syntax and we switched away from it ( https://github.com/element-hq/ess-helm/pull/550 ). Does it not work correctly?

tests/integration/test_networking.py
gaelgatelement: I dropped it as it should not happen.

benbz: It is still here. You've dropped it from `test_pods_monitored` though

charts/matrix-stack/configs/matrix-rtc/sfu/config-overrides.yaml.tpl
gaelgatelement: It does not indeed. Where did we see it got deprecated ? The config-sample.yaml of the livekit repo still uses this syntax : https://github.com/livekit/livekit/blob/master/config-sample.yaml#L145

benbz: The logging in the `Pod` itself (example line in #550)

benbz: https://github.com/livekit/livekit/blob/master/pkg/service/server.go#L156-L172

benbz: It does seem to be there and should be used:
* https://github.com/livekit/livekit/blob/v1.9.0/pkg/config/config.go#L56-L58
* https://github.com/livekit/livekit/blob/v1.9.0/pkg/config/config.go#L287-L291
* https://github.com/livekit/livekit/blob/v1.9.0/pkg/service/server.go#L161-L172

Did I typo it in some way?

benbz (approved):

PR #589, opened 2025-06-30
Some CI fixes after pull_request switch

benbz:
Fixes up from #586 as discovered with #588.

The changes to `dyff` jobs can't be tested as the new `workflow_run` workflow doesn't exist on the default branch yet. I *think* once it exists I'll be able to modify it in PRs, so might need to merge and iterate quickly


newsfragments/589.changed.md
```diff
@@ -0,0 +1 @@
+Don't push chart OCI images for every PR, only tags and main.
```
gaelgatelement:
```suggestion
Don't push chart OCI images for every PR, only tags, main, and manual workflows triggers.
```

benbz: Do you have objections to me going the other way and doing the below?

```suggestion
Don't push chart OCI images for every PR.
```

PR #595, opened 2025-07-02
ci: dont check services matching terminating pods

gaelgatelement:
Let's see if this removes the flakes we have been seeing on components version upgrades.


tests/integration/test_networking.py
```diff
@@ -40,6 +40,8 @@ async def test_services_have_matching_labels(
         label_selectors = {label: value for label, value in service.spec.selector.items()}
 
         async for pod in kube_client.list(Pod, namespace=generated_data.ess_namespace, labels=label_selectors):
+            if pod.status and pod.status.phase == "Terminating":
```
benbz, line 43: Or actually it may be better to skip anything that isn't `Running`

gaelgatelement: In practice this does not seem to fix the issue. I'm going to add more logs in the assert message.

PR #598, opened 2025-07-02
service/pods labels matching tests: also skip succeeded

gaelgatelement:
We were still seeing some flakes 

> E AssertionError: pytest-ig3ti0a4-element-web-6467f678d9-gzg9k does not have the correct label [app.kubernetes.io/version=v1.11.105](http://app.kubernetes.io/version=v1.11.105) (pod status phase : Succeeded

### DIFF
--- a/tests/integration/test_networking.py
+++ b/tests/integration/test_networking.py
@@ -58,8 +58,10 @@ async def test_services_have_matching_labels(
                 assert value.startswith(
                     pod.metadata.labels[label.replace("k8s.element.io/target-", "app.kubernetes.io/")]
                 ), (
-                    f"{value} does not starts with "
-                    f"{pod.metadata.labels[label.replace('k8s.element.io/target-', 'app.kubernetes.io/')]}"
+                    f"Service/{service.metadata.name} has {label=}={value=} "
+                    f"which does not start with the value from the Pod/{pod.metadata.name} "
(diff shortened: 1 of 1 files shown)


newsfragments/598.internal.md
```diff
@@ -0,0 +1 @@
+Tests : Dont check services matching labels against terminating pods.
```
benbz:
```suggestion
Tests: don't check services matching labels against terminating pods.
```

As per changes in #597

tests/integration/test_networking.py
```diff
@@ -58,7 +58,8 @@ async def test_services_have_matching_labels(
                 assert value.startswith(
                     pod.metadata.labels[label.replace("k8s.element.io/target-", "app.kubernetes.io/")]
                 ), (
-                    f"{pod.metadata.name} does not have the correct label {label}={value} "
+                    f"{value} does not starts with "
+                    f"{pod.metadata.labels[label.replace('k8s.element.io/target-', 'app.kubernetes.io/')]}"
```
benbz:
```suggestion
                    f"Service/{service.metadata.name} has {label=}={value=} "
                    f"which does not start with the value from the Pod/{pod.metadata.name} "
                    f"{label.replace('k8s.element.io/target-', 'app.kubernetes.io/')}="
                    f"{pod.metadata.labels[label.replace('k8s.element.io/target-', 'app.kubernetes.io/')]} "
```

benbz (approved):

PR #613, opened 2025-07-10
Upgrade Synapse to v1.134.0

benbz:
https://github.com/element-hq/synapse/releases/tag/v1.134.0

### DIFF
--- a/newsfragments/613.changed.md
+++ b/newsfragments/613.changed.md
@@ -1,0 +1,7 @@
+Upgrade Synapse to v1.134.0.
+
+Highlights:
+* [MSC4267](https://github.com/matrix-org/matrix-spec-proposals/pull/4267) support - automatically forgetting rooms on leave
+
+Full Changelog:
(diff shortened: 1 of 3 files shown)


newsfragments/613.changed.md
```diff
@@ -0,0 +1,7 @@
+Upgrade Synapse to v1.134.0.
+
+Highlights:
+* [MSC4267](https://github.com/matrix-org/matrix-spec-proposals/pull/4267) support - automatically forgetting rooms on leave
```
jaywink, line 4: Should "Support for https://github.com/matrix-org/matrix-spec-proposals/pull/4235: via query param for hierarchy endpoint." be mentioned? :thinking: Not blocking on this, just for consideration, seems like a good feature to highlight.

benbz: I thought about it but didn't know how best to describe it and it isn't implemented in the Element clients AFAIK so left it

PR #617, opened 2025-07-15
document traefik timeouts

gaelgatelement:
See the conversation in the [ESS Community room](https://matrix.to/#/!RkXKfjpGXIVObpCqOI:element.io/$kvX8dpfZ4PSFEHArFrLaPMje7AZVwV22JaCLFbcOnWE?via=matrix.org&via=element.io&via=k8ekat.dev)

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -40,7 +40,36 @@ To configure your own PostgreSQL Database in your installation, copy the file `c
 
 K3s by default deploys the storage in `/var/lib/rancher/k3s/storage/`. If you want to change the path, you will have to run the K3s setup with the parameter `--default-local-storage-path <your path>`.
 
-# Monitoring
+## Configuring Traefik ingress timeouts when using K3s
+
+If you are experiencing timeouts when uploading large files to ESS, you will want to customize Traefik timeouts creating the file `traefik-config.yaml` in `/var/lib/rancher/k3s/server/manifests`. If the file already exists because you have configured custom ports for Traefik, add the example below to the existing file.
+
+```yml
+apiVersion: helm.cattle.io/v1
+kind: HelmChartConfig
+metadata:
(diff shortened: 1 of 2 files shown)


docs/advanced.md
```diff
@@ -40,7 +40,34 @@ To configure your own PostgreSQL Database in your installation, copy the file `c
 
 K3s by default deploys the storage in `/var/lib/rancher/k3s/storage/`. If you want to change the path, you will have to run the K3s setup with the parameter `--default-local-storage-path <your path>`.
 
-# Monitoring
+## Configuring Traefik timeouts
```
benbz:
```suggestion
## Configuring Traefik timeouts when using K3s
```

docs/advanced.md
```diff
@@ -40,7 +40,34 @@ To configure your own PostgreSQL Database in your installation, copy the file `c
...
+            readTimeout: "<timeout in seconds>s"
+            writeTimeout: "<timeout in seconds>s"
+            idleTimeout: "<timeout in seconds>s"
+      websecure:
+        transport:
+          respondingTimeouts:
+            readTimeout: "<timeout in seconds>s"
+            writeTimeout: "<timeout in seconds>s"
+            idleTimeout: "<timeout in seconds>s"
+```
```
benbz, line 68: Insert below (suggestions are being wierd)
> The above values correspond to the Traefik installation managed by K3s. If you are installing Traefik by other means, the exact structure of the configuration may differ.

If I've followed the discussion in the room correctly there was initially some confusion around the values used as the upstream chart was looked at and so this structure is only for Traefik in K3s yes?

gaelgatelement: Mmmh indeed. It's mostly the same chart, but still it's a fork that they sync regularly. The precision is probably helpful.

PR #621, opened 2025-07-16
Tweak CI values files

benbz:
* Introduce `charts/matrix-stack/ci/example-default-enabled-components-checkov-values.yaml` so that we've something that covers `deploymentMarkers` explicitly
* Remove `deploymentMarkers` from `{synapse,matrix-authentication-service}-minimal-values.yaml` as removing them doesn't require additional values to be set
  * At which point it also disappears from the corresponding `checkov` values file
* Remove `deploymentMarkers` from the Synapse based PyTests
  * The `syn2mas` test covers starting with legacy auth
  * The MAS PyTest does have `deploymentMarkers` available

### DIFF
--- a/charts/matrix-stack/ci/example-default-enabled-components-checkov-values.yaml
+++ b/charts/matrix-stack/ci/example-default-enabled-components-checkov-values.yaml
@@ -1,0 +1,64 @@
+# Copyright 2024-2025 New Vector Ltd
+#
+# SPDX-License-Identifier: AGPL-3.0-only
+#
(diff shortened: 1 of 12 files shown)


benbz:
No `dyff` changes to `matrix-authentication-service-checkov-values.yaml` and `matrix-authentication-service-minimal-values.yaml` as currently the deployment marker manifests are only deployed if the in-chart Synapse is deployed. This may need to be revisited

gaelgatelement (approved):

PR #627, opened 2025-07-25
Fix flaking tests

gaelgatelement:
It seems that `usefixtures()` does not wait for the fixtures to finish up. Let's add it as a fixture in the test function parameters instead.


benbz:
PR description is wrong however

PR #629, opened 2025-07-28
wait for all replicasets to be ready

gaelgatelement:
Trying to reduce CI flakes.

### DIFF
--- a/tests/integration/lib/helpers.py
+++ b/tests/integration/lib/helpers.py
@@ -15,6 +15,7 @@ from lightkube.models.core_v1 import (
     SecurityContext,
 )
 from lightkube.models.meta_v1 import ObjectMeta
+from lightkube.resources.apps_v1 import ReplicaSet
 from lightkube.resources.core_v1 import ConfigMap, Endpoints, Namespace, Pod, Secret
 
 from ..artifacts import CertKey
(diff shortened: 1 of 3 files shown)


tests/integration/lib/helpers.py
```diff
@@ -166,3 +167,19 @@ async def run_pod_with_args(kube_client: AsyncClient, namespace, image_name, pod
...
+    start_time = time.time()
+    now = time.time()
+    while start_time + timeout > now:
+        now = time.time()
+        await asyncio.sleep(0.5)
+        async for rs in kube_client.list(ReplicaSet, namespace=namespace):
+            if not rs.status or not rs.status.readyReplicas:
+                break
+            if rs.status.readyReplicas != rs.status.replicas:
+                break
```
benbz, line 183: Isn't this the wrong way around. It is going to exit the loop if no `readyReplicas` or the `readyReplicas != replicas` ?

gaelgatelement: If no `readyReplicas` or `readyReplicas != replicas`, this breaks the for loop and goes into one more `while` iteration. 
In practice, it mens if any `replicaSet` is not ready, it breaks and does one more `while` iteration.

If it does not break, it goes into the `else` clause and returns instantly ?

benbz: Ah right. I was focusing on the `while` loop and not really seeing the `async for` loop.

EMORECOFFEE

PR #630, opened 2025-07-28
skip pods part of a previous replicaset generation

gaelgatelement:
More flake fixes.


tests/integration/test_networking.py
```diff
@@ -42,6 +43,20 @@ async def test_services_have_matching_labels(
...
+                    ReplicaSet,
+                    namespace=generated_data.ess_namespace,
+                    labels={"pod-template-hash": op.equal(pod.metadata.labels["pod-template-hash"])},
+                ):
+                    # if any rs matches the pod's template hash,
+                    # the pod is one of the current replicaset and must be checked
+                    break
+                else:
+                    # Skip pods which do not have a ReplicaSet
+                    continue
```
benbz: I don't think this is the correct way to do this. There could be multiple `ReplicaSet` for a given `Deployment`. I think you want to look at whether the given `ReplicaSet` is desiring replicas too

PR #636, opened 2025-07-30
ci: fix showing warning about tags

gaelgatelement:
Noticed : 
image


benbz (approved):
Removes the warning, but if `tags: true` was doing nothing, do we really need `fetch-tags: true`?

Issue #615, opened 2025-07-14
Make experimental for mas configurable


suse-coder:
**Is your feature request related to a problem? Please describe.**
I would like to set these by myself. Currently they are hard coded into the helm chart.
```
[experimental](https://element-hq.github.io/matrix-authentication-service/reference/configuration.html#experimental)
Settings that may change or be removed in future versions. Some of which are in this section because they don't have a stable place in the configuration yet.

experimental:
  # Time-to-live of OAuth 2.0 access tokens in seconds. Defaults to 300, 5 minutes.
  #access_token_ttl: 300

  # Time-to-live of compatibility access tokens in seconds, when refresh tokens are supported. Defaults to 300, 5 minutes.
  #compat_token_ttl: 300

  # Experimental feature to automatically expire inactive sessions
  # Disabled by default
  #inactive_session_expiration:
     # Time after which an inactive session is automatically finished in seconds
     #ttl: 32400

     # Should compatibility sessions expire after inactivity. Defaults to true.
     #expire_compat_sessions: true

     # Should OAuth 2.0 sessions expire after inactivity. Defaults to true.
     #expire_oauth_sessions: true

     # Should user sessions expire after inactivity. Defaults to true.
     #expire_user_sessions: true
```

suse-coder:
This is what i have in the helm chart under additional for mas:
```
upstream_oauth2:
      providers:
        - id: "01H8Psdfsdf4YGH1RWV279"
          issuer: "https://test.de/realms/test-services"
          token_endpoint_auth_method: client_secret_basic
          client_id: "https://mas.test.de"
          client_secret: "liKtsdfsdddfNUFLcE10MrJLr4d"
          scope: "openid profile email profile"
          label: "Continue with test.de/realms/-services"
          button_text: "Continue with test.de/realms/-services"
          claims_imports:
            localpart:
              action: require
              template: "{{ user.preferred_username }}"
            displayname:
              action: suggest
              template: "{{ user.name }}"
            email:
              action: suggest
              template: "{{ user.email }}"
              set_email_verification: always
experimental:
      access_token_ttl: 300000
      compat_token_ttl: 300000
      inactive_session_expiration:
        ttl: 32400
        expire_compat_sessions: true
        expire_oauth_sessions: true
        expire_user_sessions: true
```

it is rendered into the secret but only the oauth2 works, the experimental seems to be not applied as it still has the old value. My guess is as it is hard coded it will override it from what I added.

[...]

PR #637, opened 2025-07-31
Update _helpers.tpl

suse-coder:
This should fix that 
https://github.com/element-hq/ess-helm/issues/615

### DIFF
--- a/charts/matrix-stack/templates/matrix-authentication-service/_helpers.tpl
+++ b/charts/matrix-stack/templates/matrix-authentication-service/_helpers.tpl
@@ -283,7 +283,7 @@ config.yaml: |
               "nameSuffix" "matrix-authentication-service"
               "containerName" (.containerName | default "render-config")
               "templatesVolume" (.templatesVolume | default "plain-config")
-              "overrides" (list "config.yaml")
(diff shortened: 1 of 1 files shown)


benbz (changes requested):
As mentioned in https://github.com/element-hq/ess-helm/issues/615#issuecomment-3070124164 we'd only be happy with `experimental` and `policy` being underrides, and so overridable with user config. This change makes everything overridable. `config.yaml` will need to be split into 2 files with appropriate settings in each.

If there's other settings you think should be overridable with user config, I'm happy to have a discussion

suse-coder:
What about removing it: access_token_ttl fully from that (as the other experimental settings are also not there). So everybody can choose his value.

benbz:
> What about removing it: access_token_ttl fully from that (as the other experimental settings are also not there). So everybody can choose his value.

We've discussed this internally and this is also an acceptable solution. On reflection we think the original trigger we had for raising the `access_token_ttl` has been resolved

benbz:
Closing as CLA was never signed and change in this PR was not appropriate

#745 will accomplish what this PR wanted

PR #639, opened 2025-08-01
Upgrade to Synapse v1.135.0

benbz:
Relies on the following PRs:
* https://github.com/element-hq/synapse/pull/18581
  * Some of the changes here show that the `receipts` event stream supports multiple workers, but it isn't documented as such. The `account_data` event stream doesn't and we currently bundle these 2 together. Enabling `receipts` as a scalable worker would require splitting the 2. Updating the documentation to make this obvious in https://github.com/element-hq/synapse/pull/18760 and implementing this in ESS in https://github.com/element-hq/ess-helm/pull/640
  * Incorporates the pending documentation changes in https://github.com/element-hq/synapse/pull/18761 given matrix.org's configuration
* https://github.com/element-hq/synapse/pull/18716
* https://github.com/element-hq/synapse/pull/18520
* <del>https://github.com/element-hq/matrix-authentication-service/pull/4801</del> we're explicitly using `synapse_legacy` for a few releases

### DIFF
--- a/charts/matrix-stack/configs/matrix-authentication-service/config.yaml.tpl
+++ b/charts/matrix-stack/configs/matrix-authentication-service/config.yaml.tpl
@@ -64,6 +64,9 @@ We don't want MAS to change data in Synapse
 */}}
 {{- if and .syn2mas.enabled .syn2mas.dryRun }}
   kind: synapse_read_only
+{{- else }}
+{{- /* Switch to synapse_modern after a release or 2 so that we're more likely to have a Synapse that supports this API on redeploying MAS. */}}
+  kind: synapse_legacy
 {{- end }}
 {{- end }}
 
--- a/charts/matrix-stack/ci/fragments/synapse-all-workers-running.yaml
+++ b/charts/matrix-stack/ci/fragments/synapse-all-workers-running.yaml
@@ -12,6 +12,8 @@ synapse:
       enabled: true
     client-reader:
       enabled: true
+    device-lists:
+      enabled: true
     encryption:
       enabled: true
     event-creator:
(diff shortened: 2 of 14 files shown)


charts/matrix-stack/configs/matrix-authentication-service/config.yaml.tpl
```diff
@@ -64,6 +64,8 @@ We don't want MAS to change data in Synapse
 */}}
 {{- if and .syn2mas.enabled .syn2mas.dryRun }}
   kind: synapse_read_only
+{{- else }}
+  kind: synapse_modern
```
sandhose: For rollout purposes, it's better to have MAS on the old API with this version, and we'll switch to the new API on the next version. This way, we don't end up in a situation where MAS tries to talk to Synapse using the new API, but Synapse is mid-rollout to 1.135. Unless of course if you have a way to first update Synapse then rollout the config change in MAS?

benbz: We don't have the mechanism to order Synapse redeploys before MAS redeploys.

We could leave this config change to a subsequent ESS Helm release however there's no guarantees people won't skip versions when upgrading and so theoretically this would always be a problem (though lesser and lesser as time goes by)

Are the MAS -> Synapse requests retried?

benbz: We're going to explicitly use `synapse_legacy` for a few releases especially given `Deployments` (MAS) are updated before `StatefulSets` (Synapse).

This doesn't provide any guarantee that people have upgraded to an intermediate release but it at least reduces the number of people who could get bit by this

sandhose: The plan is for us to MAS 1.0 soon™, and switch to the modern API by default at that point

PR #659, opened 2025-08-08
manifests tests: fix test does not create service account

gaelgatelement:
This simplifies the tests drastically by disabling all SA creation and checking that none exists.

We already verify `test_uses_serviceaccount_named_as_values_if_specified` and `test_uses_serviceaccount_named_as_per_pod_controller_by_default` so we dont need to check SA creation individually per component.

### DIFF
--- a/tests/manifests/test_serviceaccounts.py
+++ b/tests/manifests/test_serviceaccounts.py
@@ -104,6 +104,8 @@ async def test_does_not_create_serviceaccounts_if_configured_not_to(values, make
         serviceaccount_names = set()
         covered_serviceaccount_names = set()
         for template in await make_templates(values_to_modify):
+            if template_to_deployable_details(template) != deployable_details:
+                continue
             if template["kind"] in ["Deployment", "StatefulSet", "Job"]:
                 id_suffix = f" (for {deployable_details.name})"
                 workloads_by_id[f"{template_id(template)}{id_suffix}"] = template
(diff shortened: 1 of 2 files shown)


tests/manifests/test_serviceaccounts.py
```diff
@@ -104,6 +104,8 @@ async def test_does_not_create_serviceaccounts_if_configured_not_to(values, make
         serviceaccount_names = set()
         covered_serviceaccount_names = set()
         for template in await make_templates(values_to_modify):
+            if template_to_deployable_details(template) != deployable_details:
```
gaelgatelement: We are re-generating all templates when disabling the service account for only 1 deployable details. This caused issues such as the test detected a ServiceAccount not related to the currently tested deployable.

```
E                   AssertionError: Job/pytest-phwgyajqijitsjyhgvdeke-synapse-check-config (for synapse) specified an existing ServiceAccount: pytest-phwgyajqijitsjyhgvdeke-synapse-check-config
```

tests/manifests/test_serviceaccounts.py
```diff
@@ -123,6 +125,6 @@ async def test_does_not_create_serviceaccounts_if_configured_not_to(values, make
             else:
                 covered_serviceaccount_names.add(serviceaccount_name)
 
-        assert serviceaccount_names == covered_serviceaccount_names, (
+        assert serviceaccount_names.intersection(covered_serviceaccount_names) == set(), (
```
gaelgatelement: Because disabling service account creation does not disable setting the default service account name. This caused issues where the service account name was present but the service account properly absent, and the test failed.

```

E           AssertionError: Deployment/pytest-iwjypsiyvckpknjzhufqth-synapse-fed-reader (for synapse-federation-reader) created ServiceAccounts that it shouldn't have
E           assert set() == {'pytest-iwjy...fqth-synapse'}
E             
E             Extra items in the right set:
E             'pytest-iwjypsiyvckpknjzhufqth-synapse'
E             
E             Full diff:
E             + set()
E             - {
E             -     'pytest-iwjypsiyvckpknjzhufqth-synapse',
E             - }

```

benbz (approved):

PR #660, opened 2025-08-08
Set ipFamilyPolicy on RTC SFU NodePorts

rarbab:
If the cluster is IPv6 [enabled](https://docs.k3s.io/networking/basic-network-options#dual-stack-ipv4--ipv6-networking), NodePorts created for RTC SFU will still default to SingleStack. Set ipFamilyPolicy on these ports to PreferDualStack instead so we can connect with both IPv4 and v6.

This should cause no change in behavior on an IPv4-only install.

### DIFF
--- a/charts/matrix-stack/templates/matrix-rtc/sfu_rtc_tcp_service.yaml
+++ b/charts/matrix-stack/templates/matrix-rtc/sfu_rtc_tcp_service.yaml
@@ -18,6 +18,7 @@ metadata:
 spec:
   type: NodePort
   externalTrafficPolicy: Local
+  ipFamilyPolicy: PreferDualStack
   ports:
   - name: "rtc-tcp"
     protocol: "TCP"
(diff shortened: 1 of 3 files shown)


benbz:
Thanks for this contribution. Some notes / thoughts
* We're attempting to fix the integration test failures for you in #661 
* Should changes happen to other `Services` as well or is it only important for the Matrix RTC `Services`?
* Is there value in this being configurable rather than hard-coded to `PreferDualStack`?

gaelgatelement:
Sorry for closing the PR, linking the PRs made github think it should close it.

rarbab:
> * Should changes happen to other `Services` as well or is it only important for the Matrix RTC `Services`?

I think it only matters for ports that get exposed outside the cluster. Other than the RTC ports I'm enabling in this PR, this chart only exposes http/https via traefik LoadBalancer ingress, and those seem to listen to IPv6 automatically when I enabled it.

> * Is there value in this being configurable rather than hard-coded to `PreferDualStack`?

I don't think so. There are only a few options for this resource:
* `SingleStack`: "The Service uses either IPv4 or IPv6, but not both."
* `PreferDualStack`: "The Service will attempt to use both IPv4 and IPv6, but will fall back to a single stack if dual-stack is not available."
* `RequireDualStack`: "The Service requires both IPv4 and IPv6 addresses and will not be created if dual-stack is not available."

I don't see why someone would pick one of others.

PR #669, opened 2025-08-12
add dockerhub secret to curl pods

gaelgatelement:
To fix things such as https://github.com/element-hq/ess-helm/actions/runs/16901759366/job/47882226641?pr=667

### DIFF
--- a/tests/integration/lib/helpers.py
+++ b/tests/integration/lib/helpers.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: AGPL-3.0-only
 
 import asyncio
+import os
 import time
 
 import pyhelm3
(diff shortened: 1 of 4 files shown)


tests/integration/lib/helpers.py
```diff
@@ -109,11 +110,14 @@ async def get_deployment_marker(kube_client, generated_data, marker: str):
...
 
-async def run_pod_with_args(kube_client: AsyncClient, namespace, image_name, pod_name, args):
+async def run_pod_with_args(kube_client: AsyncClient, generated_data, image_name, pod_name, args):
     pod = Pod(
-        metadata=ObjectMeta(name=pod_name + "-" + str(int(time.time() * 1000)), namespace=namespace),
+        metadata=ObjectMeta(name=pod_name + "-" + str(int(time.time() * 1000)), namespace=generated_data.ess_namespace),
         spec=PodSpec(
             restartPolicy="Never",
+            imagePullSecrets=[
+                LocalObjectReference(name=f"{generated_data.release_name}-dockerhub"),
```
benbz: This is fine if we don't have the DH pull secret isn't it? i.e. on a fork

benbz: You gone and gated the addition of the pull secret above but not here?

PR #674, opened 2025-08-14
MAS: Update to v1.0.0

jaywink:
https://github.com/element-hq/matrix-authentication-service/releases/tag/v1.0.0

### DIFF
--- a/newsfragments/changed.674
+++ b/newsfragments/changed.674
@@ -1,0 +1,9 @@
+Update Matrix Authentication Service to v1.0.0.
+
+Highlights:
+
+* Support for stable Matrix native OIDC scopes.
+* Add `clients.[].client_secret_file` config option.
+* Add `matrix.secret_file` config option.
+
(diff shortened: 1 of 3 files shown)


newsfragments/changed.674
```diff
@@ -0,0 +1,9 @@
+Update Matrix Authentication Service to v1.0.0.
+
+Highlights:
+
+* Support for stable Matrix native OIDC scopes.
+* Add `clients.[].client_secret_file` config option.
+* Add `matrix.secret_file` config option.
```
benbz, line 7: We should update the config to use both of these rather than `clients.[].client_secret` / `matrix.secret`.

Can happen in a different PR though

newsfragments/changed.674
benbz, line 1: Should be `674.changed.md`

benbz:
FYI I'm proposing replacing this with #689

PR #692, opened 2025-08-22
Support configuring a different cluster domain for internal Service references.

### LINKED ISSUE #691: Support configurable Kubernetes cluster domain
Yes. When running the Element Server Suite (ESS) on a Kubernetes cluster whose domain is not the default `cluster.local`, internal endpoints generated by ESS components (HAProxy, matrix-authentication-service, etc.) are incorrect and cause connection failures. All generated URLs still use the hard-coded domain `cluster.local`.
Example:
- Expected endpoint: `http://ess-synapse-main.ess.svc.sub.example.com:8008/`
-  Actual (broken) endpoint: `http://ess-synapse-main.ess.svc.cluster.local:8008/`

Introduce a configurable cluster-domain value in the ESS Helm charts:
1. Add a new global parameter `.Values.global.clusterDomain` (default: `cluster.local`).
2. Replace every hard-coded cluster.local reference in templates with
`{{ .Values.global.clusterDomain }}`.
3. Ensure that all affected components (Synapse, MAS, HAProxy, well-known endpoints, etc.) use this value consistently.

- Post-install manual editing of ConfigMaps and Secrets (error-prone and overwritten on upgrades).
- DNS-level CNAME aliases (does not work for components that rely on TLS or SNI).
- Maintaining a private fork of the charts (high maintenance cost).

**Additional context**
The change allow seamless deployment on clusters with custom domains, improving portability and reducing operational friction.

I’m happy to accept any alternative suggestions or to be corrected if I’ve missed a simpler approach—please let me know if there’s a better way to handle this!

benbz:
Fixes #691 and also should reduce the DNS queries for these names 5-fold as the search domains are no longer considered

### DIFF
--- a/charts/matrix-stack/configs/matrix-authentication-service/config.yaml.tpl
+++ b/charts/matrix-stack/configs/matrix-authentication-service/config.yaml.tpl
@@ -41,7 +41,7 @@ database:
   uri: "postgresql://{{ .user }}:${POSTGRES_PASSWORD}@{{ tpl .host $root }}:{{ .port }}/{{ .database }}?{{ with .sslMode }}sslmode={{ . }}&{{ end }}application_name=matrix-authentication-service"
 {{- end }}
 {{- else if $root.Values.postgres.enabled }}
-  uri: "postgresql://matrixauthenticationservice_user:${POSTGRES_PASSWORD}@{{ $root.Release.Name }}-postgres.{{ $root.Release.Namespace }}.svc.cluster.local:5432/matrixauthenticationservice?sslmode=prefer&application_name=matrix-authentication-service"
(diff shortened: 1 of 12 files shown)


AndrewFerr (approved):
Looks good. Wasn't aware of trailing dots being beneficial (let alone allowed) for DNS lookups!

Maybe this test could be updated to use the config:
https://github.com/element-hq/ess-helm/blob/b27942de02346be7e6af6a18ddf50bc8e98a2d99/tests/integration/test_networking.py#L247
Leaving it to you to decide whether that's appropriate.

benbz:
> Looks good. Wasn't aware of trailing dots being beneficial (let alone allowed) for DNS lookups!
> 
> Maybe this test could be updated to use the config:
> 
> https://github.com/element-hq/ess-helm/blob/b27942de02346be7e6af6a18ddf50bc8e98a2d99/tests/integration/test_networking.py#L247
> 
> Leaving it to you to decide whether that's appropriate.

It doesn't hurt but doesn't really help either as it is a one-shot thing run by the tests only. I'm going to leave it as is for clarity

PR #705, opened 2025-08-27
tests: check labels values against validation regex

gaelgatelement:
It's handy to catch this before integration tests.

### DIFF
--- a/tests/manifests/test_labels.py
+++ b/tests/manifests/test_labels.py
@@ -147,3 +147,12 @@ async def test_workloads_selector_matches_labels(templates):
                     template["spec"]["template"]["metadata"]["labels"][label]
                     == template["spec"]["selector"]["matchLabels"][label]
                 )
+
+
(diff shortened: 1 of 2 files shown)


tests/manifests/test_labels.py
```diff
@@ -147,3 +147,12 @@ async def test_workloads_selector_matches_labels(templates):
                     template["spec"]["template"]["metadata"]["labels"][label]
                     == template["spec"]["selector"]["matchLabels"][label]
                 )
+
+
+@pytest.mark.parametrize("values_file", values_files_to_test)
+@pytest.mark.asyncio_cooperative
+async def test_labels_values_are_valid(templates):
+    label_regex = re.compile(r"^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?$")
```
benbz, line 157: Can we have a comment linking to the source of this regex

PR #718, opened 2025-09-08
Update MAS to v1.2.0

benbz:
Draft as RC. Changelog written as per release

### DIFF
--- a/newsfragments/718.changed.md
+++ b/newsfragments/718.changed.md
@@ -1,0 +1,7 @@
+Update Matrix Authentication Service to v1.2.0.
+
+Highlights:
+* Translation updates
+
+Full Changelogs:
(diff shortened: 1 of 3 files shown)


newsfragments/718.changed.md
```diff
@@ -0,0 +1,7 @@
+Update Matrix Authentication Service to v1.2.0.
+
+Highlights:
```
jaywink, line 3: Not sure if the only feature "Automatically derive the kid from the key fingerprint if missing" would be worth mentioning instead of translation updates, it feels more significant, without fully knowing how significant? :grin: 

Not a blocker anyway.

benbz: We explicitly specify the `kid` for all keys we provide and we don't want to unset that given the `kid` must be stable across restarts. I'm unclear whether there can be more than 1 key of a given type so I didn't want to mention this at all

PR #724, opened 2025-09-10
.helmignore: add vim swapfiles

AndrewFerr:


benbz (commented):
I can't see why we'd ever want these committed but equally I can't see why anyone would want these committed to any repo, so `~/.gitignore` (`git config --global core.excludesFile '~/.gitignore'`) would be better than per-repo?

AndrewFerr:
> `~/.gitignore` (`git config --global core.excludesFile '~/.gitignore'`) would be better than per-repo?

It absolutely would be, but unfortunately Helm doesn't support global `.helmignore` files.

PR #728, opened 2025-09-12
tests: add Synapse API scope to MAS admin users

AndrewFerr:
When creating a MAS user for tests, if the user is to be a Synapse admin, ensure that its OAuth 2.0 token is granted the scope for making requests to the Synapse Admin API.


tests/integration/lib/matrix_authentication_service.py
```diff
@@ -101,6 +101,8 @@ async def create_mas_user(
     scopes = [
         "urn:matrix:org.matrix.msc2967.client:api:*",
     ]
+    if admin:
+        scopes.append("urn:synapse:admin:*")
```
benbz, line 105: Is the MAS admin scope also required?

AndrewFerr: None of the current tests need it.

AndrewFerr:
The licensing check is failing on the changelog file.  Is there a way to make the test skip over the `newsfragments` directory altogether?

benbz:
> The licensing check is failing on the changelog file. Is there a way to make the test skip over the `newsfragments` directory altogether?

Your fragment is missing the `.md` extension, which is then covered with https://github.com/element-hq/ess-helm/blob/main/REUSE.toml#L23-L26

PR #729, opened 2025-09-12
use apply instead of create and avoid silencing errors

gaelgatelement:
While testing some complex setups, I discovered that how we do `helm_prerequisites` today might hide errors from the pytest results. Typically, it errors out on upgrade because we try to recreate resources already existing.

### DIFF
--- a/tests/integration/fixtures/cluster.py
+++ b/tests/integration/fixtures/cluster.py
@@ -4,9 +4,7 @@
 
 import asyncio
 import os
-from collections.abc import AsyncGenerator
 from pathlib import Path
-from typing import Any
 
 import pyhelm3
 import pytest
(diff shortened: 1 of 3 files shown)


tests/integration/fixtures/cluster.py
benbz, line 1: I'm not sure I understand why this changes in this file are needed. AFAICT this just means that it doesn't yield the `Namespace` object but otherwise has the same behaviour

gaelgatelement: Indeed, but it turns out if we did not hide the error, it was failing with `Namespace type is not hashable`

gaelgatelement: (Something in asyncio_cooperative was trying to pass the args to the coroutines or something. I did not investigate much as we are actually not using it anywhere)

PR #733, opened 2025-09-15
Matrix RTC networking

gaelgatelement:


charts/matrix-stack/configs/matrix-rtc/sfu/config-overrides.yaml.tpl
```diff
@@ -23,6 +23,16 @@ logging:
 
 # WebRTC configuration
 rtc:
+  use_external_ip: {{ .useStunToDiscoverPublicIP }}
```
benbz: I think this should go into `config-underrides.yaml.tpl` for at least 1 monthly release cycle. This would normal be the correct place for it but given people may be configuring this manually for now, I think we should respect their manual configurations temporarily to give them time to upgrade

charts/matrix-stack/source/matrix-rtc.yaml.j2
```diff
@@ -18,7 +18,7 @@ restrictRoomCreationToLocalUsers: false
 # livekitAuth:
   {{- sub_schema_values.credential("The keys.yaml file for the LiveKit SFU\n## This is required if `sfu.enabled` is set to `false`", "keysYaml", initIfAbsent=False, commented=True) | indent(2) }}
   ## Provide a key and secret if not using an existing keys.yaml
-  # key: ""
+  # key: "matrix-rtc"
```
benbz: This either needs to be uncommented or the `| default "matrix-rtc"` added back to both envs

gaelgatelement: Yup I rolled back on this change.

zackman0010:
Hi @gaelgatelement! I had the bad luck of trying to set up my installation shortly after you had merged this in, so I was getting the following error while following the documentation for configuring mRTC:
```
matrix-stack:
- matrixRTC.sfu: Additional property useStunToDiscoverPublicIP is not allowed
- matrixRTC.sfu: Additional property manualIP is not allowed
```

Do you know when this merge will be released?

PR #737, opened 2025-09-16
tests: restart curl pods on failure to fetch metrics

gaelgatelement:


benbz (approved):
This will keep retrying until the 60s are up yes?

gaelgatelement:
> This will keep retrying until the 60s are up yes?

I expect so yes

PR #738, opened 2025-09-16
matrix-tools: set container args, not command

AndrewFerr:
The matrix-tools image defines an entrypoint, so define containers with "args" instead of "command", as the latter overrides the entrypoint.


AndrewFerr:
The purpose of this PR is to make it easier to swap out `matrix-tools` with a debug image, which would have a different `ENTRYPOINT`.

It's also good practice to make use of `ENTRYPOINT` in general.

PR #740, opened 2025-09-17
syn2mas: Migrations steps when Synapse was already using OAuth

kenodai:
Synapse Config Hooks throws an error when the oidc configuration is still included.
syn2mas Job throws an error when oidc configuration is NOT part of the synapse config.

Solution is to disable the config checker hook.

Thanks to @gaelgatelement for the help.


gaelgatelement (approved):
LGTM. Thanks for the PR!

PR #747, opened 2025-09-22
CI: move synapse.appservices into its own fragment

AndrewFerr:
This allows aggregate values files to include external secrets for Synapse without having to include synapse.appservices.


AndrewFerr:
Closing because this is no longer needed.

PR #748, opened 2025-09-22
CI: disable initSecrets for *-secrets-*.yaml

AndrewFerr:
Disable initSecrets for test values files that are to use secrets defined in Helm or external secrets.


AndrewFerr:
> Did we also want to remove `init-secrets-minimal.yaml` given `initSecrets` is default enabled and has no properties?

I'd rather that go in a different PR; plenty of other files still use it and I'd like for this PR to stick to changing just the `*secrets-*` files.

PR #752, opened 2025-09-23
Update Synapse to 1.139.0

benbz:
Draft as RC. Changelog written as per release.


benbz:
Changelog fragment will either remain the same or will edit #755's fragment depending if an ESS release has happened or not

PR #754, opened 2025-09-24
add go cache while building matrix-tools

gaelgatelement:


matrix-tools/Dockerfile
```diff
@@ -7,10 +7,12 @@ FROM --platform=${BUILDPLATFORM} golang:1.24 AS buildstage
 WORKDIR /app
 COPY . /app
 
-RUN go mod download
+RUN --mount=type=cache,target=/go/pkg/mod \
+  go mod download
 
 ARG TARGETOS TARGETARCH
-RUN GOOS=$TARGETOS GOARCH=$TARGETARCH CGO_ENABLED=0 go build -o /app/matrix-tools cmd/main.go
+RUN --mount=type=cache,target=/go/pkg/mod \
```
benbz: Looking at https://www.reddit.com/r/golang/comments/1f4vvir/caching_go_mod_download_in_docker/ / https://medium.com/@marcin.niemira/optimise-docker-build-for-go-c03d6eb8b4b there appears to be another cache, not just the module cache. Unsure what the default is without `go env -w GOCACHE=<path>`

gaelgatelement: MMmh I followed the example from https://docs.docker.com/guides/bake/#testing-and-linting. 

We might be better with explicit configuration. I'll follow the suggestion from these posts.

benbz (approved):
Approved as this is no worse in CI than before and may help locally. As discussed may need https://github.com/reproducible-containers/buildkit-cache-dance to gain the full benefits of this in CI

gaelgatelement:
Merging this for now, this can be revisited later to improve CI as well.

PR #743, opened 2025-09-17
Add admin console

gaelgatelement:
WIP

PR #762, opened 2025-09-30
Upgrade Element Admin to 0.1.3

benbz:
As per https://github.com/element-hq/element-admin/compare/v0.1.2...v0.1.3.

Changelog is a c/p of #743 so that both PRs are linked


benbz:
FTR the security header changes for Element Admin mean it can't be embedded in an iframe but does mean it matches upstream. I feel that's acceptable at this time

gaelgatelement (approved):

PR #768, opened 2025-10-02
Initial troubleshooting guide for `MISSING_MATRIX_RTC_FOCUS`

benbz:
'Error Code: `MISSING_MATRIX_RTC_FOCUS`' has come up a few times. Generally there have been problems with well-knowns with it. Let's document that

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,8 @@ A full comparison between the editions can be found [here](https://element.io/pr
     - [Creating an initial user](#creating-an-initial-user)
     - [Verifying the setup](#verifying-the-setup)
 - [Advanced setup](#advanced-setup)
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -530,6 +532,10 @@ To verify the setup, you can:
```
gaelgatelement, line 532: Should we insert in `Verifying the setup` that federation checks must succeed for Element Call features to work ?

benbz: The federation tester is only going to work if you have DNS entries that are publicly accessible. Whereas all the tests I've run here are local ones

gaelgatelement: It might still be worth mentioning that it must work for public access maybe.

benbz: Done

Issue #770, opened 2025-10-06
Synapse pods fail on restart due to matrix-tools render-config permissions issue


wmTJc9IK0Q:
**Describe the bug**
When the synapse pod is restarted due to failing a liveness probe or hitting the memory limit, the matrix-tools init container for rendering config crashes due to not having permissions to write to the already rendered config file:

```
Error writing to file: open /conf/homeserver.yaml: permission denied 
```

This is because the file is written in 0440 mode (read only to group and owner) on the first pod startup: https://github.com/element-hq/ess-helm/blob/8bef958b4bfefc456fb4cc15f719c806d2d64b7f/matrix-tools/internal/cmd/render-config/cmd.go#L65

This file survives pod restarts due to being in an emptyDir volume, but matrix-tools does not handle this scenario.

**To Reproduce**
1. Install matrix-ess chart
2. Somehow get a pod to restart, such as failing a livenessProbe (not sure of a reliable way to do this)

**Expected behavior**
The synapse pod needs to be able to restart with failing.

Probably best option is to only render the config if the file doesn't already exist.

Another option would be to make the file writeable to the user for the matrix-tool container, but it would need to be changed to be a different user than the one that runs the synapse container.

## Workaround

I am in need of a workaround but I can't find one without changing the chart.
- I can't remove the liveness probes from values.yml
- I can't change the command used for the matrix-tools init container to check for the file
- I can't modify the user used by the matrix-tool container (via `containersSecurityContext`) without changing it for all containers I think??

wmTJc9IK0Q:
I was able to workaround by building an image from https://github.com/element-hq/ess-helm/pull/771 which skips re-creating the file if it exists and overriding `matrixTools.image`

benbz:
It would be interesting to know what caused the first failure. You can raise the thresholds and frequencies for any of the probes so that's a potential approach.

The security context of the various `matrix-tools` containers is the same as the security context of the component they're being a helper for. So changing Synapse's would change the `render-config` container's security context for Synapse only. There isn't a `matrix-tools` security context set of settings.

Will take a look at #771 thanks

wmTJc9IK0Q:
I found that it's actually happening due to a sandbox change from containerd: `Pod sandbox changed, it will be killed and re-created.`

I don't yet know what's triggering that, but my guess it's something to do with my CNI. Regardless, matrix-tools needs to handle this scenario with the default config.

benbz:
Fixed by #771 which will be in the next release

PR #771, opened 2025-10-06
Skip render-config when output exists

wmTJc9IK0Q:
See https://github.com/element-hq/ess-helm/issues/770


wmTJc9IK0Q:
I considered that but I think there is an edge case there that could cause a config to be applied that maybe shouldn't be:
- helm upgrade starts, configmap and secrets are updated
- something fails during the upgrade, before the `k8s.element.io/synapse-config-hash` and `k8s.element.io/synapse-secret-hash` labels on the statefulset are updated which would cause a rollout of a new pod.
- if the synapse pod is restarted in this scenario, it would get an updated config which could be wrong vs. continuing to work with the previous config

I think it's slightly safer to assume immutability of the config and only have the pod change it's config with the hash labels update.

benbz:
> I considered that but I think there is an edge case there that could cause a config to be applied that maybe shouldn't be:
> 
>     * helm upgrade starts, configmap and secrets are updated
> 
>     * something fails during the upgrade, before the `k8s.element.io/synapse-config-hash` and `k8s.element.io/synapse-secret-hash` labels on the statefulset are updated which would cause a rollout of a new pod.
> 
>     * if the synapse pod is restarted in this scenario, it would get an updated config which could be wrong vs. continuing to work with the previous config
> 
> 
> I think it's slightly safer to assume immutability of the config and only have the pod change it's config with the hash labels update.

Because of the check config hook the (chart provided) Synapse config won't be updated until that job has passed and the config is known valid. I think the likelihood of the the `helm upgrade` being interrupted between applying the updated `ConfigMaps`/`Secrets` and applying the updated `StatefulSet` is too small to worry about. For other components, yes we don't have check config hooks to validate their config but an admin should notice a failing `helm upgrade` command.

There is another case were the `Pod` could be restarted and get different config from when `helm install`/`helm upgrade` was last run: external Secrets change then `Pod` gets evicted from a `Node`. In that scenario the only thing I can see us being able to do to prevent that is that the `Pod` is given permissions to create/update `Secrets` in cluster to save off the config it started with as part of the `render-config` init-container. I dislike that as I don't think the applications should have permissions like that in general.

Given that we're left with at least one situation where the `Pods` could restart with different config to what they were originally started with, I would lean towards us being consistent and saying that config is always regenerated on startup regardless of whether it is a `Pod` being freshly (re)created or restarting.

wmTJc9IK0Q:
@benbz I have updated the PR to delete. I am testing this on my own cluster now.

gaelgatelement:
Thanks for the PR! This will be part of next release through https://github.com/element-hq/ess-helm/pull/782/files

PR #782, opened 2025-10-08
matrix-tools 0.5.6

gaelgatelement:


newsfragments/782.fixed.md
benbz, line 1: Can this be copied to `771.fixed.md` as well please

benbz (approved):

PR #792, opened 2025-10-09
Add a gofmt lint job

benbz:
Draft as I want to see this job fail before I actually run `gofmt` on the code


benbz:
https://github.com/element-hq/ess-helm/actions/runs/18378159301/job/52357300781 was a failure prior to the 2nd commit

PR #794, opened 2025-10-10
update poetry to 2.x in CI

gaelgatelement:


benbz:
I think CI is using 2.1.1 based on `matrix-org/setup-python-poetry` but it is difficult to tell as there's no explicit `poetry --version` called anywhere and caches are used.

Should we set a lower bound in our `pyproject.toml`?

PR #787, opened 2025-10-08
Upgrade to MAS v1.4.0

benbz:
Draft as RC. Changelog written as per release

PR #805, opened 2025-10-13
Avoid showing hash or version label changes in `dyff`

benbz:
e.g. https://github.com/element-hq/ess-helm/pull/787 has a difficult to follow `dyff` because of all the label changes. Let's ignore them as we can't review hashes manually and have an automated test to check that hashes do change. We've already got tests (`test_pods_with_tags_and_no_digests` & `test_pods_with_digests_and_tags`) that validate `app.kubernetes.io/version` matches the tag so also ignore it

### DIFF
--- a/.github/workflows/templates-dyff.yml
+++ b/.github/workflows/templates-dyff.yml
@@ -139,8 +139,18 @@ jobs:
               continue
             fi
 
+            excludes=()
+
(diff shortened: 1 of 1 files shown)


benbz:
It looks like `dyff between` will always show new hash labels regardless of `--exclude-regexp`. I think this is a good thing

PR #806, opened 2025-10-13
ci: create a dedicated serverName fragments

gaelgatelement:
We want to avoid defining servername in multiple fragments as that could cause users to get confused and have multiple `serverName` defined in their values files.

### DIFF
--- a/charts/matrix-stack/ci/example-default-enabled-components-checkov-values.yaml
+++ b/charts/matrix-stack/ci/example-default-enabled-components-checkov-values.yaml
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: AGPL-3.0-only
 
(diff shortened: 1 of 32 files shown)


charts/matrix-stack/user_values/.gitkeep
benbz, line 1: Why's this disappearing?

gaelgatelement:
> serverName: "{{ $.Values.global.baseDomain }}"

It is only used in well-known pytest run. What would be the rationale in moving it to its own fragment? Making it obvious that one can do so?

benbz:
> > serverName: "{{ $.Values.global.baseDomain }}"
> 
> It is only used in well-known pytest run. What would be the rationale in moving it to its own fragment? Making it obvious that one can do so?

It is probably fine in there for now, was just asking the question. However, as soon as we want another dynamic `serverName` values file (e.g. `example-default-enabled-components-dynamic-values.yaml` to better exercise the test added in https://github.com/element-hq/ess-helm/pull/791) then it should come out to its own fragment or we're reopening the same issue

PR #817, opened 2025-10-20
Config/Secrets/Paths consistency tests refactoring

gaelgatelement:
This is a re-writing of https://github.com/element-hq/ess-helm/blob/main/tests/manifests/test_configs_and_mounts_consistency.py which were starting to be very difficult to maintain and understand.

All tests should be here but 

```py
            assert len(mount_paths) == len(set(mount_paths)), (
                f"Secrets mount paths are not unique in {template['metadata']['name']}: {mounted_keys}"
            )
```

As I'm not sure if they make sense. When implementing it with the new structure, it raised against the SFU which is mounting the `/conf` twice, but using `subPath`.

### DIFF
--- a/tests/manifests/test_configs_consistency.py
+++ b/tests/manifests/test_configs_consistency.py
@@ -395,22 +395,11 @@ class GenericContainerSpecPathConsumer(PathConsumer):
         )
 
     def path_is_used_in_content(self, path) -> bool:
-        return find_path_in_content(
-            path,
-            list(self.env.values())
-            + self.args
-            + self._empty_dir_rendered_content()
-            + list(self.exec_properties.values()),
-        )
+        return find_path_in_content(path, self._all_container_content())
 
     def get_all_paths_in_content(self, deployable_details: DeployableDetails):
         paths = []
-        for content in (
-            list(self.env.values())
-            + list(self.exec_properties.values())
-            + self.args
-            + self._empty_dir_rendered_content()
-        ):
+        for content in self._all_container_content():
             paths += match_path_in_content(content)
         return paths
 


tests/manifests/test_configs_and_mounts_consistency.py
benbz, line 1: Can we add a comment here that this should disappear after a period of time

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,524 @@
...
+    get_volume_from_mount,
+    match_path_in_content,
+)
+from .utils import template_id, template_to_deployable_details
+
+
+## Gets all mounted files in a given container from the ConfigMaps referenced by the volume mounts
+def get_all_mounted_files(workload_spec, container_name, templates):
+    found_files = {}
+    for container_spec in workload_spec["containers"] + workload_spec.get("initContainers", {}):
```
benbz:
```suggestion
    for workload_spec.get("initContainers", []) + container_spec in workload_spec["containers"]:
```

As `initContainers` are a list and come first. This function should probably also short-circuit once it has processed the desired container

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,524 @@
...
+from . import DeployableDetails, secret_values_files_to_test, values_files_to_test
+from .test_configs_and_mounts_consistency import (
+    assert_exists_according_to_hook_weight,
+    find_keys_mounts_in_content,
+    get_configmap,
+    get_or_empty,
+    get_secret,
+    get_volume_from_mount,
+    match_path_in_content,
+)
```
benbz: If this test is to be eventual only one, these functions should move into here and be imported from the existing test

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,524 @@
...
+# A mounted configmap will be the source of a mounted path for each configmap key
+@dataclass(frozen=True)
+class MountedConfigMap(SourceOfMountedPaths):
+    data: dict[str, str] = field(default_factory=dict)
+    mount_point: str = field(default_factory=str)
+
+    @classmethod
+    def from_template(cls, template, volume_mount):
+        assert template["kind"] == "ConfigMap"
+        # When secret data is empty, `data:` is None, so use `get_or_empty`
```
benbz:
```suggestion
        # When ConfigMap data is empty, `data:` is None, so use `get_or_empty`
```

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,524 @@
...
+    render_config_outputs: set[str] = field(default_factory=set)
+    mount_point: str = field(default_factory=str)
+
+    @classmethod
+    def from_workload_spec(cls, workload_spec, before_index, volume_mount):
+        outputs = []
+        # We look for render-config containers which ran before the current container
+        for container_spec in workload_spec["initContainers"][:before_index]:
+            if "render-config" in container_spec["name"]:
+                args = container_spec.get("args") or container_spec["command"][1:]
```
benbz: I'd prefer to look at the container image and args rather than the name. We know `command` isn't going to be used for `matrix-tools` due to https://github.com/element-hq/ess-helm/pull/820

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,524 @@
...
+
+    @classmethod
+    def from_workload_spec(cls, workload_spec, before_index, templates):
+        potential_input_files = {}
+        inputs_files = {}
+        # We look for render-config containers which ran before the current container
+        for container_spec in workload_spec["initContainers"][:before_index]:
+            if container_spec["name"].startswith("render-config"):
+                potential_input_files = get_all_mounted_files(workload_spec, container_spec["name"], templates)
+                args = container_spec.get("args") or container_spec["command"][1:]
```
benbz: Detect by `image` and `args` not `name`. No need to consult `command`

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,524 @@
...
+                source_files = args[3:]
+                for p, k in potential_input_files.items():
+                    if p in source_files:
+                        inputs_files[p] = k
```
benbz: If `p` isn't in `source_files` is that a problem?

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,524 @@
...
+@pytest.mark.parametrize("values_file", values_files_to_test | secret_values_files_to_test)
+@pytest.mark.asyncio_cooperative
+async def test_secrets_consistency(templates, other_secrets):
+    workloads = [t for t in templates if t["kind"] in ("Deployment", "StatefulSet", "Job")]
+    for template in workloads:
+        deployable_details = template_to_deployable_details(template)
+        # Gather all containers and initContainers from the template spec
+        containers = template["spec"]["template"]["spec"].get("containers", []) + template["spec"]["template"][
+            "spec"
+        ].get("initContainers", [])
```
benbz: Let's extract out a `workload_spec = template["spec"]["template"]["spec"]` and do `initContainers` before `containers` as per their logical ordering

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,524 @@
...
+                templates,
+                other_secrets,
+            )
+            validated_container_config.check_paths_used_in_content(deployable_details.paths_consistency_noqa)
+            validated_container_config.check_paths_lookalikes_matchs_source_of_mounted_paths(
+                deployable_details.paths_consistency_noqa
+            )
+            validated_container_config.check_all_paths_matches_an_actual_mount(
+                deployable_details.skip_path_consistency_for_files, deployable_details.paths_consistency_noqa
+            )
```
benbz: This smells like we should have 3 separate tests and the actual body of the tests should then come out into those 3 tests

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,524 @@
...
+            # Determine which secrets are mounted by this container
+            mounted_files = []
+            has_rendered_config = False
+
+            for volume_mount in container_spec.get("volumeMounts", []):
+                current_volume = get_volume_from_mount(workload_spec, volume_mount)
+                if "secret" in current_volume:
+                    # Extract the paths where this volume's secrets are mounted
+                    secret = get_secret(templates, other_secrets, current_volume["secret"]["secretName"])
+                    assert_exists_according_to_hook_weight(secret, weight, name)
```
benbz: I really wanted to pull this checking out into `test_volumes_mounts_exists` in `tests/manifests/test_volume_mounts.py`. i.e. this test should be able to assume that all its volumeMounts are corretly present

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,524 @@
...
+                    validated_config.sources_of_mounted_paths.append(MountedSecret.from_template(secret, volume_mount))
+                elif "configMap" in current_volume:
+                    # Parse config map content
+                    configmap = get_configmap(templates, current_volume["configMap"]["name"])
+                    assert_exists_according_to_hook_weight(configmap, weight, name)
+                    # We do not consume ConfigMaps in render-config, their configuration
+                    # will actually be consumed later by the container using the rendered-config
+                    if not container_spec["name"].startswith("render-config"):
```
benbz: Detect by image and args not name

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,524 @@
...
+                        validated_config.sources_of_mounted_paths.append(
+                            MountedConfigMap.from_template(configmap, volume_mount)
+                        )
+                        validated_config.paths_consumers.append(ConfigMapPathConsumer.from_configmap(configmap))
```
benbz: Why do we need `ConfigMapPathConsumer` but not a `SecretPathConsumer`?

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,524 @@
...
+                elif "emptyDir" in current_volume:
+                    if current_volume["name"] == "rendered-config" and not container_spec["name"].startswith(
+                        "render-config"
+                    ):
```
benbz: Hmm, ideally I'd like not to have to rely on specific volume names. I'd like that we detect a prior render config `initContainer` and know what paths it has populated in a given `emptyDir`

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,524 @@
...
+
+            mounted_files = [
+                node_path(parent_mount, mount_node)
+                for source in validated_config.sources_of_mounted_paths
+                for parent_mount, mount_node in source.get_mounted_paths()
+            ]
+            assert len(mounted_files) == len(set(mounted_files)), (
+                f"Mounted files are not unique in {name}: {validated_config.sources_of_mounted_paths}"
+            )
+            if container_spec["name"].startswith("render-config"):
```
benbz: Detect by `image` and `args` not name

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,524 @@
...
+        for key, content in self.data.items():
+            if key in skip_path_consistency_for_files:
+                continue
+            paths += match_path_in_content(content)
+        return paths
+
+
+# A consumer which refers to files through input env values and args/command of the container
+@dataclass(frozen=True)
+class GenericContainerSpecPathConsumer(PathConsumer):
```
benbz, line 353: We should look at exec probes too

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,524 @@
...
+                validated_config.paths_consumers.append(
+                    RenderedConfigPathConsumer.from_workload_spec(workload_spec, container_index, templates)
+                )
+            else:
+                validated_config.paths_consumers.append(
+                    GenericContainerSpecPathConsumer.from_container_spec(container_spec)
+                )
+        return validated_config
+
+    def check_paths_used_in_content(self, paths_consistency_noqa):
```
benbz: This function has no `assert` and nothing checking its return value.

tests/manifests/test_configs_consistency.py
benbz, line 1: Lots clearer than the current test.

There's a fair amount of passing around of all the containers to various functions. I think this should be avoided. We should just need the current volumeMounts and ensuring that any `emptyDirs` have been mutated with files generated in prior containers

gaelgatelement: This has now be refactored with the suggested logic.

tests/manifests/test_configs_consistency.py
gaelgatelement: What kind of paths would you like to catch in exec probes ? It feels it would contain paths coming from the container image, not from the chart templates?

tests/manifests/test_configs_consistency.py
gaelgatelement: I've moved all but `assert_exists_according_to_hook_weight` as it should move to another test in the end, and I'd prefer to do this move in a dedicated PR.

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,706 @@
...
+    def path_is_used_in_content(self, path) -> bool:
+        return find_keys_mounts_in_content(
+            path,
+            list(self.env.values())
+            + self.args
+            + [
+                rendered_content
+                for empty_dir in self.mounted_empty_dirs.values()
+                for file, rendered_content in empty_dir.render_config_outputs.items()
+            ],
```
benbz: Extract to a common variable/function as it is repeated

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,706 @@
...
+        )
+
+    def get_parent_mount_lookalikes(self, paths_consistency_noqa, parent_mount: ParentMount) -> list[str]:
+        lookalikes = []
+
+        for match_in in [
+            v
+            for v in list(self.env.values())
```
benbz: Why is this done differently to the above and below functions?

tests/manifests/test_configs_consistency.py
gaelgatelement: In the end, because of circular dependency, I've moved the assert function as well. But let's move it out in a future PR.

tests/manifests/test_configs_consistency.py
benbz: In Pro we've got at least 1 exec probe that runs Python to check for the existence of a file

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,706 @@
...
+        mounted_files = [
+            node_path(parent_mount, mount_node)
+            for source in validated_config.sources_of_mounted_paths
+            for parent_mount, mount_node in source.get_mounted_paths()
+        ]
+        assert len(mounted_files) == len(set(mounted_files)), (
+            f"Mounted files are not unique in {container_spec['name']}\n"
+            f"Duplicated files : { {item for item, count in Counter(mounted_files).items() if count > 1} }\n"
+            f"From Mounted Sources : {validated_config.sources_of_mounted_paths}"
+        )
```
benbz: This feels like it should be its own test, not part of the construction of a `ValidatedContainerConfig`

tests/manifests/__init__.py
```diff
@@ -110,6 +110,7 @@ class DeployableDetails(abc.ABC):
 
     paths_consistency_noqa: tuple[str, ...] = field(default=(), hash=False)
     skip_path_consistency_for_files: tuple[str, ...] = field(default=(), hash=False)
+    content_volumes_mapping: dict[str, tuple[str, ...]] = field(default_factory=dict, hash=False)
```
benbz, line 123: Can we document the differences between all 3 of these?

Edit: potentially we need `content_volumes_mapping`  and then different variables for each of the 3 checks (that I want split out into their own tests)

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,706 @@
...
+        return validated_config
+
+    def check_paths_used_in_content(self, skip_path_consistency_for_files, paths_consistency_noqa):
+        paths_not_found = []
+        skipped_paths = []
+        for source in self.sources_of_mounted_paths:
+            for parent_mount, mount_node in source.get_mounted_paths():
+                if (
+                    node_path(parent_mount, mount_node) in paths_consistency_noqa
+                    or parent_mount.path.startswith("/secrets")
```
benbz, line 607: Do we need this skip if `RenderConfigContainerPathConsumer.path_is_used_in_content` always returns `true`?

tests/manifests/test_configs_consistency.py
gaelgatelement: Because so far the logic has always been to not store configuration in secrets. We can certainly add an equivalent for Secrets anyway, just in case we need something later.

benbz: Let's not add it pre-emptively but let's add a comment where it would be needed explaining the situation

tests/manifests/test_configs_consistency.py
gaelgatelement: That's been cleaned up, thanks for catching

tests/manifests/test_configs_consistency.py
gaelgatelement: That's been cleaned up, thanks for catching

tests/manifests/test_configs_consistency.py
gaelgatelement: This is now done

tests/manifests/test_configs_consistency.py
gaelgatelement: I've reworked the `RenderConfigContainerPathConsumer` so that it tries to check a couple of paths that are testable.

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,770 @@
...
+
+@pytest.mark.parametrize("values_file", values_files_to_test | secret_values_files_to_test)
+@pytest.mark.asyncio_cooperative
+async def test_any_path_found_matches_an_actual_mount(templates, other_secrets):
+    traverse_containers_and_run_consistency_check(
+        templates,
+        other_secrets,
+        ValidatedContainerConfig,
+        ValidatedContainerConfig.check_all_paths_matches_an_actual_mount,
+    )
```
benbz: Naively looking at the test names here, I don't understand the difference between them. i.e. If there was a typo in a file name surely it would fail both tests

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,770 @@
...
+            for parent_mount, mount_node in source.get_mounted_paths():
+                mounted_paths.append(node_path(parent_mount, mount_node))
+        for source in self.sources_of_mounted_paths:
+            for parent_mount, _ in source.get_mounted_paths():
+                if parent_mount in paths_consistency_noqa:
+                    continue
+                for path_consumer in self.paths_consumers:
+                    for lookalikes in path_consumer.get_parent_mount_lookalikes(paths_consistency_noqa, parent_mount):
+                        assert lookalikes in mounted_paths, (
+                            f"{self.name} : Found path consumer mismatch in {path_consumer}"
```
benbz: If this assertion failed I would not understand what's gone wrong other than which container this was in.

tests/manifests/__init__.py
```diff
@@ -108,8 +108,15 @@ class DeployableDetails(abc.ABC):
     makes_outbound_requests: bool = field(default=None, hash=False)  # type: ignore[assignment]
     is_synapse_process: bool = field(default=False)
 
+    # Use this to noqa given paths that are found by the consistency checks
```
benbz: Sorry. It still isn't clear to me from the comment as to how this differs from `skip_path_consistency_for_files`. If this is meant to be "ignore these path-like strings anywhere", then let's keep the old variable for the old test and design a new one(s) that does much more specific match. e.g. I assume all the `/tmp` additions are because we've got a `volumeMount` that we don't reference, so let's have a `ignore_unreferenced_volume_mount = dict` which goes container name -> volume name.

tests/manifests/test_configs_consistency.py
gaelgatelement: I also think that, actually, the old "lookalike" test might not be required anymore now that we much more strictly discover mismatches. I'll check if we can get rid of it.

gaelgatelement: After testing, I'm convinced we can get rid of it. Let's drop it altogether now.

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,709 @@
...
+        pass
+
+    # Mutate empty dirs after the container has been consistency-checked
+    @abc.abstractmethod
+    def mutate_empty_dirs(self, container_spec, workload_spec, mutable_empty_dirs: dict[str, MountedEmptyDir]):
+        pass
+
+
+## Gets all mounted files in a given container
+def get_all_mounted_files(
```
benbz, line 292: This is used in 1 place that already has a `container_spec`. Pass that in rather than iterating over the containers again?

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,709 @@
...
+                templates,
+                other_secrets,
+                all_workload_empty_dirs,
+            )
+            consistency_check(
+                validated_container_config,
+                deployable_details.skip_path_consistency_for_files,
+                deployable_details.ignore_unreferenced_mounts,
+                deployable_details.ignore_paths_mismatches,
+            )
```
benbz: Rather than passing all these in, `ValidatedContainerConfig.from_container_spec` takes a `DeployableDetails`. It should save that and each test uses that as appropriate

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,709 @@
...
+            mounted_empty_dirs=mounted_empty_dirs,
+        )
+
+    def path_is_used_in_content(self, path) -> bool:
+        return find_keys_mounts_in_content(
+            path,
+            list(self.env.values()) + self.args + self._empty_dir_rendered_content() + list(self.exec_probes.values()),
+        )
+
+    def get_all_paths_in_content(self, skip_path_consistency_for_files):
```
benbz: `skip_path_consistency_for_files` isn't used here like it is for the other `PathConsumers`. Should it be?

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,709 @@
...
+                path,
+                [node_path(*self.output)]
+                + list(self.env.values())
+                + list(self.inputs_files.keys())
+                + list(self.inputs_files.values()),
+            )
+            # for now we deliberately mount too many files in config-templates
+            or path.startswith("/conf")
+            # we also deliberately ignore files which are in the same directory as our output
+            # as we are most certainly cumulating files here
```
benbz:
```suggestion
            # as we are most certainly accumulating files here
```

This does feel a bit strange to me though as we're only writing 1 file out here?

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,709 @@
...
+            f"Looked in {self.sources_of_mounted_paths}\n"
+        )
+
+    def mutate_empty_dirs(self, container_spec, workload_spec):
+        for consumer in self.paths_consumers:
+            consumer.mutate_empty_dirs(container_spec, workload_spec, self.mutable_empty_dirs)
+
+
+def traverse_containers_and_run_consistency_check(
+    templates, other_secrets, validated_container_class, consistency_check
```
benbz: We shouldn't need `validated_container_class` as it is always `ValidatedContainerConfig`.

We could make this function yield the current `ValidatedContainerConfig` and then the tests loop over this function and directly call the test in question

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,709 @@
...
+
+    def _empty_dir_rendered_content(self):
+        return [
+            rendered_content
+            for empty_dir in self.mounted_empty_dirs.values()
+            for file, rendered_content in empty_dir.render_config_outputs.items()
+        ]
+
+    @classmethod
+    def from_container_spec(cls, workload_spec, container_spec, previously_mounted_empty_dirs):
```
benbz, line 375: I've just noticed the Postgres `StatefulSet` has `spec.template.spec.lifecycle.preStop.exec.command` which should be checked too

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,709 @@
...
+    def check_paths_used_in_content(
+        self, skip_path_consistency_for_files, ignore_unreferenced_mounts, ignore_paths_mismatches
+    ):
+        paths_not_found = []
+        skipped_paths = []
+        for source in self.sources_of_mounted_paths:
+            for parent_mount, mount_node in source.get_mounted_paths():
+                if (
+                    node_path(parent_mount, mount_node) in ignore_unreferenced_mounts.get(self.name, [])
+                    or parent_mount.path.startswith("/secrets")
```
benbz, line 607: Can this be documented like the `/conf` ignore is elsewhere?

Is it possible to push trhis down into a specific `PathConsumer` or does it apply to all?

tests/manifests/test_configs_consistency.py
gaelgatelement: For Syn2Mas and MatrixUser we are writing 2 files (one for the `homeserver.yaml` and one for `mas-config.yaml`)

tests/manifests/test_configs_consistency.py
gaelgatelement: I think we dont need to consume it here, as these are not files which are being parsed but containers properties. This skip variable is here to skip a large block of text which would be painful to manually add an exception for all ignored paths.

I'll pass deployment details instead in here so that it makes more sense to be free to use the property or not in these methods implementations.

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,710 @@
...
+    env: dict[str, str] = field(default_factory=dict)
+    args: list[str] = field(default_factory=list)
+    mounted_empty_dirs: dict[str, MountedEmptyDir] = field(default_factory=dict)
+    exec_properties: dict[str, str] = field(default_factory=dict)
+
+    def _empty_dir_rendered_content(self):
+        return [
+            rendered_content
+            for empty_dir in self.mounted_empty_dirs.values()
+            for file, rendered_content in empty_dir.render_config_outputs.items()
```
benbz: (Nit) `file` is unused here, iterate over `values()`

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,710 @@
...
+        )
+
+    def get_all_paths_in_content(self, deployable_details: DeployableDetails):
+        paths = []
+        for content in (
+            list(self.env.values())
+            + list(self.exec_properties.values())
+            + self.args
+            + self._empty_dir_rendered_content()
+        ):
```
benbz: (Nit) These 2 methods construct the same list but in a different order. Unifiy

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,717 @@
...
+            mounted_empty_dirs=mounted_empty_dirs,
+        )
+
+    def path_is_used_in_content(self, path) -> bool:
+        return find_path_in_content(
+            path,
+            list(self.env.values())
+            + self.args
+            + self._empty_dir_rendered_content()
+            + list(self.exec_properties.values()),
```
benbz:
```suggestion
            self._all_container_content()
```

tests/manifests/test_configs_consistency.py
```diff
@@ -0,0 +1,717 @@
...
+        )
+
+    def get_all_paths_in_content(self, deployable_details: DeployableDetails):
+        paths = []
+        for content in (
+            list(self.env.values())
+            + list(self.exec_properties.values())
+            + self.args
+            + self._empty_dir_rendered_content()
+        ):
```
benbz:
```suggestion
        for content in self._all_container_content():
```

PR #818, opened 2025-10-20
Update Chart metadata

benbz:
Set more metadata in `Chart.yaml` as per https://github.com/helm/helm/issues/11062 to hopefully get renovate changelogs working and enhance things in artifacthub.io.

The `kubeVersion` is a bit of a guess, based on the fact that we rely on the new-style `Ingress` spec. Arguably we only support k8s versions that are upstream supported but lets be expansive for now

### DIFF
--- a/charts/matrix-stack/Chart.yaml
+++ b/charts/matrix-stack/Chart.yaml
@@ -4,7 +4,28 @@
 
 apiVersion: v2
 name: matrix-stack
-description: A Helm meta-chart for deploying a Matrix Stack from Element
-type: application
 version: 25.10.3-dev
+kubeVersion: '>= 1.22.0'
+description: A Helm chart for deploying a Matrix Stack from Element
+type: application
+keywords:
+  - ESS
+  - Element
+  - Matrix
+  - Synapse
+  - ElementWeb
+  - MAS
+  - MatrixAuthenticationService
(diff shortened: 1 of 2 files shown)


charts/matrix-stack/Chart.yaml
```diff
@@ -4,7 +4,27 @@
...
+kubeVersion: '>= 1.22.0'
+description: A Helm chart for deploying a Matrix Stack from Element
+type: application
+keywords:
+  - ESS
+  - Element
+  - Matrix
+  - Synapse
+  - ElementWeb
+  - MAS
```
gaelgatelement, line 17:
```suggestion
  - MAS
  - MatrixAuthenticationService
```

For completness sake

charts/matrix-stack/Chart.yaml
```diff
@@ -4,7 +4,27 @@
...
+  - ElementAdmin
+  - MatrixRTC
```
gaelgatelement, line 20:
```suggestion
  - MatrixRTC
  - LiveKit
```

Unsure about this one but maybe?

benbz: I'm going to leave this one out unless you feel strongly (sounds like you don't)

PR #819, opened 2025-10-20
fix(well-known): prioritize baseDomainRedirect.url

cleverer:
Previously, whenever elementWeb was enabled, the url property was silently ignore instead of, as expected, taking precedence.

### DIFF
--- a/newsfragments/819.fixed.md
+++ b/newsfragments/819.fixed.md
@@ -1,0 +1,3 @@
+Prioritize wellKnownDelegation.baseDomainRedirect.url over elementWeb.ingress.host.
+
+Previously, whenever elementWeb was enabled, the url property was silently ignored instead of, as expected, taking precedence.


cleverer:
Will try to add some tests covering the behavior in the evening.

benbz:
Thanks!

cleverer:
Well that's what i wanted to improve, as test with that would have caught this early on ;)

But if it's good enough for you we can merge no problem, i can just as well check if i can make some tests as a separate PR :)

benbz:
> Well that's what i wanted to improve, as test with that would have caught this early on ;)
> 
> But if it's good enough for you we can merge no problem, i can just as well check if i can make some tests as a separate PR :)

Happy to take a test if you want, especially as it would prevent regression here. Discussing internally the suggestions were adding an integration test that covered all default enabled components would be more valuable than just Well-Known Delegation + Element Web. Or a manifest test in `tests/manifests/test_haproxy.py` that looked for the specific HAProxy config line

PR #822, opened 2025-10-20
Check that files have copyright of the new Element legal entity on changes

benbz:
`Element Creations Ltd` is the new legal entity. We need to ensure that our checks look at this entity


benbz:
This PR doesn't change all files over to have the new copyright header yet. It updates our infra to handle multiple copyright headers and any generated files have the new copyright header.

Switching all files over will come in a distinct PR

gaelgatelement (approved):

PR #855, opened 2025-11-06
Configure MSC4143 support in Synapse when MatrixRTC is enabled

benbz:
Makes use of https://github.com/element-hq/synapse/pull/18967 for future proofing Matrix RTC support

### DIFF
--- a/charts/matrix-stack/configs/synapse/synapse-04-homeserver-overrides.yaml.tpl
+++ b/charts/matrix-stack/configs/synapse/synapse-04-homeserver-overrides.yaml.tpl
@@ -108,6 +108,8 @@ experimental_features:
 {{- if $root.Values.matrixRTC.enabled }}
   # MSC3266: Room summary API. Used for knocking over federation
   msc3266_enabled: true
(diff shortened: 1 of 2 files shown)


charts/matrix-stack/configs/synapse/synapse-04-homeserver-overrides.yaml.tpl
```diff
@@ -108,6 +108,8 @@ experimental_features:
 {{- if $root.Values.matrixRTC.enabled }}
   # MSC3266: Room summary API. Used for knocking over federation
   msc3266_enabled: true
+  # MSC4143: Matrix RTC Transport using Livekit Backend. Synapse supports enables  client-Server API for discovery of Matrix RTC backends
```
jaywink: Feels like there is a typo here, not sure what the intention was, but `Synapse supports enables client-server` doesn't feel right?

benbz: I think I changed what I was trying to say part way through. Does that look better now?

PR #859, opened 2025-11-10
Add a script + job to validate the image architectures

benbz:
Ensure we don't have any regressions on ARM64 support


jaywink (approved):
Looks good to me :+1:

PR #866, opened 2025-11-12
Merge replica tests for Deployments and StatefulSets

benbz:
Bring the `Deployment` and `StatefulSet` replica tests in sync.

Unclear why we don't do `podAntiAffinity` for `StatefulSets`. I suspect we should do for both or at least move both to `topologySpreadConstraints` given we don't have a way to set `podAntiAffinity` with values yet. WDYT about that plan for a future PR @gaelgatelement ?

### DIFF
--- a/newsfragments/866.internal.md
+++ b/newsfragments/866.internal.md
@@ -1,0 +1,1 @@
+CI: unify manifest tests around `Pod` replicas.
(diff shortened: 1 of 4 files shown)


benbz:
> Unclear why we don't do `podAntiAffinity` for `StatefulSets`. I suspect we should do for both or at least move both to `topologySpreadConstraints` given we don't have a way to set `podAntiAffinity` with values yet. WDYT about that plan for a future PR @gaelgatelement ?

This is being actioned in #867

jaywink (approved):

PR #871, opened 2025-11-14
Use k3d instead of kind

benbz:
k3d integrates a LB, metrics-server and a registry so we can simplify our config a fair bit.

It also appears faster to startup, at the cost of needing to retry 
```
resp = <Response [429 Too Many Requests]>

    @staticmethod
    def raise_for_status(resp):
        try:
            resp.raise_for_status()
        except httpx.HTTPError as e:
>           raise transform_exception(e)
E           lightkube.core.exceptions.ApiError: storage is (re)initializing
```


.github/workflows/pytest.yml
```diff
@@ -120,17 +119,17 @@ jobs:
       if: ${{ failure() }}
       shell: bash
       run: |
-        kind export logs --name ess-helm ./ess-helm-logs
```
gaelgatelement, line 127: We are loosing kubernetes system logs and pod logs with this move to k3d. It's quite useful to understand what happened in CI. We should have an equivalent with k3d before merging this I believe.

gaelgatelement (changes requested):

.github/workflows/pytest.yml
benbz: Yup I've run into the lack of this whilst debugging the CI failures. Oustanding things are
* Pod logs in CI
* Wait for 443/TCP to be available locally before `ingress_ready` returns
* Fix the flakiness with the pod networking tests which is now much frequent
* Fix whatever is causing HAProxy failures?
* Possibly do things with DockerHub in the k3d config itself

benbz:
HAProxy failures were e.g. `[pod/pytest-ejumskwk-haproxy-55c9d64477-lvv4p/haproxy] 2025-11-25T10:29:59.071210596Z [ALERT]    (8) : [haproxy.main()] Cannot raise FD limit to 80037, limit is 65536.` which is going to be resolved by #890 but doesn't help the upgrade case

.github/workflows/pytest.yml
```diff
@@ -124,17 +123,23 @@ jobs:
...
-        kind export kubeconfig --name ess-helm
-        ns=$(kubectl --context kind-ess-helm get ns -l app.kubernetes.io/managed-by=pytest  -o jsonpath='{.items[].metadata.name}')
+        mkdir ess-helm-logs
+        k3d kubeconfig merge ess-helm -ds
+        for ns in $(kubectl --context k3d-ess-helm get ns -o custom-columns=NS:.metadata.name --no-headers); do
+          mkdir -p "./ess-helm-logs/$ns"
+          for pod in $(kubectl --context k3d-ess-helm -n "$ns" get pod -o custom-columns=NS:.metadata.name --no-headers); do
+            kubectl --context k3d-ess-helm -n "$ns" logs --all-containers --prefix --timestamps "$pod" > "./ess-helm-logs/$ns/$pod.logs"
```
gaelgatelement: I'm wondering if we also want to have `--previous` logs ?

newsfragments/871.internal.md
```diff
@@ -0,0 +1 @@
+CI: switch from kind to k3d for integration tests.
```
gaelgatelement, line 1: `DEVELOPERS.md` needs an update

tests/integration/test_networking.py
```diff
@@ -23,6 +23,7 @@
 async def test_services_have_matching_labels(
     kube_client: AsyncClient,
     generated_data: ESSData,
+    matrix_stack,
```
gaelgatelement: matrix_stack fixture has `autouse` enabled. Do we really need this ? 


```
@pytest.fixture(autouse=True, scope="session")
async def matrix_stack(
```

benbz: This came about as a result of e.g. https://github.com/element-hq/ess-helm/actions/runs/19760559262/job/56621295246. That test has `matrix_stack` but doesn't have `prometheus_operator_crds` which is another `autouse=True` CRD. The only way I could see a complete failure of monitored pods but all monitorable pods being present is if the `ServiceMonitor` CRD wasn't yet present in cluster, which makes the reliance on `autouse=True` highly suspect.

I can try dropping these and just keeping the `prometheus_operator_crds` fixture being an explicit dependency of the chart

benbz: This appears to have worked. On reflection `autouse=True` just means that the fixture is used before the test / any more tightly scoped fixtures. There'd be a race between the ESS & Prometheus Operator CRD chart installations. The only required change was to make the Prometheus Operator CRD chart installation and explicit dependency of the ESS chart installation

PR #879, opened 2025-11-20
Fix element-web and matrix-rtc sfu ingresses not targetting port names

gaelgatelement:


newsfragments/879.fixed.md
```diff
@@ -0,0 +1 @@
+Fix element-web and matrix-rtc sfu ingresses not targetting service backend port using port names.
```
benbz:
```suggestion
Change Element Web and MatrixRTC SFU `Ingresses` to target `Service` port names rather than numbers
```

And the fragment should be a changed one as I don't think there's a bug fix here

benbz (approved):

PR #884, opened 2025-11-21
Improve the hook tests by asserting presence/absence of Hook annotations based on DeployableDetails

benbz:
Improves #880 so that we can catch stray `helm.sh/hook` annotations on resources that shouldn't have it or missing annotations where they should be

### DIFF
--- a/charts/matrix-stack/templates/matrix-authentication-service/configmap_hook.yaml
+++ b/charts/matrix-stack/templates/matrix-authentication-service/configmap_hook.yaml
@@ -15,6 +15,7 @@ metadata:
   namespace: {{ $.Release.Namespace }}
   labels:
     {{- include "element-io.matrix-authentication-service.labels" (dict "root" $ "context" $masContext) | nindent 4 }}
+    "k8s.element.io/hook-for": "synapse-check-config"
   annotations:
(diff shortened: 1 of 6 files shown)


charts/matrix-stack/templates/matrix-authentication-service/configmap_hook.yaml
```diff
@@ -18,6 +18,7 @@ metadata:
   annotations:
     "helm.sh/hook": pre-install,pre-upgrade
     "helm.sh/hook-weight": "-1"
+    "k8s.element.io/hook-for": "synapse-check-config"
```
gaelgatelement: Do we want to enforce having this metadata on all CM/Secret created during hook phases?

benbz: We can but I don't think we can assert the value of annotation so it is of questionable value other than making the hook test pass.

benbz: Unless you have strong feelings I'm going to omit this (now) label from everything else for now. If/when we find a need for it, we can add to everything

PR #890, opened 2025-11-24
Remove explicit configuration of HAProxy `maxconn` at the global and frontend level.

### LINKED ISSUE #886: HAProxy 80,064 file descriptor exceeds container limit 65535
- Reduced HAProxy maxconn from 40,000 to 30,000 in the global section
- Reduced HAProxy maxconn from 20,000 to 15,000 in the defaults section
This fixed the max file descriptor issue.

benbz:
Fixes #886.

At the global level it is computed based on `ulimit -n` as per https://docs.haproxy.org/3.1/configuration.html#maxconn.

At the frontend level, it defaults to the global `maxconn` if omitted.

Removing `fullconn` from `defaults` removes it from each `backend`. This defaults to 10% of each the `maxconn` for each `frontend` that routes to the backend. Given each `server` has much lower `maxconn` it feels fine to remove this.

The previous values were a problem on at least microk8s which didn't raise the ulimits like most other clusters

### DIFF
--- a/charts/matrix-stack/configs/haproxy/haproxy.cfg.tpl
+++ b/charts/matrix-stack/configs/haproxy/haproxy.cfg.tpl
@@ -9,7 +9,6 @@ SPDX-License-Identifier: AGPL-3.0-only
 {{- with required "haproxy/haproxy.cfg.tpl missing context" .context -}}
 
 global
-  maxconn 40000
   log stdout format raw local0 info
 
   # Allow for rewriting HTTP headers (e.g. Authorization) up to 4k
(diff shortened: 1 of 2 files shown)


benbz:
Output from HAProxy using `socat`

Before:
```
$ echo "show info" | socat stdio tcp4:localhost:1999
...
Ulimit-n: 80076
Maxsock: 80076
Maxconn: 40000
Hard_maxconn: 40000
...
```

After:
```sh
$ echo "show info" | socat stdio tcp4:localhost:1999
...
Ulimit-n: 1048576
Maxsock: 1048576
Maxconn: 524250
Hard_maxconn: 524250
...
```

gaelgatelement (approved):

PR #905, opened 2025-12-02
Listen on IPv6 addresses for HAProxy traffic

bardiharborow:
This PR modifies the HAProxy configuration file to direct HAProxy to create sockets in dual-stack mode so that traffic can be received via IPv4 or IPv6. Previously, sockets were created in IPv4 mode only. This was particularly problematic in IPv6-only environments where IPv4 is not routable.

Docs: https://www.haproxy.com/documentation/haproxy-configuration-tutorials/proxying-essentials/configuration-basics/frontends/#listen-on-multiple-ip-addresses-and-ports

### DIFF
--- a/charts/matrix-stack/configs/haproxy/haproxy.cfg.tpl
+++ b/charts/matrix-stack/configs/haproxy/haproxy.cfg.tpl
@@ -83,12 +83,14 @@ resolvers kubedns
 
 frontend prometheus
   bind *:8405
+  bind [::]:8405
   http-request use-service prometheus-exporter if { path /metrics }
   monitor-uri /haproxy_test
   no log
 
 frontend http-blackhole
   bind *:8009
+  bind [::]:8009
 
   # same as http log, with %Th (handshake time)
   log-format "%ci:%cp [%tr] %ft %b/%s %Th/%TR/%Tw/%Tc/%Tr/%Ta %ST %B %CC %CS %tsc %ac/%fc/%bc/%sc/%rc %sq/%bq %hr %hs %{+Q}r"
--- a/charts/matrix-stack/configs/synapse/partial-haproxy.cfg.tpl
+++ b/charts/matrix-stack/configs/synapse/partial-haproxy.cfg.tpl
@@ -10,6 +10,7 @@ SPDX-License-Identifier: AGPL-3.0-only
 
 frontend startup
    bind *:8406
+   bind [::]:8406
    acl synapse_dead nbsrv(synapse-main) lt 1
 {{- range $workerType, $_ := (include "element-io.synapse.enabledWorkers" (dict "root" $root)) | fromJson }}
 {{- if not (include "element-io.synapse.process.canFallbackToMain" (dict "root" $root "context" $workerType)) }}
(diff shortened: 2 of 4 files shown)


benbz:
Thanks for raising this. We're not opposed to these changes however we're very nervous about the HAProxy `bind` change - we started off with this exact `bind` in ESS Classic and ended up changing to the current IPv4 only `bind` after various issues. We need to carefully think about how to test this and whether it should be configurable given the history of how we got here.

In terms of the `ipFamilyPolicy` change, that feels much less controversial. I see you've found the prior stalled attempt in #660. Given you highlight that there's value on doing this on not just the MatrixRTC `Services`, I think this should be consistently done in every `Service` and validated with a manifest test in https://github.com/element-hq/ess-helm/blob/main/tests/manifests/test_services.py. We'd probably prefer to take this as an independent PR if that's ok?

bardiharborow:
> we started off with this exact bind in ESS Classic

@benbz ah, it makes sense to exercise some caution then. I am happy to help with the testing plan if you can establish what the historical issues were. Synapse also isn't listening over IPv6 currently. That was going to be the subject of a different PR, but perhaps that also contributed to the historical issues, I'm not sure.

Another option which may behave differently is to create two different sockets:

```
bind *:8008
bind [::]:8008
```

> We'd probably prefer to take this as an independent PR if that's ok?

I have separated out the `ipFamilyPolicy` changes to #907.

benbz:
> > we started off with this exact bind in ESS Classic
> 
> @benbz ah, it makes sense to exercise some caution then. I am happy to help with the testing plan if you can establish what the historical issues were. Synapse also isn't listening over IPv6 currently. That was going to be the subject of a different PR, but perhaps that also contributed to the historical issues, I'm not sure.
> 
> Another option which may behave differently is to create two different sockets:
> 
> ```
> bind *:8008
> bind [::]:8008
> ```
> 
> > We'd probably prefer to take this as an independent PR if that's ok?
> 
> I have separated out the `ipFamilyPolicy` changes to #907.

You've nerd-sniped me today and I've been investigating the situation. Disabling IPv6 via sysctl doesn't appear enough to trigger what the customer was seeing. The error in question was `Address family not supported by protocol`. I believe this means that either the server itself was started with `ipv6.disable=1` or the kernel had IPv6 support compiled out. I suspect that means anything using IPv6 addresses for binding would fail, so splitting out the binds probably wouldn't work either.

I *think* Element Admin, Element Web, Postgres, the MatrixRTC authoriser and the Matrix SFU all currently bind IPv6. At least some of them probably do it in a way that would work if IPv6 was disabled, but I think Element Admin & Element Web would fail as they do explicit `[::]` binds in their nginx config.

Given we've got some hard-coded IPv6 binds today I'm inclined to press forward with supporting IPv6 binds across all components without configuration. I'd suggest they're done as 2 separate binds like you've proposed above so that it is easier to gate one or the other later on

benbz:
Remaining failure is unrelated to this PR and is fixed going forward in #909. Merging

TYVM :pray:

benbz:
> Given we've got some hard-coded IPv6 binds today I'm inclined to press forward with supporting IPv6 binds across all components without configuration. I'd suggest they're done as 2 separate binds like you've proposed above so that it is easier to gate one or the other later on

Of course immediately someone popped out of the woodwork with issues in HAProxy after deploying this. So #921 makes it configurable. 

I do wonder if Element Web/Element Admin doesn't have issues for them because the `listen [::]:8080` has implicit `ipv6only=on` that changes the behaviour

PR #907, opened 2025-12-02
Expose services as dual-stack where possible

bardiharborow:
This PR directs Kubernetes to assign both IPv4 and IPv6 ClusterIPs to all services where possible. This change is applied to all services [following discussion](https://github.com/element-hq/ess-helm/pull/905#issuecomment-3601556460) with @benbz.

Closes #660 as redundant.


benbz (approved):
Failure is due to lack of GHA Secrets. Will resolve separately

PR #910, opened 2025-12-03
Matrix Authentication Service: maxSurge to 0

gaelgatelement:
Change Matrix Authentication Service deployment `maxSurge` to 0.

We have seen migrations race conditions happening during Matrix Authentication Service pods
rollout. This sets `maxSurge` to 0 to try to make sure only 1 pod at a time runs the
migration process.

### DIFF
--- a/newsfragments/910.changed.md
+++ b/newsfragments/910.changed.md
@@ -1,4 +1,4 @@
-Change Matrix Authentication Service deployment `maxSurge` to 1.
+Change Matrix Authentication Service deployment `maxSurge` to 0.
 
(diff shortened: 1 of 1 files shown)


gaelgatelement:
> Have you confirmed that the update behaviour is 1-by-1? As I could believe that with `replicas>1` it kills 1 old one (`maxUnavailable: 1`), starts 2 new ones (`maxSurge: 1`). Possibly for the 1-by-1 case we also need to do `maxUnavailable: 0`

While testing 2 replicas with `maxSurge: 0`, MAS did a one-by-one rollout. In which regards do you think `maxUnavailable: 0` would help here?

PR #921, opened 2025-12-04
Introduce networking.ipFamily to switch between IPv4, IPv6 or both binds

benbz:
MAS' behaviour relies on [binding both if `host` is omitted](https://github.com/element-hq/matrix-authentication-service/blob/v1.7.0/crates/cli/src/server.rs#L377-L380).

Synapse relies on [ignoring errors binding to 0.0.0.0 2nd](https://github.com/element-hq/synapse/blob/v1.143.0/synapse/app/__init__.py#L38).

`v6only` is introduced on the HAProxy binds so that IPv4 mapped addresses don't show up, they go to the IPv4 bind

### DIFF
--- a/charts/matrix-stack/templates/element-admin/deployment.yaml
+++ b/charts/matrix-stack/templates/element-admin/deployment.yaml
@@ -15,6 +15,7 @@ metadata:
 {{- end }}
   labels:
     {{- include "element-io.element-admin.labels" (dict "root" $ "context" .) | nindent 4 }}
+    k8s.element.io/element-admin-config-hash: "{{( include "element-io.element-admin.configmap-data" (dict "root" $ "context" .)) | sha1sum }}"
   name: {{ $.Release.Name }}-element-admin
   namespace: {{ $.Release.Namespace }}
 spec:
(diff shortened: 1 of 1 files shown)


charts/matrix-stack/configs/haproxy/haproxy.cfg.tpl
```diff
@@ -82,15 +82,23 @@ resolvers kubedns
   hold refused 600s
 
 frontend prometheus
+{{- if has $root.Values.networking.ipFamily (list "ipv4" "dual-stack") }}
   bind *:8405
-  bind [::]:8405
+{{- end }}
+{{- if has $root.Values.networking.ipFamily (list "ipv6" "dual-stack") }}
+  bind [::]:8405 v6only
```
gaelgatelement: Can you mention the comment from the PR here, I suspect it will be rather confusing for future selves.

benbz: Done after tweaking this to be conditionally `v6only`/`v6v4` depending on the `ipFamily`

PR #635, opened 2025-07-30
matrix-authorisation-service: upgrade lk-jwt-service

gaelgatelement:
- add `restrictRoomCreationToLocalUsers`, default to false

PR #926, opened 2025-12-08
Toggle SFU room creation restrictions to on

benbz:
Added in https://github.com/element-hq/ess-helm/pull/635 with `0.3.0` of the MatrixRTC authoriser service. Controls whether `LIVEKIT_FULL_ACCESS_HOMESERVERS` is set to `$.Values.serverName` (true) or `*` (false).

Unsure if this can be toggled yet, so draft and cc @fkwp for advice

### DIFF
--- a/charts/matrix-stack/source/matrix-rtc.yaml.j2
+++ b/charts/matrix-stack/source/matrix-rtc.yaml.j2
@@ -10,8 +10,7 @@ enabled: true
 
 # Allows restricting room creation to local users only
 # Remote federated users can still join the room
-# Defaults to false for now until clients are upgraded to support the new mechanism
-restrictRoomCreationToLocalUsers: false
+restrictRoomCreationToLocalUsers: true
 
 # LiveKit Authentication Configuration
 # This section allows you to configure authentication for the LiveKit SFU.
--- a/charts/matrix-stack/values.yaml
+++ b/charts/matrix-stack/values.yaml
@@ -416,8 +416,7 @@ matrixRTC:
 
   # Allows restricting room creation to local users only
   # Remote federated users can still join the room
-  # Defaults to false for now until clients are upgraded to support the new mechanism
-  restrictRoomCreationToLocalUsers: false
+  restrictRoomCreationToLocalUsers: true
 
   # LiveKit Authentication Configuration
   # This section allows you to configure authentication for the LiveKit SFU.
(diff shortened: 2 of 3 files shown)


fkwp:
does this change also configures the SFU to restrict room creation ?

```
room:
  auto_create: false
```

benbz:
> does this change also configures the SFU to restrict room creation ?
> 
> ```
> room:
>   auto_create: false
> ```

That is hard-coded off regardless of the value of `matrixRTC.restrictRoomCreationToLocalUsers`.

fkwp:
> Added in #635 with `0.3.0` of the MatrixRTC authoriser service. Controls whether `LIVEKIT_FULL_ACCESS_HOMESERVERS` is set to `$.Values.serverName` (true) or `*` (false).
> 
> Unsure if this can be toggled yet, so draft and cc @fkwp for advice

It would work. It's important that the lk-jwt-service can reach out to the SFU to create the room in advance. However, that's already the case as the config is currently defaulting to `*`

benbz:
> > Added in #635 with `0.3.0` of the MatrixRTC authoriser service. Controls whether `LIVEKIT_FULL_ACCESS_HOMESERVERS` is set to `$.Values.serverName` (true) or `*` (false).
> > Unsure if this can be toggled yet, so draft and cc @fkwp for advice
> 
> It would work. It's important that the lk-jwt-service can reach out to the SFU to create the room in advance. However, that's already the case as the config is currently defaulting to `*`

So as per the original comment `until clients are upgraded to support the new mechanism`, enough clients now support the new mechanism and enough time has passed since they gained support for it?

fkwp:
> > > Added in #635 with `0.3.0` of the MatrixRTC authoriser service. Controls whether `LIVEKIT_FULL_ACCESS_HOMESERVERS` is set to `$.Values.serverName` (true) or `*` (false).
> > > Unsure if this can be toggled yet, so draft and cc @fkwp for advice
> > 
> > 
> > It would work. It's important that the lk-jwt-service can reach out to the SFU to create the room in advance. However, that's already the case as the config is currently defaulting to `*`
> 
> So as per the original comment `until clients are upgraded to support the new mechanism`, enough clients now support the new mechanism and enough time has passed since they gained support for it?

I See, it should be fine.

benbz:
> Change ok. I am wondering however if this should be listed in the breaking changes section of the changelog? On the fence so not pushing for it.

I don't think it is particularly important, but let's do it just in case

PR #928, opened 2025-12-08
Make RSA key length explicit

benbz:
Builds on top of #927 as all the secrets types weren't documented beforehand


gaelgatelement (approved):

benbz:
Holding off on this until we have another reason to do `matrix-tools` release as this is just cleanup

benbz:
Replaced by #973

PR #950, opened 2025-12-23
element-web: Add support for extra volumes

gaelgatelement:
This is a standard kubernetes feature that can be used to mount files in the element-web container.


tests/manifests/utils.py
```diff
@@ -186,6 +191,35 @@ def find_credential(values_fragment):
...
+    for cm_name in find_extra_configmaps(values):
+        yield {
+            "kind": "ConfigMap",
+            "metadata": {
+                "name": cm_name,
+                "annotations": {
+                    # We simulate the fact that it exists before the chart deployment
+                    # using the hook weight.
+                    # Actually it does not have any
+                    # but this is necessary for tests/manifests/test_configs_and_mounts_consistency.py
```
benbz:
```suggestion
                    # but this is necessary for tests/manifests/test_configs_consistency.py
```

gaelgatelement:
Done with #957

PR #952, opened 2025-12-23
matrix-tools: now uses go 1.25, adjust CI accordingly.

gaelgatelement:
The version changed when merging https://github.com/element-hq/ess-helm/pull/947


benbz:
> The version changed when merging #947

Is there anything we can/should do in CI to catch this in future? A naive grep and compare of the 3 places we've got the Go version set?

PR #965, opened 2026-01-08
postgres-exporter: support extra volume mounts

gaelgatelement:
There's no reason not to support volume mounts for sidecars.

### DIFF
--- a/newsfragments/965.added.md
+++ b/newsfragments/965.added.md
@@ -1,0 +1,1 @@
+Add support for `extraVolumes` and `extraVolumeMounts` in components values.


tests/manifests/__init__.py
```diff
@@ -241,7 +241,6 @@ def __post_init__(self):
             PropertyType.PodSecurityContext: ValuesFilePath.not_supported(),
             PropertyType.ServiceAccount: ValuesFilePath.not_supported(),
             PropertyType.Volumes: ValuesFilePath.not_supported(),
```
benbz, line 243:
```suggestion
```

Anything listed in the sidecar's `extraVolumeMounts` needs to be loaded from its `extraVolumes`

gaelgatelement: @benbz Arguably the `postgres` `extraVolumes` are where the sidecar volumes needs to be put ?

charts/matrix-stack/source/element-web.yaml.j2
```diff
@@ -17,7 +17,8 @@ enabled: true
 ## Most settings are configurable but some settings are owned by the chart and can't be overwritten
 additional: {}
 
-{{- sub_schema_values.extraVolumesAndMounts("ElementWeb") }}
+{{- sub_schema_values.extraVolumes("ElementWeb") }}
```
benbz:
```suggestion
{{- sub_schema_values.extraVolumes("Element Web") }}
```

charts/matrix-stack/source/element-web.yaml.j2
```diff
@@ -17,7 +17,8 @@ enabled: true
...
+{{- sub_schema_values.extraVolumeMounts("ElementWeb") }}
```
benbz:
```suggestion
{{- sub_schema_values.extraVolumeMounts("Element Web") }}
```

tests/manifests/__init__.py
benbz: I can get on board with that now that the values fragments have been split and sidecars don't get `extraVolumes`.

There is an argument that sidecars could have `extraVolumes` and they'd be added to the parent components `volumes`.

But what's here works

PR #972, opened 2026-01-16
redis: move redis to a shared instance

gaelgatelement:
- We are going to re-use the same redis with no persistence for multiple components in the near future (hookshot, synapse)

### DIFF
--- a/newsfragments/972.removed.md
+++ b/newsfragments/972.removed.md
@@ -1,1 +1,3 @@
-redis: move synapse-redis to a shared component that can be used by multiple components of the chart. There is no impact when using default values, but if you have customize `synapse.redis` values, you will need to update them to use `redis`.
+Move Synapse's Redis to a top-level shared component that can be used by multiple components of the chart.
+
+There is no impact when using the default values, but if you have customised values under the `synapse.redis` key, you will need to update them to be under the new top-level `redis` redis.


charts/matrix-stack/ci/pytest-matrix-authentication-service-syn2mas-values.yaml
benbz, line 1: Needs `redis-pytest-base-extras.yaml`. Same comment to all the other PyTest CI files below

newsfragments/972.changed.md
benbz, line 1: I wonder whether this should be `.removed` as that shows up as `Removed / Breaking Changes`. We'd want to say that there should be no impact with the default values, but if you've customised your `synapse.redis` values in any way then there will be some breakage

charts/matrix-stack/templates/redis/_helpers.tpl
```diff
@@ -0,0 +1,27 @@
...
+redis.conf: |
+{{- ($root.Files.Get "configs/redis/redis.conf") | nindent 2 -}}
+{{- end -}}
+
+
+{{- define "element-io.redis.labels" -}}
+{{- $root := .root -}}
+{{- with required "element-io.redis.labels missing context" .context -}}
+{{ include "element-io.ess-library.labels.common" (dict "root" $root "context" (dict "labels" .labels "withChartVersion" .withChartVersion)) }}
+app.kubernetes.io/component: matrix-server-pubsub
```
benbz: Is this description going to be accurate with how Hookshot is going to use Redis?

benbz:
```suggestion
app.kubernetes.io/component: matrix-pubsub-small-cache
```

charts/matrix-stack/configs/redis/redis.conf.tpl
benbz, line 1: Should have `maxmemory` (from values or computed from requests/limits) and `maxmemory-policy` as discussed

newsfragments/972.removed.md
```diff
@@ -0,0 +1 @@
+redis: move synapse-redis to a shared component that can be used by multiple components of the chart. There is no impact when using default values, but if you have customize `synapse.redis` values, you will need to update them to use `redis`.
```
benbz:
```suggestion
Move Synapse's Redis to a top-level shared component that can be used by multiple components of the chart.

There is no impact when using the default values, but if you have customised values under the `synapse.redis` key, you will need to update them to be under the new top-level `redis` redis.
```

charts/matrix-stack/templates/redis/_helpers.tpl
```diff
@@ -0,0 +1,29 @@
...
+{{- (tpl ($root.Files.Get "configs/redis/redis.conf") (dict "root" $root "context" .)) | nindent 2 -}}
+{{- end -}}
+{{- end -}}
+
+
+{{- define "element-io.redis.labels" -}}
+{{- $root := .root -}}
+{{- with required "element-io.redis.labels missing context" .context -}}
+{{ include "element-io.ess-library.labels.common" (dict "root" $root "context" (dict "labels" .labels "withChartVersion" .withChartVersion)) }}
+app.kubernetes.io/component: matrix-server-pubsub-small-cache
```
benbz:
```suggestion
app.kubernetes.io/component: matrix-pubsub-small-cache
```

As it isn't just for the `matrix-server` component anymore

charts/matrix-stack/source/redis.yaml.j2
```diff
@@ -0,0 +1,27 @@
+{#
+Copyright 2026 Element Creations Ltd
+
+SPDX-License-Identifier: AGPL-3.0-only
+#}
+
+# redis maxmemory and maxmemory-policy settings. See https://redis.io/docs/latest/develop/reference/eviction/
+maxMemory: 40mb
```
benbz, line 9:
```suggestion
# This value should be updated with changes to Redis' memory requests and limits below
maxMemory: 40mb
```

charts/matrix-stack/ci/pytest-matrix-authentication-service-syn2mas-values.yaml
benbz: These 3 PyTest files don't have Synapse workers, so Redis isn't being enabled

PR #976, opened 2026-01-19
Add support for Turn-TLS to the SFU

gaelgatelement:
- This adds the possibility to configure service types
- This adds support to configure a cert-manager Certificate for the Turn TLS service

### DIFF
--- a/charts/matrix-stack/templates/matrix-rtc/sfu_rtc_tcp_service.yaml
+++ b/charts/matrix-stack/templates/matrix-rtc/sfu_rtc_tcp_service.yaml
@@ -14,7 +14,7 @@ kind: Service
 metadata:
   labels:
     {{- include "element-io.matrix-rtc-sfu-rtc.labels" (dict "root" $ "context" .) | nindent 4 }}
-  {{- with .annotations }}
+  {{- with .exposedServices.rtcTcp.annotations }}
   annotations:
     {{- toYaml . | nindent 4 }}
   {{- end }}
--- a/charts/matrix-stack/templates/matrix-rtc/sfu_rtc_udp_muxer_service.yaml
+++ b/charts/matrix-stack/templates/matrix-rtc/sfu_rtc_udp_muxer_service.yaml
@@ -14,7 +14,7 @@ kind: Service
 metadata:
   labels:
     {{- include "element-io.matrix-rtc-sfu-rtc.labels" (dict "root" $ "context" .) | nindent 4 }}
-  {{- with .annotations }}
+  {{- with .exposedServices.rtcMuxedUdp.annotations }}
   annotations:
     {{- toYaml . | nindent 4 }}
   {{- end }}
--- a/charts/matrix-stack/templates/matrix-rtc/sfu_turn_tls_service.yaml
+++ b/charts/matrix-stack/templates/matrix-rtc/sfu_turn_tls_service.yaml
@@ -14,7 +14,7 @@ kind: Service
 metadata:
   labels:
     {{- include "element-io.matrix-rtc-sfu-turn-tls.labels" (dict "root" $ "context" .) | nindent 4 }}
-  {{- with .annotations }}
+  {{- with .exposedServices.turnTLS.annotations }}
   annotations:
     {{- toYaml . | nindent 4 }}
   {{- end }}
--- a/tests/manifests/__init__.py
+++ b/tests/manifests/__init__.py
@@ -13,6 +13,7 @@ class PropertyType(Enum):
     AdditionalConfig = "additional"
     Enabled = "enabled"
     Env = "extraEnv"
+    ExposedServices = "exposedServices"
     HostAliases = "hostAliases"
     Image = "image"
     Ingress = "ingress"
@@ -101,6 +102,7 @@ class DeployableDetails(abc.ABC):
 
     has_additional_config: bool = field(default=None, hash=False)  # type: ignore[assignment]
     has_db: bool = field(default=False, hash=False)
+    has_exposed_services: bool = field(default=False, hash=False)
     has_image: bool = field(default=None, hash=False)  # type: ignore[assignment]
     has_ingress: bool = field(default=True, hash=False)
     has_automount_service_account_token: bool = field(default=False, hash=False)
@@ -531,6 +533,7 @@ all_components_details = [
             SubComponentDetails(
                 name="matrix-rtc-sfu",
                 values_file_path=ValuesFilePath.read_write("matrixRTC", "sfu"),
+                has_exposed_services=True,
                 has_ingress=False,
                 has_replicas=False,
                 makes_outbound_requests=False,
--- a/tests/manifests/test_services.py
+++ b/tests/manifests/test_services.py
@@ -1,13 +1,100 @@
 # Copyright 2025 New Vector Ltd
-# Copyright 2025 Element Creations Ltd
+# Copyright 2025-2026 Element Creations Ltd
 #
 # SPDX-License-Identifier: AGPL-3.0-only
 
 import pytest
 import yaml
 
-from . import services_values_files_to_test, values_files_to_test
-from .utils import template_id
+from . import (
+    DeployableDetails,
+    PropertyType,
+    all_deployables_details,
+    services_values_files_to_test,
+    values_files_to_test,
+)
+from .utils import iterate_deployables_parts, template_id, template_to_deployable_details
+
+
+@pytest.mark.parametrize("values_file", ["matrix-rtc-exposed-services-values.yaml"])
+@pytest.mark.asyncio_cooperative
+async def test_exposed_services_configurability(values, make_templates):
+    num_of_expected_ports = {}
+    num_of_expected_exposed_services = {}
+
+    def exposed_service_mode(deployable_details: DeployableDetails, mode: str):
+        nonlocal num_of_expected_ports
+        nonlocal num_of_expected_exposed_services
+        exposed_services = deployable_details.get_helm_values(values, PropertyType.ExposedServices)
+        assert exposed_services is not None
+        test_annotations_labels = {
+            "exposed-service-port-type": mode,
+        }
+        exposed_services_values = dict(exposed_services)
+        for service_name in exposed_services:
+            exposed_services_values[service_name]["portType"] = mode
+            exposed_services_values[service_name]["annotations"] = test_annotations_labels
+            num_of_expected_exposed_services[deployable_details.name] += 1
+            if "portRange" in exposed_services_values[service_name]:
+                num_of_expected_ports[deployable_details.name] += (
+                    exposed_services_values[service_name]["portRange"]["endPort"]
+                    - exposed_services_values[service_name]["portRange"]["startPort"]
+                )
+            else:
+                num_of_expected_ports[deployable_details.name] += 1
+        deployable_details.set_helm_values(values, PropertyType.ExposedServices, exposed_services_values)
+        deployable_details.set_helm_values(values, PropertyType.Labels, test_annotations_labels)
+
+    found_exposed_services = {}
+    found_services = {}
+    for service_mode in ("NodePort", "HostPort", "LoadBalancer"):
+        for deployable_details in all_deployables_details:
+            num_of_expected_ports[deployable_details.name] = 0
+            num_of_expected_exposed_services[deployable_details.name] = 0
+
+        for deployable_details in all_deployables_details:
+            found_exposed_services[deployable_details.name] = []
+            found_services[deployable_details.name] = []
+
+        iterate_deployables_parts(
+            lambda deployable_details, service_mode=service_mode: exposed_service_mode(
+                deployable_details, service_mode
+            ),
+            lambda deployable_details: deployable_details.has_exposed_services,
+        )
+        for template in await make_templates(values):
+            deployable_details = template_to_deployable_details(template)
+            if deployable_details.has_exposed_services:
+                if template["kind"] == "Service":
+                    found_services[deployable_details.name].append(template)
+                    if "exposed-service-port-type" in template["metadata"].get("annotations", {}):
+                        assert (
+                            template["spec"]["type"] == template["metadata"]["annotations"]["exposed-service-port-type"]
+                        )
+                        found_exposed_services[deployable_details.name].append(template)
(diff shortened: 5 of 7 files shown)


.github/workflows/templates-dyff.yml
benbz, line 1: And in `.github/workflows/build-test.yml`

charts/matrix-stack/source/common/exposedServicePortRange.json
benbz, line 1: Any reason to leave off `annotations` here?

charts/matrix-stack/templates/matrix-rtc/sfu_turn_tls_certificate.yaml
```diff
@@ -0,0 +1,31 @@
...
+{{- with .sfu -}}
+{{- if $.Capabilities.APIVersions.Has "cert-manager.io/v1/Certificate" }}
+{{- if and .enabled .exposedServices.turnTLS.enabled (not .tlsSecret) $.Values.certManager -}}
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: {{ $.Release.Name }}-turn-tls
+  namespace: {{ $.Release.Namespace }}
+spec:
+  commonName: {{ tpl $.Values.turn.domain $ }}
```
benbz:
```suggestion
  commonName: {{ tpl .exposedServices.turnTLS.domain $ }}
```

charts/matrix-stack/templates/matrix-rtc/sfu_turn_tls_service.yaml
benbz, line 1: This file seems to assume hard-coded to `NodePort` in places (lines 11 & 32) but configurable in others (line 24). Given the options for `matrixRTC.exposedServices.turnTLS.portType`, I'd assume lines 11 & 32 are wrong and that there should be a check regarding `hostPort` somewhere?

charts/matrix-stack/templates/matrix-rtc/sfu_deployment.yaml
```diff
@@ -157,13 +167,32 @@ spec:
...
+{{- if .exposedServices.turnTLS.enabled }}
+{{- if .exposedServices.turnTLS.tlsSecret }}
+      - name: secret-turn-tls
+        secret:
+          secretName: {{ .exposedServices.turnTLS.tlsSecret }}
+{{- else if $.Values.certManager }}
+      - name: secret-turn-tls
+        secret:
+          secretName: "{{ .Release.Name }}-turn-tls-certmanager-tls"
+{{- end }}
```
benbz, line 194: Needs an `else` here and/or a validation that one of `tlsSecret` or `certManager` is configured (taking into account whether the cert-manager CRDs are in cluster or not)

benbz (commented):
Needs integration tests and for the TURN port to be added to the test cluster

charts/matrix-stack/templates/matrix-rtc/_helpers.tpl
```diff
@@ -12,6 +12,9 @@ SPDX-License-Identifier: AGPL-3.0-only
 {{- if not .ingress.host -}}
 {{ $messages = append $messages "matrixRTC.ingress.host is required when matrixRTC.enabled=true" }}
 {{- end }}
+{{- if and .sfu.exposedServices.turnTLS.enabled (not .sfu.exposedServices.turnTLS.tlsSecret) (not $root.Values.certManager) -}}
```
benbz, line 15: Needs to consider the cert-manager CRD not being in the cluster but `Values.certManager` being configured

charts/matrix-stack/source/common/exposedServicePort.json
```diff
@@ -2,13 +2,28 @@
   "type": "object",
   "required": [
     "enabled",
+    "type",
```
benbz: Same question for all 3 exposed services fragments - what's the difference between `type` and `portType`?

gaelgatelement: `type` is the service type, `portType` is the port type.

benbz: Can we `enum` the `portType` then?

benbz: This now exposes the confusion more clearly. We've got `NodePort` as valid values in both. We've also got other values in both.

The various `Services` are only created if `portType` is `NodePort`, but then if `type` is `ClusterIP` what does that mean?

If `portType` is `HostPort` the `Services` won't be created but what if you've set `type` to `NodePort`...

My feeling is that we want to do something like
* Single `portType` field
* `Service` always created
* `type: ClusterIP` if `portType: HostPort` ?
* `spec.ports[0].nodePort` only set if `NodePort` (or `LoadBalancer`?)

gaelgatelement: I'm wondering why would we create a Service in `HostPort` mode? The port is mapped on the pod directly no ?

gaelgatelement: I wonder if the confusion is in that I allowed `ClusterIP` for the exposed services, which does not make any sense actually?

charts/matrix-stack/templates/matrix-rtc/sfu_deployment.yaml
```diff
@@ -91,6 +91,16 @@ spec:
           name: http
           protocol: TCP
 {{- with .exposedServices }}
+  {{- with .turnTLS }}
+  {{- if .enabled }}
+        - containerPort: {{ .port }}
+    {{- if eq .type "HostPort" }}
```
```suggestion
    {{- if eq .portType "HostPort" }}
```

And everywhere else. Probably want a manifest test for this

charts/matrix-stack/templates/matrix-rtc/sfu_rtc_tcp_service.yaml
```diff
@@ -1,23 +1,27 @@
...
+Copyright 2025-2026 Element Creations Ltd
 
 SPDX-License-Identifier: AGPL-3.0-only
 */ -}}
 
 {{- with $.Values.matrixRTC -}}
 {{- if .enabled -}}
 {{- with .sfu -}}
-{{- if and .enabled .exposedServices.rtcTcp.enabled (eq .exposedServices.rtcTcp.portType "NodePort") -}}
+{{- if and .enabled .exposedServices.rtcTcp.enabled (ne .exposedServices.turnTLS.portType "HostPort") -}}
```
benbz:
```suggestion
{{- if and .enabled .exposedServices.rtcTcp.enabled (ne .exposedServices.rtcTcp.portType "HostPort") -}}
```

charts/matrix-stack/templates/matrix-rtc/sfu_rtc_udp_muxer_service.yaml
```diff
@@ -1,23 +1,27 @@
...
+Copyright 2025-2026 Element Creations Ltd
 
 SPDX-License-Identifier: AGPL-3.0-only
 */ -}}
 
 {{- with $.Values.matrixRTC -}}
 {{- if .enabled -}}
 {{- with .sfu -}}
-{{- if and .enabled .exposedServices.rtcMuxedUdp.enabled (eq .exposedServices.rtcMuxedUdp.portType "NodePort") -}}
+{{- if and .enabled .exposedServices.rtcMuxedUdp.enabled (ne .exposedServices.turnTLS.portType "HostPort") -}}
```
benbz:
```suggestion
{{- if and .enabled .exposedServices.rtcMuxedUdp.enabled (ne .exposedServices.rtcMuxedUdp.portType "HostPort") -}}
```

charts/matrix-stack/templates/matrix-rtc/sfu_rtc_udp_range_service.yaml
```diff
@@ -1,14 +1,14 @@
...
+Copyright 2025-2026 Element Creations Ltd
 
 SPDX-License-Identifier: AGPL-3.0-only
 */ -}}
 
 {{- with $.Values.matrixRTC -}}
 {{- if .enabled -}}
 {{- with .sfu -}}
-{{- if and .enabled .exposedServices.rtcUdp.enabled (eq .exposedServices.rtcUdp.portType "NodePort") -}}
+{{- if and .enabled .exposedServices.rtcUdp.enabled (ne .exposedServices.turnTLS.portType "HostPort") -}}
```
benbz:
```suggestion
{{- if and .enabled .exposedServices.rtcUdp.enabled (ne .exposedServices.rtcUdp.portType "HostPort") -}}
```

charts/matrix-stack/templates/matrix-rtc/sfu_rtc_udp_muxer_service.yaml
```diff
@@ -1,23 +1,27 @@
...
+Copyright 2025-2026 Element Creations Ltd
 
 SPDX-License-Identifier: AGPL-3.0-only
 */ -}}
 
 {{- with $.Values.matrixRTC -}}
 {{- if .enabled -}}
 {{- with .sfu -}}
-{{- if and .enabled .exposedServices.rtcMuxedUdp.enabled (eq .exposedServices.rtcMuxedUdp.portType "NodePort") -}}
+{{- if and .enabled .exposedServices.rtcMuxedUdp.enabled (ne .exposedServices.rtcUdp.portType "HostPort") -}}
```
benbz:
```suggestion
{{- if and .enabled .exposedServices.rtcMuxedUdp.enabled (ne .exposedServices.rtcMuxedUdp.portType "HostPort") -}}
```

tests/manifests/__init__.py
```diff
@@ -101,6 +102,7 @@ class DeployableDetails(abc.ABC):
 
     has_additional_config: bool = field(default=None, hash=False)  # type: ignore[assignment]
     has_db: bool = field(default=False, hash=False)
+    has_exposed_services: bool = field(default=True, hash=False)
```
benbz: Would it be better to `default=False` and then remove all the `values_file_path_overrides` and `has_exposed_services=False`?

tests/manifests/test_services.py
```diff
@@ -6,8 +6,58 @@
...
+                    for port in template["spec"]["ports"]:
+                        found_exposed_service[port["name"]] = True
+                if service_mode == "HostPort":
+                    assert len(found_exposed_service) == 0, (
+                        f"Found exposed service despite using HostPort : {found_exposed_service}"
+                    )
+                if template["kind"] == "Deployment" and "exposed-service-port-type" in template["metadata"]["labels"]:
+                    container = template["spec"]["template"]["spec"]["containers"][0]
+                    for port in container["ports"]:
+                        if service_mode != "HostPort":
```
benbz, line 78: Needs an assert if the `service_mode` is `HostPort`. I imagine the best we can do is that there is `len(exposed_services)` ports with `hostPort`?

tests/manifests/test_services.py
```diff
@@ -6,8 +6,58 @@
...
+                    assert (
+                        template["metadata"]["name"].lower()
+                        == template["metadata"]["annotations"]["exposed-service-name"]
+                    )
+                    for port in template["spec"]["ports"]:
+                        found_exposed_service[port["name"]] = True
+                if service_mode == "HostPort":
+                    assert len(found_exposed_service) == 0, (
+                        f"Found exposed service despite using HostPort : {found_exposed_service}"
+                    )
```
benbz: Needs an assert that if `service_mode` isn't `HostPort`, `len(found_exposed_service) == len(exposed_services)`

tests/manifests/__init__.py
gaelgatelement: At first I wanted to default test it. Since my test has to hardcode the CI Values file used to verify them (because I need to declared all verified exposed services in there) let's just do that yes.

tests/manifests/test_services.py
gaelgatelement: I'm only asserting that we found exposed services, as I detect exposed services using an annotation instead of the service type. So non-exposed services do not have the annotation.

tests/manifests/test_services.py
gaelgatelement: I'm now asserting that exposed services list is empty when HostPort is used.

tests/manifests/__init__.py
```diff
@@ -382,6 +385,7 @@ def make_synapse_worker_sub_component(worker_name: str, worker_type: str) -> Sub
         f"synapse-{worker_name}",
         values_file_path=ValuesFilePath.read_write("synapse", "workers", worker_name),
         values_file_path_overrides=values_file_path_overrides,
+        has_exposed_services=False,
```
benbz: All these can disappear

tests/manifests/test_services.py
benbz: But this is the `Deployment`. We want to assert there are the appropriate number of ports with `hostPort` here?

tests/manifests/test_services.py
```diff
@@ -1,13 +1,64 @@
...
+                    for port in container["ports"]:
+                        if service_mode != "HostPort":
+                            assert "hostPort" not in port, f"Found hostPort despite using {service_mode} : {port}"
+
+        if service_mode == "HostPort":
+            assert len(found_exposed_services) == 0, (
+                f"Found exposed service despite using HostPort : {found_exposed_services}"
+            )
+        else:
+            assert found_exposed_services, (
```
benbz: If `exposed_service_mode` saves off the number of `exposed_service` by `deployable_details.name` then we should be able to check we've got the correct number down here

charts/matrix-stack/source/common/exposedServicePort.json
gaelgatelement: We converged to using `portType` to determine how the service is exposed. It is either `NodePort`, `HostPort` or `LoadBalancer`.

tests/manifests/__init__.py
```diff
@@ -364,6 +366,7 @@ def make_synapse_worker_sub_component(worker_name: str, worker_type: str) -> Sub
     values_file_path_overrides: dict[PropertyType, ValuesFilePath] = {
         PropertyType.AdditionalConfig: ValuesFilePath.read_elsewhere("synapse", "additional"),
         PropertyType.Env: ValuesFilePath.read_elsewhere("synapse", "extraEnv"),
+        PropertyType.ExposedServices: ValuesFilePath.not_supported(),
```
benbz: Why's this needed for Synapse workers and nothing else?

PR #979, opened 2026-01-20
Add Hookshot

gaelgatelement:
- This adds support for dynamic registration file generation with matrix-tools
- This adds support for Hookshot setup in the chart

### DIFF
--- a/newsfragments/979.added.1.md
+++ b/newsfragments/979.added.1.md
@@ -1,1 +1,4 @@
 Add support for Hookshot installation in the chart.
+
+[Hookshot](https://matrix-org.github.io/matrix-hookshot/latest/hookshot.html) is a Matrix Bot for connecting to external services.
+It is not enabled by default, but can be enabled by setting `hookshot.enabled: true`.


charts/matrix-stack/templates/init-secrets/_helpers.tpl
```diff
@@ -82,6 +82,16 @@ app.kubernetes.io/version: {{ include "element-io.ess-library.labels.makeSafe" $
 {{- end }}
 {{- end }}
 {{- end }}
+{{- with $root.Values.hookshot }}
+{{- if .enabled }}
+{{- if not .appserviceRegistration }}
+- {{ (printf "%s-generated" $root.Release.Name) }}:HOOKSHOT_REGISTRATION:registration:/registration-templates/hookshot-registration.yaml
+{{- end }}
+{{- if not .passkey }}
+- {{ (printf "%s-generated" $root.Release.Name) }}:HOOKSHOT_RSA_PASSKEY:rsa:4096:pem
```
benbz, line 91: Does this need to be 4096 or should we do 2048?

newsfragments/979.added.1.md
```diff
@@ -0,0 +1 @@
+Add support for Hookshot installation in the chart.
```
benbz, line 1:
```suggestion
Add support for Hookshot installation in the chart.

[Hookshot](https://matrix-org.github.io/matrix-hookshot/latest/hookshot.html) is a Matrix Bot for connecting to external services.
It is not enabled by default, but can be enabled by setting `hookshot.enabled: true`.
```

charts/matrix-stack/templates/init-secrets/_helpers.tpl
gaelgatelement: Why would we want to do a less-secure RSA key ? :thinking: Is there some context I'm missing here?

gaelgatelement:
> Currently the only way to do webhooks other than the generic ones is to specify a custom registration file. And it has to be a complete registration file (with tokens) rather than a template. Is that correct?

Correct!

charts/matrix-stack/templates/init-secrets/_helpers.tpl
benbz: Slower to generate, not meaningfully more secure

gaelgatelement:
> > Currently the only way to do webhooks other than the generic ones is to specify a custom registration file. And it has to be a complete registration file (with tokens) rather than a template. Is that correct?
> 
> Correct!

I misread - Incorrect.

In practice, we can use all integrations without custom namespaces.
- alias namespaces are only used for dynamic rooms, which HS does not recommend supporting anymore : https://matrix-org.github.io/matrix-hookshot/latest/usage/dynamic_rooms.html
- users namespaces are only required if you want a per-integration dedicated user. With current registration file, you can still use any integration, the user will always be the same single hookshot user.

PR #985, opened 2026-01-22
Add tests to verify service types

gaelgatelement:
This : 

 - Adds test to make sure that we could configured service types behind ingresses
 - Make sure service type is specified with a valid default value.

### DIFF
--- a/charts/matrix-stack/templates/hookshot/hookshot_service.yaml
+++ b/charts/matrix-stack/templates/hookshot/hookshot_service.yaml
@@ -15,8 +15,12 @@ metadata:
   labels:
     {{- include "element-io.hookshot.labels" (dict "root" $ "context" .) | nindent 4 }}
 spec:
-  type: {{ .ingress.service.type | default $.Values.ingress.service.type }}
+{{ with .ingress.service.type | default $.Values.ingress.service.type }}
+  type: {{ . }}
+  {{ if eq . "ClusterIP" }}
   clusterIP: None
+  {{- end }}
+{{- end }}
   ipFamilyPolicy: PreferDualStack
   ports:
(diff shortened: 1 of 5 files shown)


charts/matrix-stack/templates/hookshot/hookshot_service.yaml
```diff
@@ -15,6 +15,7 @@ metadata:
   labels:
     {{- include "element-io.hookshot.labels" (dict "root" $ "context" .) | nindent 4 }}
 spec:
+  type: {{ .ingress.service.type | default $.Values.ingress.service.type }}
   clusterIP: None
```
benbz, line 21: Do we have to do a little dance to omit this line if `type` resolves to something that isn't `ClusterIP`?

gaelgatelement:
> I'm wondering if we need more tests (low-priority, separate PR)
> 

>     Service for StatefulSet defaults to type: ClusterIP and clusterIP: None

This exists as https://github.com/element-hq/ess-helm/blob/main/tests/manifests/test_statefulset.py#L13

>     * All other `Services` don't have `clusterIP: None`

Could happen in another PR I suppose, but this is low prio indeed.

>     * `clusterIP: None` is never present if `type != ClusterIP`

I'll add that one

benbz:
> > I'm wondering if we need more tests (low-priority, separate PR)
> 
> > ```
> > Service for StatefulSet defaults to type: ClusterIP and clusterIP: None
> > ```
> 
> This exists as https://github.com/element-hq/ess-helm/blob/main/tests/manifests/test_statefulset.py#L13

It didn't catch the issue in `charts/matrix-stack/templates/synapse/synapse_service.yaml` though?

PR #988, opened 2026-01-22
hookshot: document how to set it up

gaelgatelement:


docs/advanced.md
```diff
@@ -109,6 +110,40 @@ elementWeb:
...
+To configure Hookshot, create a values file with the YAML config to inject as keys that will be merged to Hookshot under “additional”:
+
+```yml
+hookshot:
+  additional:
+    user-config.yaml:
+      config: |
+        generic:
+          enabled: true
+          outbound: true # For outbound webhook support
```
benbz: Is this really something people should be setting up by default?

PR #989, opened 2026-01-22
matrix rtc: fix an issue when using cert-manager

gaelgatelement:
The code path was not tested at all, resulting in multiple manifest syntax error.


charts/matrix-stack/templates/matrix-rtc/sfu_turn_tls_certificate.yaml
```diff
@@ -15,14 +15,22 @@ kind: Certificate
...
+    {{- include "element-io.matrix-rtc-sfu.labels" (dict "root" $ "context" (mustMergeOverwrite (dict "suffix" "-turn-tls") .)) | nindent 4 }}
 spec:
   commonName: {{ tpl .exposedServices.turnTLS.domain $ }}
-  secretName: "{{ .Release.Name }}-turn-tls-certmanager-tls"
+  secretName: "{{ $.Release.Name }}-turn-tls-certmanager-tls"
   dnsNames:
   - {{ .exposedServices.turnTLS.domain }}
   issuerRef:
-    name: {{ $.Values.certManager.clusterIssuer }}
+{{- with $.Values.certManager.clusterIssuer }}
```
benbz, line 26: Can we have schema validation that only 1 of these is set?

charts/matrix-stack/ci/matrix-rtc-exposed-services-tls-values.yaml
benbz, line 1: Missing the cert-manager properties?

gaelgatelement: There is now 2 values, one with cert-manager enabled in `ci_extra` and one with a static tls secret in `ci`.

PR #991, opened 2026-01-23
hookshot: add support for github private key configuration

gaelgatelement:
Github configuration requires a dedicated private key pem file. We need to inject it through the chart values.


gaelgatelement:
We are going to rely on `extraVolumes` and `extraVolumeMounts` instead to do this.

Issue #922, opened 2025-12-04
Support internalTrafficPolicy in synapse.ingress.service configuration


watermelon0339:
**Is your feature request related to a problem? Please describe.**
Currently, the `synapse.ingress.service` configuration only exposes the `synapse.ingress.service.type` option. It would be very helpful to also support `internalTrafficPolicy` as part of this configuration, for example: `synapse.ingress.service.internalTrafficPolicy`.

Adding this option would provide more flexibility in controlling service traffic routing, especially for clusters where internal traffic handling policies are important for performance and reliability.

**Describe the solution you'd like**

https://github.com/element-hq/ess-helm/blob/main/charts/matrix-stack/templates/synapse/synapse_http_service.yaml

```yaml 
{{- with $.Values.synapse -}}
{{- if .enabled -}}
apiVersion: v1
kind: Service
metadata:
  labels:
    {{- include "element-io.synapse-ingress.labels" (dict "root" $ "context" $.Values.haproxy) | nindent 4 }}
  name: {{ $.Release.Name }}-synapse
  namespace: {{ $.Release.Namespace }}
spec:
  type: {{ .ingress.service.type | default $.Values.ingress.service.type }}
  internalTrafficPolicy: {{ .ingress.service.internalTrafficPolicy | default "Cluster" }} # 👈👈 Here
  ipFamilyPolicy: PreferDualStack
  ports:
  - name: haproxy-synapse
    port: 8008
    targetPort: haproxy-synapse
  - name: haproxy-403
    port: 8009
    targetPort: haproxy-403
  selector:
    app.kubernetes.io/instance: "{{ $.Release.Name }}-haproxy"
{{- end -}}
{{- end -}}
```

**Describe alternatives you've considered**
```bash
helm upgrade --install  xxxx
kubectl patch service ess-synapse -n ess -p '{"spec":{"internalTrafficPolicy":"Local"}}'
```

benbz:
Thanks for the request. We're not opposed to this but it isn't a high priority for us to implement ourselves unless there's requests from paying customers.

We'd happily accept a community PR along the lines of
* Per-`Ingress` `service.internalTrafficPolicy`
* Defaults to global-/common-`Ingress` `service.internalTrafficPolicy` if omitted
* Per-`Ingress` setting is omitted by default
* Global-/Common-`Ingress` setting is `Cluster` by default
* Appropriate manifest tests around this

[...]

PR #999, opened 2026-01-26
Add support for `internalTrafficPolicy` configuration for ingress services

gaelgatelement:
- Ingresses services now support setting `internalTrafficPolicy`
- It defaults to `Cluster` using the root value `$.Values.services.internalTrafficPolicy`
- It can be adjusted per-service like service types

Resolves https://github.com/element-hq/ess-helm/issues/922

### DIFF
--- a/tests/manifests/test_ingresses.py
+++ b/tests/manifests/test_ingresses.py
@@ -306,8 +306,9 @@ async def test_component_ingressClassName_beats_global(values, make_templates):
 
 @pytest.mark.parametrize("values_file", values_files_to_test)
 @pytest.mark.asyncio_cooperative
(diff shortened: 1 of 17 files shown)


tests/manifests/test_ingresses.py
benbz, line 1: What's here LGTM but is there nowhere we assert the default `type` and (now) `internalTrafficPolicy`?

gaelgatelement: The defaults are the global, so this is verified through `global_service_properties` test AFAICT?

benbz: Hmm I suppose that's fine, in that it asserts the `Service` respects the `global`, even if we don't assert what the default is

PR #1000, opened 2026-01-26
Support configuring externalTrafficPolicy for ingress services

gaelgatelement:
- This allows setting `externalTrafficPolicy` for ingress services of type `NodePort` and `LoadBalancer`
- The default value is configured by the root property `$.Values.ingress.type`

### DIFF
--- a/tests/manifests/test_services.py
+++ b/tests/manifests/test_services.py
@@ -243,15 +243,14 @@ async def test_services_specify_type_by_default(templates):
             assert template["spec"].get("type") in ("ClusterIP", "NodePort"), (
                 f"{template_id(template)} service type is not ClusterIP"
             )
-            if template["spec"]["type"] != "ClusterIP":
-                assert "clusterIP" not in template["spec"], (
-                    f"{template_id(template)} has a clusterIP defined for a non-ClusterIP service"
-                )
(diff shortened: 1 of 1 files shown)


tests/manifests/test_services.py
```diff
@@ -247,3 +247,12 @@ async def test_services_specify_type_by_default(templates):
                 assert "clusterIP" not in template["spec"], (
                     f"{template_id(template)} has a clusterIP defined for a non-ClusterIP service"
                 )
+            if template["spec"]["type"] == "ClusterIP":
```
benbz, line 246:
```suggestion
            else:
```

tests/manifests/test_services.py
```diff
@@ -247,3 +247,12 @@ async def test_services_specify_type_by_default(templates):
...
+                assert "externalTrafficPolicy" not in template["spec"], (
+                    f"{template_id(template)} has an externalTrafficPolicy defined for a ClusterIP service"
+                )
+            else:
+                assert "externalTrafficPolicy" in template["spec"], (
+                    f"{template_id(template)} does not specify an externalTrafficPolicy "
+                    "despite being a NodePort service"
+                )
```
benbz, line 257: Move this up into the `if template["spec"]["type"] != "ClusterIP"`

PR #1001, opened 2026-01-26
Add support for configuring traffic policy of exposed services

gaelgatelement:
- This changes the default of the SFU exposed services to `Cluster` to stay consistent with kubernetes defaults

### DIFF
--- a/tests/manifests/test_services.py
+++ b/tests/manifests/test_services.py
@@ -3,6 +3,8 @@
 #
 # SPDX-License-Identifier: AGPL-3.0-only
 
+import itertools
+
 import pytest
 import yaml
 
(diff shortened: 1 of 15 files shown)


tests/manifests/test_services.py
```diff
@@ -174,6 +176,61 @@ def exposed_service_certificate(deployable_details: DeployableDetails, issuer_na
                 )
 
 
+@pytest.mark.parametrize("values_file", services_values_files_to_test)
+@pytest.mark.asyncio_cooperative
+async def test_exposed_services_traffic_policy(values, make_templates):
+    num_of_expected_ports = {}
+    num_of_expected_exposed_services = {}
```
benbz: No assertion that these counts are reached

PR #1006, opened 2026-01-27
Add support for external IPs in exposed services

gaelgatelement:


charts/matrix-stack/templates/matrix-rtc/sfu_rtc_tcp_service.yaml
```diff
@@ -23,6 +23,9 @@ metadata:
 spec:
   type: {{ .exposedServices.rtcTcp.portType }}
   ipFamilyPolicy: PreferDualStack
+  {{- if hasKey .exposedServices.rtcTcp "externalIPs" }}
```
benbz, line 26: For my benefit this is done with `if` and not `with` so that it puts `[]` into the manifest if that's configured?

tests/manifests/test_services.py
```diff
@@ -121,6 +122,65 @@ def exposed_service_mode(deployable_details: DeployableDetails, mode: str):
...
+        for service_name in exposed_services:
+            test_annotations_labels = {
+                "exposed-service-external-ip": str(external_ip),
+            }
+            exposed_services_values[service_name]["annotations"] = test_annotations_labels
+            exposed_services_values[service_name]["externalIPs"] = ("127.0.0.1", str(external_ip))
+            external_ip += 1
+            num_of_expected_exposed_services[deployable_details.name] += 1
+        deployable_details.set_helm_values(values, PropertyType.ExposedServices, exposed_services_values)
+        deployable_details.set_helm_values(values, PropertyType.Labels, test_annotations_labels)
```
benbz: Why is this set? We don't appear to read labels again?

tests/manifests/test_services.py
```diff
@@ -121,6 +122,65 @@ def exposed_service_mode(deployable_details: DeployableDetails, mode: str):
                 ), f"Did not find expected exposed service despite using {service_mode}. Services: {found_services}"
 
 
+@pytest.mark.parametrize("values_file", services_values_files_to_test)
+@pytest.mark.asyncio_cooperative
+async def test_exposed_services_external_ips(values, make_templates):
+    # we set deployables external ips on 10.0.X.1
+    # and each exposed services increments the last digit of the ip
+    expected_deployable_external_ips = {}
```
benbz, line 130: This is set but never read AFAICS and the IP set has no relationship to what's set in `exposed_external_ip`. Remove?

charts/matrix-stack/templates/matrix-rtc/sfu_rtc_tcp_service.yaml
gaelgatelement: Yes I figured one might want to explicitly set an empty array

tests/manifests/test_services.py
gaelgatelement: Urg the algo was totally broken :D Thanks for catching !

tests/manifests/test_services.py
gaelgatelement: Unused indeed

tests/manifests/test_services.py
```diff
@@ -121,6 +122,66 @@ def exposed_service_mode(deployable_details: DeployableDetails, mode: str):
...
+    num_of_expected_exposed_services = {}
+
+    def exposed_external_ip(deployable_details: DeployableDetails, external_ip: ipaddress.IPv4Address):
+        nonlocal num_of_expected_exposed_services
+        exposed_services = deployable_details.get_helm_values(values, PropertyType.ExposedServices)
+        assert exposed_services is not None
+        exposed_services_values = dict(exposed_services)
+        for service_name in exposed_services:
+            test_annotations_labels = {
```
benbz:
```suggestion
            test_labels = {
```

Nit

PR #1010, opened 2026-01-28
Hookshot widget : 404 under synapse ingress

gaelgatelement:
This was detected during internal testing.

This (probably) requires a Hookshot fix.


gaelgatelement:
:heavy_check_mark:

PR #1016, opened 2026-01-30
Setup test env with pytest only

gaelgatelement:
This removes the long standing duplication that has been between bash & pytest.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -1,8 +1,18 @@
 # Copyright 2024 New Vector Ltd
-# Copyright 2025 Element Creations Ltd
+# Copyright 2025-2026 Element Creations Ltd
 #
 # SPDX-License-Identifier: AGPL-3.0-only
+import pytest
 
 pytest_plugins = [
     "integration.fixtures",
 ]
+
+
+# this overrides the pytest_kubernetes autouse teardown fixture
+# to make it compatible with asyncio_cooperative by making it an async fixture
+# In theory it would be used to teardown cached clusters, but we do not use this feature
+# in our pytest test suite. Our `cluster` fixture takes care of the teardown itself.
+@pytest.fixture(scope="session", autouse=True)
+async def remaining_clusters_teardown():
+    return
--- a/newsfragments/1016.internal.md
+++ b/newsfragments/1016.internal.md
@@ -1,0 +1,1 @@
+CI: Remove code duplication that existed between `pytest` integration test suite and `setup_test_cluster.sh` script.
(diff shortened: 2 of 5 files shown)


tests/integration/fixtures/cluster.py
```diff
@@ -129,6 +148,86 @@ async def ingress(cluster, kube_client):
...
+        "cert-manager",
+        chart,
+        yaml.safe_load((Path(__file__).parent / "files/charts/cert-manager.yml").open()),
+        namespace="cert-manager",
+        create_namespace=True,
+        atomic=True,
+        wait=True,
+    )
+
+    ca_folder = Path().resolve() / ".ca"
```
benbz: Where is this path anchored?

tests/integration/fixtures/cluster.py
```diff
@@ -129,6 +148,86 @@ async def ingress(cluster, kube_client):
...
+                    "tls.key": b64encode(ca_pem_path.read_text()),
+                    "ca.crt": b64encode(ca_crt_path.read_text()),
+                },
+            )
+        )
+    await kube_client.apply(
+        ClusterIssuer(metadata={"name": "ess-selfsigned"}, spec={"ca": {"secretName": "ess-ca"}}),
+        field_manager="pytest",
+    )
+    ess_ca_secret = await kube_client.get(Secret, name="ess-ca", namespace="cert-manager")
```
benbz, line 224: These were only written previously if they didn't both exist, can we maintain that behaviour

tests/integration/test_setup.py
benbz, line 1: What's this for?

gaelgatelement: It's to have at least 1 test which runs, else the fixtures are all skipped.

tests/integration/fixtures/cluster.py
gaelgatelement: It depends from where `pytest` is called. I guess we better use `__file__` special variable instead, it should be more reliable ?

tests/integration/conftest.py
```diff
@@ -1,8 +1,16 @@
...
 # SPDX-License-Identifier: AGPL-3.0-only
+import pytest
 
 pytest_plugins = [
     "integration.fixtures",
 ]
+
+
+# this overrides the pytest_kubernetes autouse teardown fixture
+# to make it compatible with asyncio_cooperative
```
benbz:
```suggestion
# to make it compatible with asyncio_cooperative by skipping the cluster deletion
```

i.e. https://github.com/Blueshoe/pytest-kubernetes/blob/0.7.2/pytest_kubernetes/plugin.py#L52-L56 is the thing you're nuking out

PR #1018, opened 2026-02-02
hookshot: revert internal routing

gaelgatelement:
- This is only used by Widgets
- This requires users to configure an allow ip list to work even if their cluster support harpinning to the front door
- In pytest, we will use the `openidOverride` feature to make it work without hostAliases. It can also be an example about how to make it work in practice.

### DIFF
--- a/charts/matrix-stack/ci/fragments/hookshot-pytest-extras.yaml
+++ b/charts/matrix-stack/ci/fragments/hookshot-pytest-extras.yaml
@@ -23,6 +23,8 @@ hookshot:
           enableHttpGet: false
           allowedIpRanges:
           - 10.0.0.0/8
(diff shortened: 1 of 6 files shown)


benbz:
I think `charts/matrix-stack/ci/test-cluster-mixin.yaml` needs updating to either have `openIdOverrides` or remove `allowedIpRanges` and do `hostAliases` + disable TLS env vars

benbz:
> I think `charts/matrix-stack/ci/test-cluster-mixin.yaml` needs updating to either have `openIdOverrides` or remove `allowedIpRanges` and do `hostAliases` + disable TLS env vars

Done with https://github.com/element-hq/ess-helm/pull/1023

PR #1021, opened 2026-02-04
ci: cache asdf tool setup

gaelgatelement:
There are some failure when k3d pull are throttled in CI. Let's try to cache those.


gaelgatelement:
Done in #1075

PR #1026, opened 2026-02-06
Add strict schema validation test for MAS config

sandhose:
This makes sure that all the MAS config we render with the test values
match the upstream JSON schema.

- Add jsonschema and types-jsonschema as dev dependencies
- Fetch the upstream MAS config JSON schema matching the chart's image tag
- Strictify the schema by recursively adding additionalProperties: false
  to all object sub-schemas (under both $defs and definitions)
- Merge config underrides and overrides per ConfigMap, then validate
- Parametrize over all CI values files, skipping when MAS is not enabled

Mostly done by Claude Opus

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -185,7 +185,7 @@ jobs:
       uses: actions/cache@cdf6c1fa76f9f475f3d7449005a359c84ca0f306  # v5
       with:
         path: .pytest_cache
-        key: pytest-manifests-cache-${{ matrix.manifest-test }}
+        key: pytest-manifests-cache-${{ matrix.manifest-test }}-${{ hashFiles('charts/matrix-stack/values.yaml') }}
         restore-keys: |
           pytest-manifests-cache-
 


tests/manifests/test_mas_config_schema.py
```diff
@@ -0,0 +1,131 @@
...
+
+@pytest.fixture(scope="session")
+async def strict_mas_schema(pytestconfig: pytest.Config, base_values: dict[str, Any]) -> dict[str, Any]:
+    tag = base_values["matrixAuthenticationService"]["image"]["tag"]
+    cacheable = False
+    if tag.startswith("sha-"):
+        # sha-xxyyzz → commit ref xxyyzz
+        ref = tag.removeprefix("sha-")
+        cacheable = True
+    elif tag[0].isdigit():
```
gaelgatelement:
```suggestion
    elif semver.Version.is_valid(tag):
```

tests/manifests/test_mas_config_schema.py
```diff
@@ -0,0 +1,131 @@
...
+        url = (
+            f"https://raw.githubusercontent.com/element-hq/matrix-authentication-service/{ref}/docs/config.schema.json"
+        )
+        try:
+            timeout = aiohttp.ClientTimeout(total=30)
+            async with aiohttp.ClientSession() as session, session.get(url, timeout=timeout) as resp:
+                resp.raise_for_status()
+                raw = await resp.read()
+        except (aiohttp.ClientError, TimeoutError):
+            pytest.skip(f"Could not fetch MAS config schema from {url}")
```
gaelgatelement: I would have expected a failure

sandhose: Hmm fair, although is for cases I suppose where we can't reliably map the MAS image tag back to a git ref (for a `pr-*` tag, the `latest` tag), so I don't want the test to start failing if we're in those edge cases?

sandhose: Reduced it to only fail on 404, wdyt? 5e53de8d

.github/workflows/pytest.yml
```diff
@@ -181,5 +181,13 @@ jobs:
         tool_versions: |
           helm 3.19.4
 
+    - name: Restore pytest cache
+      uses: actions/cache@cdf6c1fa76f9f475f3d7449005a359c84ca0f306  # v5
+      with:
+        path: .pytest_cache
+        key: pytest-manifests-cache-${{ matrix.manifest-test }}
```
gaelgatelement:
```suggestion
        key: pytest-manifests-cache-${{ matrix.manifest-test }}-${{ hashFiles('charts/matrix-stack/values.yaml') }}
```

we want to reset the cache if MAS version change, and this file contains it.

tests/manifests/test_mas_config_schema.py
gaelgatelement: Mmmh I'm not really comfortable with that either :thinking:  If the file moves on MAS repo, it will 404 and this test will permanently skip until we discover it.

I see 2 potential solutions : 
 - Easy but not great : Can you fallback to main if it 404?
 - Might be overengineered, but can we fetch the docker metadata of the image for the git source annotation?

gaelgatelement: I did the "overgineering" solution. Let me know what you think about it.

PR #1050, opened 2026-02-18
Add Gateway/HTTPRoute alternative ingress

JoshuaHassler:
This adds support for the Kubernetes v1 Gateway spec as an alternative to the traditional Ingress resources. This logic supports using Ingresses as they are now, using HTTPRoutes only, or a mix. HTTPRoutes can be connected to an existing gateway, or a new gateway can be generated for the user. I tried to unify the template logic and reuse existing code wherever possible. Currently using this in production on my home server, and I have run what tests I could figure out how to locally :slightly_smiling_face: 

A few key notes:
- The change does reorganize a few keys under the "ingress" config and is therefore technically a "breaking" change. I could restore these without affecting the HTTPRoute logic. Edit: I see you have deprication logic already so that is probably the way to handle any moving keys
- defaults to ingress resources
- Added an option to disable ingress altogether since I was writing the conditional logic anyway. This can be removed, but it is a common feature on Helm charts as an escape hatch for complex deployments.
- I have intentionally not handled routing SFU via TCP/UDP Routes in this PR. I can add them, but they have mixed support within gateway controllers, and this PR was already getting long as is.

As an outside contributor, how should I handle the copyright notices? I don't see a CLA, but all the files are currently copyrighted by Element.
I am not sure what test fragments you may want me to add, since I intentionally tried to make it pass without changing the minimum config.

Thanks for all the work making ess-comunity possible, and I look forward to your review!

### DIFF
--- a/charts/matrix-stack/source/common/routes_global.json
+++ b/charts/matrix-stack/source/common/routes_global.json
@@ -1,26 +1,35 @@
 {
   "type": "object",
   "properties": {
+    "existingGateways": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "required": [
+          "name"
+        ],
+        "properties": {
+          "name": { "type": "string" },
+          "namespace": { "type": "string" },
+          "group": { "type": "string" },
+          "kind": { "type": "string" },
+          "sectionName": { "type": "string" },
+          "port": { "type": "integer" }
+        }
+      },
+      "minItems": 1
+    },
     "annotations": {
       "type": "object",
       "additionalProperties": {
         "type": "string"
       }
     },
-    "className": {
-      "type": "string"
-    },
     "tlsEnabled": {
       "type": "boolean"
     },
     "tlsSecret": {
       "type": "string"
-    },
-    "controllerType": {
-      "type": "string",
-      "enum": [
-        "ingress-nginx"
-      ]
     },
     "service": {
       "type": "object",
--- a/charts/matrix-stack/source/common/routes_global.json
+++ b/charts/matrix-stack/source/common/routes_global.json
@@ -31,6 +31,9 @@
     "tlsSecret": {
       "type": "string"
     },
+    "disableHTTPRedirect": {
+      "type": "boolean"
+    },
     "service": {
       "type": "object",
       "required": [
--- a/charts/matrix-stack/source/common/routes_global.json
+++ b/charts/matrix-stack/source/common/routes_global.json
@@ -9,12 +9,24 @@
           "name"
         ],
         "properties": {
-          "name": { "type": "string" },
-          "namespace": { "type": "string" },
-          "group": { "type": "string" },
-          "kind": { "type": "string" },
-          "sectionName": { "type": "string" },
-          "port": { "type": "integer" }
+          "name": {
+            "type": "string"
+          },
+          "namespace": {
+            "type": "string"
+          },
+          "group": {
+            "type": "string"
+          },
+          "kind": {
+            "type": "string"
+          },
+          "sectionName": {
+            "type": "string"
+          },
+          "port": {
+            "type": "integer"
+          }
         }
       },
       "minItems": 1
--- a/charts/matrix-stack/source/element-admin.yaml.j2
+++ b/charts/matrix-stack/source/element-admin.yaml.j2
@@ -14,6 +14,7 @@ enabled: true
 replicas: 1
 {{- sub_schema_values.image(registry='oci.element.io', repository='element-admin', tag='0.1.10') -}}
 {{- sub_schema_values.ingress() -}}
+{{- sub_schema_values.routes() -}}
 {{- sub_schema_values.labels() -}}
 {{- sub_schema_values.workloadAnnotations() -}}
 {{- sub_schema_values.extraEnv() -}}
--- a/charts/matrix-stack/source/element-admin.yaml.j2
+++ b/charts/matrix-stack/source/element-admin.yaml.j2
@@ -14,6 +14,7 @@ enabled: true
 replicas: 1
 {{- sub_schema_values.image(registry='oci.element.io', repository='element-admin', tag='0.1.12') -}}
 {{- sub_schema_values.ingress() -}}
+{{- sub_schema_values.routes() -}}
 {{- sub_schema_values.labels() -}}
 {{- sub_schema_values.workloadAnnotations() -}}
 {{- sub_schema_values.extraEnv() -}}
--- a/charts/matrix-stack/source/values.schema.json
+++ b/charts/matrix-stack/source/values.schema.json
@@ -101,6 +101,31 @@
         }
       }
     },
+    "gateway": {
+      "type": "object",
+      "properties": {
+        "create": {
+          "type": "boolean"
+        },
+        "className": {
+          "type": "string"
+        },
+        "annotations": {
+          "type": "object",
+          "additionalProperties": {
+            "type": "string"
+          }
+        }
+      }
+    },
+    "inboundTrafficHandler": {
+      "type": "string",
+      "enum": [
+        "ingress",
+        "routes",
+        "none"
+      ]
+    },
     "deploymentMarkers": {
       "$ref": "file://deployment-markers.json"
     },
--- a/charts/matrix-stack/source/values.schema.json
+++ b/charts/matrix-stack/source/values.schema.json
@@ -79,6 +79,9 @@
     "ingress": {
       "$ref": "file://common/ingress_global.json"
     },
+    "routes": {
+      "$ref": "file://common/routes_global.json"
+    },
     "tolerations": {
       "$ref": "file://common/tolerations.json"
     },
--- a/charts/matrix-stack/source/values.schema.json
+++ b/charts/matrix-stack/source/values.schema.json
@@ -133,6 +133,12 @@
           "additionalProperties": {
             "type": "string"
           }
+        },
+        "listenFor": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
         }
       }
     },
--- a/charts/matrix-stack/source/values.schema.json
+++ b/charts/matrix-stack/source/values.schema.json
@@ -116,6 +116,31 @@
         }
       }
     },
+    "gateway": {
+      "type": "object",
+      "properties": {
+        "create": {
+          "type": "boolean"
+        },
+        "className": {
+          "type": "string"
+        },
+        "annotations": {
+          "type": "object",
+          "additionalProperties": {
+            "type": "string"
+          }
+        }
+      }
+    },
+    "inboundTrafficHandler": {
+      "type": "string",
+      "enum": [
+        "ingress",
+        "routes",
+        "none"
+      ]
+    },
     "deploymentMarkers": {
       "$ref": "file://deployment-markers.json"
     },
--- a/charts/matrix-stack/source/values.schema.json
+++ b/charts/matrix-stack/source/values.schema.json
@@ -82,6 +82,9 @@
     "storage": {
       "$ref": "file://common/storage_global.json"
     },
+    "routes": {
+      "$ref": "file://common/routes_global.json"
+    },
     "tolerations": {
       "$ref": "file://common/tolerations.json"
     },
--- a/charts/matrix-stack/templates/element-admin/ingress.yaml
+++ b/charts/matrix-stack/templates/element-admin/ingress.yaml
@@ -5,7 +5,7 @@ Copyright 2025 Element Creations Ltd
 SPDX-License-Identifier: AGPL-3.0-only
 */ -}}
 {{- with .Values.elementAdmin -}}
-{{- if .enabled -}}
+{{- if eq (include "element-io.ess-library.ingress.isEnabled" (dict "root" $ "context" (dict "ingress" .ingress "type" "Ingress"))) "true" -}}
 apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
@@ -16,7 +16,7 @@ metadata:
   namespace: {{ $.Release.Namespace }}
 spec:
 {{- include "element-io.ess-library.ingress.tls" (dict "root" $ "context" (dict "ingress" .ingress "ingressName" "element-admin")) | nindent 2 }}
-{{- include "element-io.ess-library.ingress.className" (dict "root" $ "context" .ingress.className) | nindent 2 }}
+{{- include "element-io.ess-library.ingress.className" (dict "root" $ "context" .ingress.Ingress.className) | nindent 2 }}
   rules:
   - host: {{ (tpl .ingress.host $) | quote }}
     http:
--- a/charts/matrix-stack/templates/element-admin/ingress.yaml
+++ b/charts/matrix-stack/templates/element-admin/ingress.yaml
@@ -5,7 +5,7 @@ Copyright 2025 Element Creations Ltd
 SPDX-License-Identifier: AGPL-3.0-only
 */ -}}
 {{- with .Values.elementAdmin -}}
-{{- if eq (include "element-io.ess-library.ingress.isEnabled" (dict "root" $ "context" (dict "ingress" .ingress "type" "Ingress"))) "true" -}}
+{{- if eq (include "element-io.ess-library.inboundTrafficHandler.isEnabled" (dict "root" $ "context" (dict "component" . "trafficHandler" "ingress"))) "true" -}}
 apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
@@ -16,7 +16,7 @@ metadata:
   namespace: {{ $.Release.Namespace }}
 spec:
 {{- include "element-io.ess-library.ingress.tls" (dict "root" $ "context" (dict "ingress" .ingress "ingressName" "element-admin")) | nindent 2 }}
-{{- include "element-io.ess-library.ingress.className" (dict "root" $ "context" .ingress.Ingress.className) | nindent 2 }}
+{{- include "element-io.ess-library.ingress.className" (dict "root" $ "context" .ingress.className) | nindent 2 }}
   rules:
   - host: {{ (tpl .ingress.host $) | quote }}
     http:
--- a/charts/matrix-stack/templates/ess-library/_inboundTrafficHandler.tpl
+++ b/charts/matrix-stack/templates/ess-library/_inboundTrafficHandler.tpl
@@ -35,6 +35,7 @@ SPDX-License-Identifier: AGPL-3.0-only
 {{- $serviceName := required "element-io.ess-library.httpRedirectRoute context missing ServiceName" .serviceName -}}
 {{- $component := required "element-io.ess-library.httpRedirectRoute context missing Component" .component -}}
 {{- $host := required "element-io.ess-library.httpRedirectRoute context missing Host" .host -}}
+{{- $labels := required "element-io.ess-library.httpRedirectRoute context missing Labels" .labels -}}
 {{- $path := .path | default "/" -}}
 ---
 apiVersion: gateway.networking.k8s.io/v1
@@ -42,7 +43,7 @@ kind: HTTPRoute
 metadata:
 {{- include "element-io.ess-library.ingress.annotations" (dict "root" $root "context" (dict "ingress" $component.routes)) | nindent 2 }}
   labels:
-    {{- include "element-io.element-admin.labels" (dict "root" $root "context" $component) | nindent 4 }}
+    {{- $labels | nindent 4 }}
   name: {{ printf "%s-%s-redirect" $root.Release.Name $serviceName }}
   namespace: {{ $root.Release.Namespace }}
 spec:
--- a/charts/matrix-stack/templates/ess-library/_inboundTrafficHandler.tpl
+++ b/charts/matrix-stack/templates/ess-library/_inboundTrafficHandler.tpl
@@ -34,9 +34,11 @@ SPDX-License-Identifier: AGPL-3.0-only
 {{- with required "element-io.ess-library.httpRouteRedirect missing context" .context -}}
 {{- $serviceName := required "element-io.ess-library.httpRedirectRoute context missing ServiceName" .serviceName -}}
 {{- $component := required "element-io.ess-library.httpRedirectRoute context missing Component" .component -}}
+{{- $disabled := coalesce $component.disableHTTPRedirect $root.Values.routes.disableHTTPRedirect false -}}
 {{- $host := required "element-io.ess-library.httpRedirectRoute context missing Host" .host -}}
 {{- $labels := required "element-io.ess-library.httpRedirectRoute context missing Labels" .labels -}}
 {{- $path := .path | default "/" -}}
+{{- if not $disabled -}}
 ---
 apiVersion: gateway.networking.k8s.io/v1
 kind: HTTPRoute
@@ -61,5 +63,6 @@ spec:
           scheme: https
           statusCode: 301
         type: RequestRedirect
+{{- end -}}
 {{- end -}}
 {{- end -}}
--- a/charts/matrix-stack/templates/ess-library/_inboundTrafficHandler.tpl
+++ b/charts/matrix-stack/templates/ess-library/_inboundTrafficHandler.tpl
@@ -5,6 +5,30 @@ Copyright 2025-2026 Element Creations Ltd
 SPDX-License-Identifier: AGPL-3.0-only
 */ -}}
 
+{{- define "element-io.ess-library.inboundTrafficHandler.host" -}}
+{{- $root := .root -}}
+{{- with required "element-io.ess-library.inboundTrafficHandler.host missing context" .context -}}
+{{- $component := required "element-io.ess-library.inboundTrafficHandler.host missing component" .component -}}
+{{- $handler := coalesce $component.inboundTrafficHandler $root.Values.inboundTrafficHandler "none" -}}
+{{- if eq $handler "ingress" -}}
+{{- tpl (dig "ingress" "host" "" $component) $root -}}
+{{- else if eq $handler "routes" -}}
+{{- tpl (dig "routes" "host" "" $component) $root -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
+
+{{- define "element-io.ess-library.inboundTrafficHandler.name" -}}
+{{- $root := .root -}}
+{{- with required "element-io.ess-library.inboundTrafficHandler.name missing context" .context -}}
+{{- $component := required "element-io.ess-library.inboundTrafficHandler.name missing component" .component -}}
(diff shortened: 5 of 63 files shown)


JoshuaHassler:
It seems the CLAassistant has answered one of my questions :smile:. I'll update the files.

benbz:
:wave: thanks for raising this PR. Sorry I meant to reply to this yesterday but it got put to one side and forgot to pick it back up.

Supporting the Gateway API is definitely something we would accept and would like in ESS Community. We would want to get the design & structure of the values correct given how fundamental this would be to the whole stack. I haven't looked at this PR in huge detail yet, but wanted to raise some initial queries. I don't necessarily know the correct answer yet to all of these but they're just things that came to mind on a first pass
* Is there a reason to put the Gateway values in with the Ingress values?
  * i.e. have `synapse.gateway.host` as well as `synapse.ingress.host` use the present/absence of `...host` to determine which to use?
  * Or pull `synapse.ingress.host` up to `synapse.host` and have your `type` value or  enabled flags under `synapse.{gateway,ingress}`
* When the chart creates its own `Gateway` are providing wildcard certs via a single value still possible?
  * I think they are due to the `element-io.gateway.tlsConfig` helper still using `coalesce` but want to confirm
  * Alternatively would your expectation be that if you need that setup you must provide an existing gateway?

Other observations
* We'd want `HTTPRoute` and `Gateway` doing capability checks that those resources available in the cluster
* There'd need to be similar tests to `tests/manifests/test_ingresses.py` for Gateway API resources
* Are there other charts you've been looking at for inspiration on how to structure the values / what functionality to provide?
  * We're not going to blindly do what other charts do, especially given we have the constraint that we're a multi-component chart where all components are optional. Conversely, we're not going to do our own thing for the sake of it, so if there's an emerging convention for the structure of Gateway API related values we're open to that

I'll try to look at this in more detail next week, especially as we discuss the questions I've asked above. Once again, thanks for raising this

JoshuaHassler:
Hey @benbz, no problem, you guys seem quite busy. I don't use GitHub for work, so I didn't know whether it would auto-add a reviewer.

To shed some initial light on your questions:
> Is there a reason to put the Gateway values in with the Ingress values?
>
>   - i.e. have synapse.gateway.host as well as synapse.ingress.host use the present/absence of ...host to determine which to use?
>   - Or pull synapse.ingress.host up to synapse.host and have your type value or enabled flags under synapse {gateway,ingress}

For the actual Gateway config, I think that it should be promoted from `ingress.gateway` to a top-level key (`gateway`). For most people using this API, they will probably only want a single gateway, so they aren't paying for a bunch of cloud LoadBalancers needlessly. Multi-gateway configs can be handled externally through the `parentRefs` as an "advanced" feature.

In terms of overall structure, I am not 100% sure what the best approach is, but this was my thought process.

- We have a set of config options that are shared between both ingress and httproute resources. To handle this we can either...
  - place shared keys in the parent component
    - This is clean for the "host" key, but would lead to things like "ingressAnnotations", etc
    - We would need an ingress-esq key top-level key anyway to group the global versions of these shared options
    - As a benifit we don't need to change the associated shared templates if other ingress types are added.
  - duplicate shared keys
    - Don't try to share keys between HTTPRoute and Ingress resources
    - Leads to more complicated templating that needs to be modified for each ingress type (some shared "core" template, then a wrapper to shim in the values from that specific config)
    - Potential code drift for shared keys
    - arguably a simpler config experience for the end user
    - works well with the existing global + local override convention
  - shared state + specific (this implementation)
    - works well with the existing global + local override convention
    - able to reuse the existing templates mostly
    - no drift on shared keys
    - IMO easier to add new ingress types in the future since you only have to create the helpers for that specific type.
    - I don't know how other people feel about specifying the "type" key

I don't know how much of this code base you share with Pro, but IMO the main benefit of the last approach is that it tends to require less refactoring when you add a new ingress type. For instance, if you have a customer request for Istio `VirtualService` or ambassador based ingresses, you don't need to modify the shared templates as much in the future. It would be mostly contained to updating `ess-community.ess-library.ingress.isEnabled`.
 
> When the chart creates its own Gateway are providing wildcard certs via a single value still possible?
> 
>     I think they are due to the element-io.gateway.tlsConfig helper still using coalesce but want to confirm
>     Alternatively would your expectation be that if you need that setup you must provide an existing gateway?
It should be. I currently use a wildcard certificate issued by Let's Encrypt. You just specify the secret name as the `ingress.tlsSecret` value.

If you are using cert manager, the built-in gateway config will cause it to issue a cert with each hostname specified. This was so it could be used without setting up a DNS-01 solver. I don't have strong opinions on this. The gateway setup is a bit more complicated than it needs to be to make it work with a standard HTTP solver. It can be simplified once cert-manger supports specifying the certificate on the HTTPRoute resource (current WIP it looks like).

> Are there other charts you've been looking at for inspiration on how to structure the values / what functionality to provide? 

Managing Kubernetes deployments is part of my day job, so I have seen far too many charts lol. From my perspective, there is no industry consensus on how to expose this config. For me, the easier to work with charts tend to handle it similarly to this structure. For example [Harbor](https://github.com/goharbor/harbor-helm) and the [Gerrit Operator](https://gerrit.googlesource.com/k8s-gerrit/+/refs/heads/master/helm-charts/gerrit-operator/).

For the functionality, basically all charts I work with that provide a gateway/http route based ingress provide the following features that I am specifically looking to add:

1. Some way to completely disable ingress as an "escape hatch."
2. generation of Ingress resources
3. generation of HTTPRoutes
4. (optional) generation of gateways.
    - In practice, a lot of production environments disable this and just use their own existing gateways, but I think this is particularly valuable to the community deployments, where they are probably only running ess-helm in the cluster.

As a last though I have structured it to be an exclusive "ingress or httproute" for each component. This is not actually a requirement, and I can change it so that both are possible. I just think having it exclusive makes a user less likely to shoot themselves in the foot by accident.

I'll look into getting the tests updated and passing this weekend.

Thanks again!

benbz:
> Hey @benbz, no problem, you guys seem quite busy. I don't use GitHub for work, so I didn't know whether it would auto-add a reviewer.
> 
> To shed some initial light on your questions:
> 
> > Is there a reason to put the Gateway values in with the Ingress values?
> > 
> > * i.e. have synapse.gateway.host as well as synapse.ingress.host use the present/absence of ...host to determine which to use?
> > * Or pull synapse.ingress.host up to synapse.host and have your type value or enabled flags under synapse {gateway,ingress}
> 
> For the actual Gateway config, I think that it should be promoted from `ingress.gateway` to a top-level key (`gateway`). For most people using this API, they will probably only want a single gateway, so they aren't paying for a bunch of cloud LoadBalancers needlessly. Multi-gateway configs can be handled externally through the `parentRefs` as an "advanced" feature.

Yup fully agree, we don't want to lead people to create lots of `Gateway`  resources. I don't think either of the structures proposed above would do that though, they're more focused on whether an `Ingress` or `HTTPRoute` is to be constructed for a given component.

Whatever is chosen I think the chart should lead users towards constructing a single `Gateway` for the whole chart or specifying an existing `Gateway` for the whole chart (with per-component overrides to specify other existing `Gateways`). If users want per-component `Gateways` I think they should construct them out of band and specify them on a per-component basis.

So a top-level `gateway` key as a sibling to the top-level `ingress` key which specifies either construction of the chart's single `Gateway` or the default external `Gateway` to use makes sense to me. However this is modulo the discussion below about how things should be constructed on a per-component basis as I'd want the top-level/global config structure to mirror the per-component one plus or minus config options that do/don't make sense in a global context

> In terms of overall structure, I am not 100% sure what the best approach is, but this was my thought process.
> 
>     * We have a set of config options that are shared between both ingress and httproute resources. To handle this we can either...
>       
>       * place shared keys in the parent component
>         
>         * This is clean for the "host" key, but would lead to things like "ingressAnnotations", etc
>         * We would need an ingress-esq key top-level key anyway to group the global versions of these shared options
>         * As a benifit we don't need to change the associated shared templates if other ingress types are added.
>       * duplicate shared keys
>         
>         * Don't try to share keys between HTTPRoute and Ingress resources
>         * Leads to more complicated templating that needs to be modified for each ingress type (some shared "core" template, then a wrapper to shim in the values from that specific config)
>         * Potential code drift for shared keys
>         * arguably a simpler config experience for the end user
>         * works well with the existing global + local override convention
>       * shared state + specific (this implementation)
>         
>         * works well with the existing global + local override convention
>         * able to reuse the existing templates mostly
>         * no drift on shared keys
>         * IMO easier to add new ingress types in the future since you only have to create the helpers for that specific type.
>         * I don't know how other people feel about specifying the "type" key

My main problems with the current approach are twofold
* Having a `HTTPRoute` subkey under `ingress` feels very confusing
* Lack of backwards compatibility for `className` and `controllerType`
  * This could be worked around and eventually removed but `<component>.ingress.Ingress.{className,controllerType}` also feels confusing

Potentially the original sin here is having `<component>.ingress` rather than `<component>.expose.ingress` like the Harbor chart. Refactoring the chart, maintaining backwards compatibility, to make this change does not feel palatable.

My gut feeling is that keeping `<component>.ingress` and introducing `<component>.gateway`, `<component>.gatewayApi` or `<component>.httpRoute` is best. We accept the duplication of properties and rely on the schema and tests to ensure common behaviour around e.g. TLS Secret handling or annotation merging. Whatever the mechanism for detecting one or the other is active (be it presence/absence of `<component>.ingress.host`, `<component>.ingress.enabled: true` or <component>.inboundTrafficMechanism: ingress`) we add validation that only 1 inbound traffic mechanism is active or not. i.e. we treat them as entirely independent things that just can't happen to be turned on at the same time.

@gaelgatelement would appreciate your thoughts on this
 
> I don't know how much of this code base you share with Pro, but IMO the main benefit of the last approach is that it tends to require less refactoring when you add a new ingress type. For instance, if you have a customer request for Istio `VirtualService` or ambassador based ingresses, you don't need to modify the shared templates as much in the future. It would be mostly contained to updating `ess-community.ess-library.ingress.isEnabled`.

ESS Community is a strict subset of ESS Pro, so all the code here gets incorporated into ESS Pro, with extra options as required. We make heavy use of the tests to ensure the same behaviour carries over to additional components that are only in ESS Pro and have further tests to validate additional behaviour that's only in ESS Pro.


> > When the chart creates its own Gateway are providing wildcard certs via a single value still possible?
> > ```
> > I think they are due to the element-io.gateway.tlsConfig helper still using coalesce but want to confirm
> > Alternatively would your expectation be that if you need that setup you must provide an existing gateway?
> > ```
> 
> It should be. I currently use a wildcard certificate issued by Let's Encrypt. You just specify the secret name as the `ingress.tlsSecret` value.
> 
> If you are using cert manager, the built-in gateway config will cause it to issue a cert with each hostname specified. This was so it could be used without setting up a DNS-01 solver. I don't have strong opinions on this. The gateway setup is a bit more complicated than it needs to be to make it work with a standard HTTP solver. It can be simplified once cert-manger supports specifying the certificate on the HTTPRoute resource (current WIP it looks like).

:+1: 

> > Are there other charts you've been looking at for inspiration on how to structure the values / what functionality to provide?
> 
> Managing Kubernetes deployments is part of my day job, so I have seen far too many charts lol. From my perspective, there is no industry consensus on how to expose this config. For me, the easier to work with charts tend to handle it similarly to this structure. For example [Harbor](https://github.com/goharbor/harbor-helm) and the [Gerrit Operator](https://gerrit.googlesource.com/k8s-gerrit/+/refs/heads/master/helm-charts/gerrit-operator/).

Looking at Harbor, everything is under an `expose` key with `type` and `ingress`/`route`/etc subkeys, which most closely maps to option 2 (`duplicate shared keys`). The Gerrit example I don't think is particularly relevant as it is just passing `ingress.type` as an env var to the operator and there's no further customisation.

> 
> For the functionality, basically all charts I work with that provide a gateway/http route based ingress provide the following features that I am specifically looking to add:
> 
>     1. Some way to completely disable ingress as an "escape hatch."
> 
>     2. generation of Ingress resources
> 
>     3. generation of HTTPRoutes
> 
>     4. (optional) generation of gateways.
>        
>        * In practice, a lot of production environments disable this and just use their own existing gateways, but I think this is particularly valuable to the community deployments, where they are probably only running ess-helm in the cluster.
> 
> 
> As a last though I have structured it to be an exclusive "ingress or httproute" for each component. This is not actually a requirement, and I can change it so that both are possible. I just think having it exclusive makes a user less likely to shoot themselves in the foot by accident.

I would be strongly in favour of exclusive `Ingress` or `HTTPRoute` too (in the `xor` sense that exactly 1 of them must be configured, configuring neither of them is not allowed)

JoshuaHassler:
Sorry for the delay in responding. I should have an updated patch tonight or tomorrow. It sounds like I am overall on track.

I've been putting some more thought into the config structure. Having the duplicate keys, I don’t think it will be as much of an issue as I had originally thought. ~I may be able to leverage the value schemas to simplify things, but I need to play around with them a bit to see how much of the JSON Schema spec they support.~ I also realized that HTTPRoute is probably a bad name, since certain components will likely use TCPRoutes in the future as an ingress option. For instance, SFU as an alternative to the NodePort model.

With that in mind, what do you think of the following structure?

* `.gateway`: Global gateway config specifically for the Gateway resource.
* `.routes / .<component>.routes`: configuration for the HTTP/TCP Routes. This is a bit more of a generic name so that each component can expose whatever route config it needs to function. For now, this is just HTTPRoutes.
* `.ingress / .<component>.ingress`: left as it is now

I am not particularly tied to `routes` as the key name, but I think it is important to distinguish the gateway config from the route config. The gateway-specific options would only be “global only,”  whereas the routes would follow the standard “global with component override” pattern.

For the selection of which ingress type to use, my initial thoughts would be to use a `.inboundTrafficMechanism / .<component>.inboundTrafficMechanism` with the options of `ingress | gatewayApi | none` for a few reasons. First, I feel like this makes it clear to a user that this is a mutually exclusive decision. Second, it has fewer edge cases to check than making sure only one “hosts” or “enabled” is set. Third, it is also easily enforceable with the values schema. Lastly, if someone is experimenting with multiple ingresses, they can set the values for both and easily switch between them.

While keeping the xor approach, I would also prefer to maintain the none option. I can understand if you would rather have me open another PR with it, but being able to disable ingress while keeping the component enabled gives much more flexibility in how things are deployed, particularly in more complex cluster setups that may not use ingress/gateways at all.

JoshuaHassler:
Turn out helm only supports a very limited subset of the JSON schema spec. This precludes any of the "fun" I was trying to have creating more composable ingress fragments. There are a few things that can probably be factored out into common fragments (like the global "service" object), but buy and large I just end up with a duplicate code path for the generation of the schema and values file.

I am still working on updating the helpers and templates, but I pushed most of the changes needed for the values file/schema. I also need to update the z_validation helpers to check that host is set on the appropriate resource instead of just on `ingress`.

Also an *extremely minor* bug I noticed while updating this. The ingress block macro only uses `global` to determine if the the `host` field should be generated. This results in it generating the `host` comment/key for the wellKnownDelegate block when it shouldn't.

I should have the updated templates and the start of testing pushed tomorrow after I finish cleaning it up.

charts/matrix-stack/source/common/routes_global.json
```diff
@@ -0,0 +1,72 @@
...
+        "type": "string"
+      }
+    },
+    "tlsEnabled": {
+      "type": "boolean"
+    },
+    "tlsSecret": {
+      "type": "string"
+    },
+    "service": {
```
benbz, line 49: Feels we need a `service_global.json` rather than having this distinct between `ingress_global.json` and `route_global.json`

charts/matrix-stack/source/element-admin.yaml.j2
```diff
@@ -14,6 +14,7 @@ enabled: true
 replicas: 1
 {{- sub_schema_values.image(registry='oci.element.io', repository='element-admin', tag='0.1.10') -}}
 {{- sub_schema_values.ingress() -}}
+{{- sub_schema_values.routes() -}}
```
benbz, line 17: I'm inclined to merge these so that we have somewhere sensible to put a commented out `inboundTrafficHandler` for each component and that it ensures that the values for one or the other can't be forgotten

charts/matrix-stack/templates/element-admin/ingress.yaml
```diff
@@ -16,7 +16,7 @@ metadata:
   namespace: {{ $.Release.Namespace }}
 spec:
 {{- include "element-io.ess-library.ingress.tls" (dict "root" $ "context" (dict "ingress" .ingress "ingressName" "element-admin")) | nindent 2 }}
-{{- include "element-io.ess-library.ingress.className" (dict "root" $ "context" .ingress.className) | nindent 2 }}
+{{- include "element-io.ess-library.ingress.className" (dict "root" $ "context" .ingress.Ingress.className) | nindent 2 }}
```
benbz: Needs reverting here and everywhere

benbz:
> With that in mind, what do you think of the following structure?
> 
>     * `.gateway`: Global gateway config specifically for the Gateway resource.
>     * `.routes / .<component>.routes`: configuration for the HTTP/TCP Routes. This is a bit more of a generic name so that each component can expose whatever route config it needs to function. For now, this is just HTTPRoutes.
>     * `.ingress / .<component>.ingress`: left as it is now

I don't object to this. The only improvement I can see would be to use `gatewayApi` for both the global gateway config and per-component routes.

Hmm. Looking through the diff so far has `gateway` for the global `Gateway` resource and a global `routes` for defaults all `HTTPRoutes`. So potentially what you have is better

> I also realized that HTTPRoute is probably a bad name, since certain components will likely use TCPRoutes in the future as an ingress option. For instance, SFU as an alternative to the NodePort model.

Unsure whether we'd do this under `*.routes`/`*.gatewayApi` or the existing `*.exposedServices` set of configuration. For a further phase

> For the selection of which ingress type to use, my initial thoughts would be to use a `.inboundTrafficMechanism / .<component>.inboundTrafficMechanism` with the options of `ingress | gatewayApi | none` for a few reasons. First, I feel like this makes it clear to a user that this is a mutually exclusive decision. Second, it has fewer edge cases to check than making sure only one “hosts” or “enabled” is set. Third, it is also easily enforceable with the values schema. Lastly, if someone is experimenting with multiple ingresses, they can set the values for both and easily switch between them.
> 
> While keeping the xor approach, I would also prefer to maintain the none option. I can understand if you would rather have me open another PR with it, but being able to disable ingress while keeping the component enabled gives much more flexibility in how things are deployed, particularly in more complex cluster setups that may not use ingress/gateways at all.

Makes sense. I can accept `none` as long as it is covered by manifest tests


> I am still working on updating the helpers and templates, but I pushed most of the changes needed for the values file/schema.

Makes sense. I've only reviewed the values/schema changes as a result.

Thanks

halkeye:
As synapse is one of my last nginx ingress, and I just found out about this chart to handle everything else as well, i'm very invested in this PR.

Is there anything I can do to help? my matrix isn't critical so I can help QA or whatever.

JoshuaHassler:
I believe I have updated all the templates for the new system and done the necessary helper refactoring. At this point I think I just need to add the testing, which has been a bit of a challenge since the public docs on how to run the tests are incomplete/wrong.

I have been a bit time constrained over the last few weeks with work, but will have time to hopefully complete the testing updates this weekend. I rebased my branch onto their latest changes in case you want to test anything. I do use my branch for my own matrix server, but I am always happy to have more eyes.

charts/matrix-stack/templates/matrix-rtc/ingress.yaml
```diff
@@ -5,13 +5,13 @@ Copyright 2025-2026 Element Creations Ltd
...
 {{- with .Values.matrixRTC -}}
-{{- if .enabled -}}
+{{- if eq (include "element-io.ess-library.ingress.isEnabled" (dict "root" $ "context" (dict "ingress" .ingress "type" "Ingress"))) "true" -}}
 apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
 {{- $extraAnnotations := dict }}
 {{- if .sfu.enabled }}
-{{- if eq "ingress-nginx" (include "element-io.ess-library.ingress-controller-type" (dict "root" $ "context" .ingress.controllerType)) }}
+{{- if eq "ingress-nginx" (include "element-io.ess-library.ingress-controller-type" (dict "root" $ "context" .ingress.Ingress.controllerType)) }}
```
halkeye:
```suggestion
{{- if eq "ingress-nginx" (include "element-io.ess-library.ingress-controller-type" (dict "root" $ "context" .ingress.ingress.controllerType)) }}
```
Pretty certain these should be lowercase ingress

charts/matrix-stack/templates/matrix-rtc/ingress.yaml
```diff
@@ -24,7 +24,7 @@ metadata:
   namespace: {{ $.Release.Namespace }}
 spec:
 {{- include "element-io.ess-library.ingress.tls" (dict "root" $ "context" (dict "ingress" .ingress "ingressName" "matrix-rtc")) | nindent 2 }}
-{{- include "element-io.ess-library.ingress.className" (dict "root" $ "context" .ingress.className) | nindent 2 }}
+{{- include "element-io.ess-library.ingress.className" (dict "root" $ "context" .ingress.Ingress.className) | nindent 2 }}
```
halkeye:
```suggestion
{{- include "element-io.ess-library.ingress.className" (dict "root" $ "context" .ingress.ingress.className) | nindent 2 }}
```

charts/matrix-stack/templates/matrix-rtc/ingress.yaml
jessebot: @JoshuaHassler sorry to bother, but could you take a quick peak at these suggestions? There are also a few merge conflicts to be resolved.

benbz:
I'm just coming up to speed on this PR again. IIUC the schema is mostly as discussed however the templates don't seem to match. e.g. there's usage of `inboundTrafficHandler`, things are still in the old format of `ingress.ingress` or `ingress.HTTPRoute`. Has something not been pushed?

benbz:
> the public docs on how to run the tests are incomplete/wrong.

If there's specific errors you're having issues please do call them out. It is hard to see the issues when you're using muscle memory day in day to do this

sebhoss:
It would be great if support for [ListenerSet](https://gateway-api.sigs.k8s.io/guides/user-guides/listener-set/) would be part of this. So instead of rendering a `Gateway`, you render a `ListenerSet` and attach your `HTTPRoute` to the `ListenerSet`

benbz:
> It would be great if support for [ListenerSet](https://gateway-api.sigs.k8s.io/guides/user-guides/listener-set/) would be part of this. So instead of rendering a `Gateway`, you render a `ListenerSet` and attach your `HTTPRoute` to the `ListenerSet`

We're not opposed to `ListenerSet` support but I think trying to do everything in 1 PR is fraught. I'd suggest
1. `HTTPRoute` construction support with `parentRefs` supporting both `kind: Gateway` and `kind: ListenerSet`
    * `Gateway` construction support - happy for this to be in the original PR (i.e. this) given it is required for GatewayAPI support to be needed in a blank cluster or a separate PR given that `HTTPRoute` support is valuable in and of itself
2. `ListenerSet` construction support, given it is an option and is an extended support feature

JoshuaHassler:
I pushed a few commits that finish the updates needed to split the traffic handling into separate "routes" and "ingress" keys under each component instead of the initial nested approach. All the current test cases are now passing and I have added the gateway as a component to the unit test framework. I have finally manged to wrap my head around the unit testing and have started implementing those.

I have to come back up to speed on the whole thread to make sure I didn't miss any comments/requests.

ListenerSets will actually simplify things quite a bit, but since they are not fully supported by a lot of gateway providers it is probably worth keeping the existing strategy for this PR and merging them separately (this is also why I didn't address TCP/UDP Routes for RTC in this PR).

For anyone who wants to test this, you should just have to make the following changes to your values file:
```yaml
inboundTrafficHandler: routes

# instead of setting host in ingress for each component, set it in routes
matrixWeb:
  routes:
    host: chat.yourdomain.tld
```

charts/matrix-stack/templates/ess-library/_inboundTrafficHandler.tpl
```diff
@@ -0,0 +1,68 @@
+{{- /*
+Copyright 2024 New Vector Ltd
+Copyright 2025-2026 Element Creations Ltd
+
+SPDX-License-Identifier: AGPL-3.0-only
+*/ -}}
+
+{{- define "element-io.ess-library.inboundTrafficHandler.host" -}}
```
benbz, line 8: Can this new helper and all associated usage + tests be extracted to a dedicated PR that can be reviewed and merged separately.

Obviously no GatewayAPI support added to the helper in that PR, it will come in this PR

charts/matrix-stack/source/values.schema.json
```diff
@@ -116,6 +119,37 @@
...
+        },
+        "listenFor": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        }
+      }
+    },
+    "inboundTrafficHandler": {
```
benbz, line 145: Can `inboundTrafficHandler` support be split out into its own PR(s). I don't have huge preference as to whether it is 2 (just `ingress` support and then adding `none` support) or 1 (`ingress` and `none` support together).

It might need to be layered on top of the PR that introduces `element-io.ess-library.inboundTrafficHandler.host` given that I think if you're making `<component>.{ingress,routes}.host` validation optional if `inboundTrafficHandler: none` then that helper should fail and all callers of that helper should be gated.

charts/matrix-stack/templates/synapse/_helpers.tpl
```diff
@@ -128,6 +129,7 @@ env:
 
 {{- define "element-io.synapse.ingress.additionalPaths" -}}
```
benbz, line 130: I'm looking at the changes to this helper and wondering whether we want to expand this pattern, e.g.
* Create `element-io.<component>.inboundTrafficHandler.paths` helper for each relevant component
* Returns a list of dictionaries with at least 4 elements: `path`, `service name`, `service port name` (for ingresses), `service port number` (for routes)
  * Synapse would have a 5th element: `availability`
* The actual `Ingress` / `HTTPRoute` manifests could then be generated in a common, consistent way
  * Again Synapse would need some special handling to deal with `availability`
* This is quite similar to how we do `env` handling for components except for the whole manifest
* Removes the possibility that we update the paths for `Ingress` or `HTTPRoute` but forget to update `HTTPRoute` or `Ingress` respectively
  * There's still the potential possibility to point at a different port by getting `name` vs `number` wrong but manifest tests can catch this

This would be a separate preparatory PR.

Thoughts?

benbz (commented):
Thanks for updating this PR. At this point in time I haven't re-engaged with the bulk of the new functionality for GatewayAPI.

What I have done is looked at where there's cross-cutting preparatory work needing to be done to make it easy to add GatewayAPI support. I've called out 2 or 3 places where I think valuable, independent dependencies can be reviewed and landed in more manageable way.

PR #1053, opened 2026-02-19
Matrix RTC: Make Turn TLS Termination on pod optional

gaelgatelement:
- Adds a new `tlsTerminationOnPod` parameter to allow to terminate TLS upstream
- This defaults to `true` for retro-compatibility with existing behaviour

Built partially with Mistral devstral-2.

### DIFF
--- a/charts/matrix-stack/configs/matrix-rtc/sfu/config-overrides.yaml.tpl
+++ b/charts/matrix-stack/configs/matrix-rtc/sfu/config-overrides.yaml.tpl
@@ -67,7 +67,7 @@ turn:
 {{ if .enabled }}
   tls_port: {{ .port }}
   domain: {{ tpl .domain $root }}
-{{- if and .tlsTerminationOnPod (or .tlsSecret $root.Values.certManager) }}
+{{- if .tlsTerminationOnPod }}
   cert_file: /turn-tls/tls.crt
   key_file: /turn-tls/tls.key
 {{- end }}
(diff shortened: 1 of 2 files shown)


charts/matrix-stack/templates/matrix-rtc/sfu_deployment.yaml
```diff
@@ -179,12 +179,14 @@ spec:
         - {{ (. | toYaml) | nindent 10 }}
 {{- end }}
 {{- if .exposedServices.turnTLS.enabled }}
+{{- if and .exposedServices.turnTLS.tlsTerminationOnPod (or .exposedServices.turnTLS.tlsSecret $.Values.certManager) }}
```
```suggestion
{{- if .exposedServices.turnTLS.tlsTerminationOnPod }}
```

Why isn't it just ^ given it was unconditional before?

charts/matrix-stack/configs/matrix-rtc/sfu/config-overrides.yaml.tpl
```diff
@@ -67,10 +67,12 @@ turn:
 {{ if .enabled }}
   tls_port: {{ .port }}
   domain: {{ tpl .domain $root }}
+{{- if and .tlsTerminationOnPod (or .tlsSecret $root.Values.certManager) }}
```
benbz:
```suggestion
{{- if .tlsTerminationOnPod }}
```

Why isn't it just ^ given it was unconditional before?

gaelgatelement: Yep, I started this by trying to rely on tlsSecret being optional, and added the boolean afterwards. But now we should rely on the boolean only indeed.

PR #1054, opened 2026-02-19
Add dynamic registration files rendering

gaelgatelement:
This drops usage of rendering of static registration files, and add supports for dynamically rendering them. This will allow to maintain them in the future, especially if we want to change hookshot registration file content for example.

Built partially with Mistral devstral-2.


charts/matrix-stack/ci/hookshot-dynamic-registration-externally-values.yaml
benbz, line 1: How does this differ from `charts/matrix-stack/ci/hookshot-secrets-externally-values.yaml`?

charts/matrix-stack/ci/hookshot-dynamic-registration-in-helm-values.yaml
benbz, line 1: How does this differ from `charts/matrix-stack/ci/hookshot-secrets-in-helm-values.yaml`?

charts/matrix-stack/source/hookshot.yaml.j2
```diff
@@ -20,7 +20,13 @@ logging:
   # Must be one of [critical, error, warning, info, debug]
   level: info
 
-{{ sub_schema_values.credential("Hookshot registration file to connect to a remote Synapse instance.", "appserviceRegistration", initIfAbsent=True) }}
+{{ sub_schema_values.credential("Hookshot registration file to connect to a remote Synapse instance.", "appserviceRegistration", initIfAbsent=False, commented=True) }}
```
benbz, line 23:
```suggestion
{{ sub_schema_values.credential("Hookshot registration file to connect to a Synapse instance.", "appserviceRegistration", initIfAbsent=False, commented=True) }}
```

This is also used for the Synapse in chart if the namespaces need to change, yes?

gaelgatelement: Yes indeed, that can be used for both

gaelgatelement:
After discussing, another approach will be done, where we allow re-generation of registration files instead.

benbz:
#1602 implements the regeneration of registration files from their templates

PR #1056, opened 2026-02-20
Migration: Add support for secret discovery

gaelgatelement:
- Each secret can exist either inline or in config
- This discovers secrets and prompt for missing ones
- Secrets are outputed in dedicated files
- Helm Values will refer to the secret data imported in kubernetes

### DIFF
--- a/scripts/migration/__main__.py
+++ b/scripts/migration/__main__.py
@@ -141,8 +141,15 @@ Examples:
         logger.setLevel(logging.CRITICAL)
 
     pretty_logger = logging.getLogger("migration:summary")
+    pretty_logger.propagate = False
     pretty_logger.setLevel(logging.CRITICAL if args.quiet else logging.INFO)
-    pretty_logger.addHandler(logging.StreamHandler())
+    pretty_sh = logging.StreamHandler()
+    pretty_sh.setFormatter(
+        logging.Formatter(
+            "%(message)s",
+        )
+    )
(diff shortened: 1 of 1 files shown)


scripts/migration/tests/test_main_e2e.py
```diff
@@ -17,11 +17,11 @@
 from .. import __main__
 
 
-def test_main_e2e_synapse_only(tmp_path, basic_synapse_config, write_synapse_config):
+def test_main_e2e_synapse_only(tmp_path, synapse_config_with_signing_key, write_synapse_config):
```
benbz, line 20: Given this test is explicitly using a config with the signing key, I'd expect to see the signing key asserted later on

scripts/migration/tests/test_synapse_secrets.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def test_discover_secrets_from_synapse_config(basic_synapse_config):
+    """Test discovering secrets from Synapse configuration."""
+    from scripts.migration.secrets import SecretDiscovery
+
+    # Start with basic config and add secret file references
+    synapse_config = basic_synapse_config.copy()
+    synapse_config["macaroon_secret_key"] = "./macaroon_key.txt"
+    synapse_config["registration_shared_secret_file"] = "my_registration_secret"
```
benbz: This looks like a direct secret and not a file

PR #1057, opened 2026-02-20
Cache oci dependencies

gaelgatelement:
This should hopefully make the flakes we are seeing disappear.

### DIFF
--- a/newsfragments/1057.internal.md
+++ b/newsfragments/1057.internal.md
@@ -1,1 +1,1 @@
-CI: Cache Helm OCI dependencies for a week.
+CI: Cache Helm OCI dependencies for the current week.


newsfragments/1057.internal.md
```diff
@@ -0,0 +1 @@
+CI: Cache Helm OCI dependencies for a week.
```
benbz:
```suggestion
CI: Cache Helm OCI dependencies for the current week.
```

Is more accurate?

tests/integration/lib/utils.py
```diff
@@ -255,3 +256,22 @@ async def forward_matching_logs(input_logs_queue: asyncio.Queue, output_matchers
...
+@asynccontextmanager
+async def chart_from_ci_cache(helm_client: pyhelm3.Client, chart_ref: str) -> AsyncGenerator[pyhelm3.Chart]:
+    project_root = Path(__file__).parent.parent.parent.parent
+    if os.environ.get("CI"):
+        helm_cache = project_root / ".pyhelm_cache"
+        helm_cache.mkdir(parents=True, exist_ok=True)
+        chart_name = chart_ref.split("/")[-1]
+        cached_ref = helm_cache / chart_name
+        if cached_ref.exists():
+            yield await helm_client.get_chart(cached_ref)
```
benbz: Do we want to do a soft (allowed to fail) pull as to whether there's a new version of the chart?

gaelgatelement: I dont think this would be very helpful, as the cache key cannot change depending on the chart version, unless we manage the version in the repository ? 🤔

benbz: My thinking is that if the chart changes during the week we've got differences between local and CI. Whereas a `pull_chart` call with Pokemon exception handling ("gotta catch em all") and falling back to the cache if it fails would mean we have the same behaviour as local (assuming it can be pulled)

gaelgatelement: That makes sense, hopefully we dont encounter strange timeouts with this strategy. Let's try.

PR #1067, opened 2026-02-24
Migration: Handle Extra Files automatically

gaelgatelement:
This feature discovers all paths found in Synapse configuration, and creates a ConfigMap per component, containing all the non-binary files discovered. The component values also gets the relevant `extraVolumes` and `extraVolumeMounts`, and has the new paths in its `additional` imported configuration.

Some additional details : 
- If the path is a key which we do not want to import, it is skipped (for example, config keys pointing to secrets we manage through ESS values)
- All paths found in the original configuration are updated when injected into Additional values
- Synapse gets `extraVolumes` and `extraVolumeMounts` with the files referenced at their expected path

### DIFF
--- a/scripts/migration/migration.py
+++ b/scripts/migration/migration.py
@@ -291,6 +291,8 @@ class ConfigValueTransformer:
             )
             config_maps.append(configmap)
 
+        extra_volume_mounts.sort(key=lambda x: x["mountPath"])
+
         if extra_volumes:
             config["extraVolumes"] = extra_volumes
             config["extraVolumeMounts"] = extra_volume_mounts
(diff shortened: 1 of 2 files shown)


scripts/migration/tests/test_extra_files.py
```diff
@@ -0,0 +1,429 @@
...
+                    config_inline="database.args.password",
+                    config_path=None,
+                ),
+                "synapse.signingKey": SecretConfig(
+                    init_if_missing_from_source_cfg=False,  # This would break federation if changing after migrating
+                    description="Signing key",
+                    config_inline="signing_key",
+                    config_path="signing_key_path",
+                ),
+            }
```
benbz: Can this be simpler as none of this is in the `config_data`?

PR #1074, opened 2026-02-26
Turn TLS tests: open socket against 127.0.0.1

gaelgatelement:
On some tests setup, the hostname cannot be resolved - this should rely on 127.0.0.1 directly instead. `server_hostname` will be used to verify the actual certificate.


benbz:
> On some tests setup, the hostname cannot be resolved

Why can't it be resolved? Why is the TURN hostname different to other hostnames?

PR #1079, opened 2026-02-26
Matrix RTC: Hardcode Turn ports in internal kubernetes traffic

gaelgatelement:
When Turn is exposed as port `443` on the LoadBalancer, this is causing the SFU to crash as it is trying to listen on the same port without having enough permissions. 

This PR changes the port and hardcodes it to the standard `5349`. This works because Livekit does not consider `turn.tls_port` value to advertise the remote access endpoint, it will always advertise `443` for Turn.

The same logic does not apply to RTC Ports : Those must be strictly the same between remote and local, as they are advertised as-is to remote clients.

### DIFF
--- a/charts/matrix-stack/configs/matrix-rtc/sfu/config-overrides.yaml.tpl
+++ b/charts/matrix-stack/configs/matrix-rtc/sfu/config-overrides.yaml.tpl
@@ -65,7 +65,11 @@ turn:
   enabled: true
 {{- with .exposedServices.turnTLS }}
 {{ if .enabled }}
+{{- if eq .portType "HostPort" }}
   tls_port: {{ .port }}
+{{- else }}
+  tls_port: 5349
+{{- end }}
   domain: {{ tpl .domain $root }}
 {{- if .tlsTerminationOnPod }}
   cert_file: /turn-tls/tls.crt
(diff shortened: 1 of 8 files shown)


charts/matrix-stack/configs/matrix-rtc/sfu/config-overrides.yaml.tpl
```diff
@@ -65,7 +65,11 @@ turn:
   enabled: true
 {{- with .exposedServices.turnTLS }}
 {{ if .enabled }}
+{{- if eq .portType "HostPort" }}
   tls_port: {{ .port }}
+{{- else }}
+  tls_port: 5349
```
jaywink, line 71: PR description mentions 3478?

charts/matrix-stack/templates/matrix-rtc/sfu_deployment.yaml
```diff
@@ -93,11 +93,13 @@ spec:
 {{- with .exposedServices }}
   {{- with .turnTLS }}
   {{- if .enabled }}
-        - containerPort: {{ .port }}
+        - name: turn-tls-tcp
     {{- if eq .portType "HostPort" }}
           hostPort: {{ .port }}
+          containerPort: {{ .port }}
+    {{- else }}
+          containerPort: 5479
```
```suggestion
          containerPort: 5349
```

charts/matrix-stack/configs/matrix-rtc/sfu/config-overrides.yaml.tpl
gaelgatelement: Yep I originally also changed the UDP port, but in practice it's not behaving like the TCP port, so I rolled back.

benbz (approved):

Issue #1028, opened 2026-02-08
Unable to configure custom KID for MAS signing keys, blocking migration from manual deployments


lucas-dclrcq:
**Describe the bug**

When migrating from a manual MAS deployment to the ESS Helm chart, there is no way to configure custom `kid` (Key ID) values for the signing keys. The KIDs are hardcoded in [`config-overrides.yaml.tpl`](https://github.com/element-hq/ess-helm/blob/main/charts/matrix-stack/configs/matrix-authentication-service/config-overrides.yaml.tpl) to static values (`rsa`, `prime256v1`, `secp256k1`, `secp384r1`):

```yaml
  - kid: rsa
    key_file: /secrets/...
  - kid: prime256v1
    key_file: /secrets/...
```

While the key material itself can be provided via `matrixAuthenticationService.privateKeys.rsa.secretKey` (or `.value`), the KID cannot be overridden. This breaks migration from existing MAS deployments where keys were registered with different KIDs, as existing tokens/sessions become invalid when the KID changes.

The [MAS documentation](https://element-hq.github.io/matrix-authentication-service/reference/configuration.html) states:
> A JWK Key ID is automatically derived from each key. To override this default, set `kid` to a custom value. The `kid` can be any case-sensitive string value as long as it is unique to this list; a key's `kid` value must be stable across restarts.

**To Reproduce**

1. Have an existing manual MAS deployment with signing keys using auto-derived or custom KIDs
2. Migrate to the ESS Helm chart, providing the same key material via `matrixAuthenticationService.privateKeys.rsa`
3. MAS starts with hardcoded `kid: rsa` instead of the original KID
4. Previously issued tokens/sessions fail validation because the KID in the JWT header no longer matches any known key

**Expected behavior**

The Helm chart should allow configuring custom `kid` values for each signing key, e.g.:

```yaml
matrixAuthenticationService:
  privateKeys:
    rsa:
      kid: "my-existing-kid"  # optional, defaults to "rsa" for backwards compatibility
      secret: existing-secret
      secretKey: key-in-secret
```

This would enable seamless migration from manual deployments by preserving the original KID values.

**Additional context**

- This is a migration blocker for anyone moving from a manual MAS deployment to the ESS Helm chart with pre-existing signing keys.
- The `values.yaml` `privateKeys` section currently only supports `value`, `secret`, and `secretKey` — there is no `kid` field.
[...]

PR #1083, opened 2026-02-27
Migration: Handle Matrix Authentication Service migration

gaelgatelement:


scripts/migration/tests/conftest.py
```diff
@@ -54,15 +54,44 @@ def synapse_config_with_signing_key(tmp_path, basic_synapse_config):
...
+@pytest.fixture
+def mas_config_with_keys(tmp_path, basic_mas_config):
+    """MAS configuration with a signing key file."""
+    # Create signing key file
+    rsa_file = tmp_path / "rsa_key.pem"
+    rsa_file.write_text("test_rsa_content")
+
+    basic_mas_config["secrets"].setdefault("keys", [])
+    basic_mas_config["secrets"]["keys"].append({"key_file": str(tmp_path / "rsa_key.pem"), "kid": "rsa"})
+    basic_mas_config["secrets"]["keys"].append({"key": "prime256v1_key_value", "kid": "prime256v1"})
```
benbz: Do we need to do #1028 before this?

gaelgatelement: Another approach would be to consider merging the list of keys. The migrated MAS instance would have the existing secret keys *and* the ESS secret keys. As long as the old secrets are not deleted, this should work just fine. I think #1028 is somewhat cleaner though, so I'll initiate a fix for it.

gaelgatelement: Arguably we will need to merge/append the current keys anyway, because we would break MAS signatures issued pre-migration.

gaelgatelement: Actually this is already the case. With https://github.com/element-hq/ess-helm/pull/1014, the existing keys will be added to the new keys generated by ESS post-migration. 

I think current behavior is fine, and maybe #1028 can already be closed ? After migration, #1028 would have ESS Keys + existing keys.

gaelgatelement: So according to @sandhose 

> so right now those keys are only used when MAS signs a JWT. This currently happens in two places:
> when it issues an id_token to clients. Matrix clients are not really supposed to validate id_tokens though, only Element Web does that
> when it uses that to 'authenticate' against an upstream provider. That's an alternative to the client_id/client_secret pair, that's when token_endpoint_auth_method is set to private_key_jwk. Pretty niche use case
In both cases, uuuusually the verifier will re-fetch the set of keys when it fails to verify (like it has a stale cached key set)

> the safest is to have both set of keys, the ones generated by ESS and the one imported. MAS will chose one when signing of them for signing, doesn't matter, what matters is that the public keys are available for clients to verify

According to this, I would let ESS generate new keys, but import the existing keys so that we have both after migration.

benbz: I've commented on https://github.com/element-hq/ess-helm/issues/1028#issuecomment-4022222986 with this, though I do think configurable `kid` is a nice to have

scripts/migration/tests/test_main_e2e.py
```diff
@@ -154,3 +154,77 @@ def test_main_e2e_synapse_only(
...
+    assert synapse_shared_secret_config["secret"] == "imported-matrix-authentication-service"
+    assert "secretKey" in synapse_shared_secret_config
+    assert synapse_shared_secret_config["secretKey"] == "matrixAuthenticationService.synapseSharedSecret"
+
+    # Check for Secret files (should be created for both Synapse and MAS secrets)
+    secret_files = list(output_dir.glob("*secret.yaml"))
+    # Should have at least two secret files (one for Synapse, one for MAS)
+    assert len(secret_files) >= 2, "Secret files should be created for both Synapse and MAS secrets"
+
+    # Verify the secret file content
```
benbz, line 231: This isn't really verifying any of the secret file content, only that it has the structure of a `Secret`

scripts/migration/tests/test_mas_secrets.py
```diff
@@ -0,0 +1,40 @@
...
+    # Should have discovered the password from database URI using transformer
+    assert "matrixAuthenticationService.postgres.password" in discovery.discovered_secrets
+    assert discovery.discovered_secrets["matrixAuthenticationService.postgres.password"].value == "mas_password"
+    assert "matrixAuthenticationService.privateKeys.rsa" in discovery.discovered_secrets
+    assert discovery.discovered_secrets["matrixAuthenticationService.privateKeys.rsa"].value == "test_rsa_content"
+    assert "matrixAuthenticationService.privateKeys.ecdsaPrime256v1" in discovery.discovered_secrets
+    assert (
+        discovery.discovered_secrets["matrixAuthenticationService.privateKeys.ecdsaPrime256v1"].value
+        == "prime256v1_key_value"
+    )
```
benbz: Can we also assert the various `kids`. Given that we need to import them as additional config with their old `kid`

scripts/migration/mas.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+# This SecretConfig factory discovers the MAS private keys according to their kids
+def extra_key_with_kid(source_config: dict[str, Any], kid: str) -> SecretConfig:
+    keys = get_nested_value(source_config, "secrets.keys")
+    config_inline = None
+    config_path = None
+    if keys:
+        for i, key in enumerate(keys):
+            if key["kid"] == kid:
```
benbz: `kid` needs to be saved

scripts/migration/mas.py
```diff
@@ -0,0 +1,217 @@
...
+            ),
+        }
+
+    def additional_secrets(self, config_data: dict[str, Any]) -> dict[str, SecretConfig]:
+        return {
+            "matrixAuthenticationService.privateKeys.rsa": extra_key_with_kid(config_data, "rsa"),
+            "matrixAuthenticationService.privateKeys.ecdsaPrime256v1": extra_key_with_kid(config_data, "prime256v1"),
+            "matrixAuthenticationService.privateKeys.ecdsaSecp256k1": extra_key_with_kid(config_data, "secp256k1"),
+            "matrixAuthenticationService.privateKeys.ecdsaSecp384r1": extra_key_with_kid(config_data, "secp384r1"),
+        }
```
benbz: As I've gone through this review I've become more convinced we should just do #1028. As it is we're going to have to save off the `kid` for each existing key and then handle the key entirely differently if the `kid` matches a value of one of the chart managed ones vs it being something distinct.

gaelgatelement: > As I've gone through this review I've become more convinced we should just do #1028. As it is we're going to have to save off the `kid` for each existing key and then handle the key entirely differently if the `kid` matches a value of one of the chart managed ones vs it being something distinct.

I agree, but I feel like this is a `pro`, not a `con`. By hardcoding kids, we know that those are the keys we want to manage with the chart. But what if the source installation has less keys than what ESS asks for ? There would be keys generated by ESS, and keys coming from an earlier setup. If the source has more keys, we still have to have keys in additional settings... 

Additionally, I feel like adding configurable kids to our values is going to generate unnecessary complexity. Typically, we cannot rely on the kid anymore to know the underlying key type. That means we will have to somehow detect the key type if we ever have to work with it in the future.

The issues I see with the current migration strategy are : 
- If a user has a kid `rsa` containing a non-rsa key, that might cause unexpected issues
- Importing existing keys in the additional config means we have to keep them forever here, which does not feel great

Considering @sandhose opinion, I think I would also be happy with not importing keys at all, and just re-generating those upon migration to ESS.

benbz: We could go the other way and remove `kid` entirely (from migration, from the chart managed keys, etc). https://element-hq.github.io/matrix-authentication-service/reference/configuration.html#secretskeys says
> A [JWK Key ID](https://datatracker.ietf.org/doc/html/rfc7517#section-4.5) is automatically derived from each key

If it is safe enough to have the `kid` change, which is what we appear to be saying (despite the docs), let's remove all this confusion

gaelgatelement: https://github.com/element-hq/ess-helm/pull/1112

gaelgatelement: OK this has now been refactored so that we do not import the existing keys at all.

scripts/migration/mas.py
gaelgatelement: This has now been refactored so that we do not import the existing keys at all (Follow up of conversations + #1112 )

scripts/migration/tests/test_mas_secrets.py
gaelgatelement: This has now been refactored so that we do not import the existing keys at all (Follow up of conversations here + https://github.com/element-hq/ess-helm/pull/1112 )

benbz: Err, my understanding was that we'd let the `kid` change but keep the key themselves

gaelgatelement: Meaning we should try to discover the type of keys found ?

benbz: Urgh that's a point. And there's different formats they could be in too

gaelgatelement: Yeah I think the discovery would have to : 

- Look for files in `keys_dir`, file pointed at `key_file` and content of `key`
- For each, try to read it as :  
  - RSA PEM
  - RSA DER
  - ECDSA Prime256 v1 PEM
  - ECDSA Prime256 v1 DER

I think that *might* be doable if we wanted to. The only cool thing is that it might help migrating to an init-secrets disabled ESS setup.

scripts/migration/tests/conftest.py
```diff
@@ -105,3 +105,187 @@ def synapse_config_with_ca_federation_list(tmp_path, basic_synapse_config):
...
+        },
+    }
+
+
+@pytest.fixture
+def basic_mas_config_with_keys(tmp_path, basic_mas_config):
+    """MAS configuration with keys directory and sample keys."""
+    from cryptography.hazmat.backends import default_backend
+    from cryptography.hazmat.primitives import serialization
+    from cryptography.hazmat.primitives.asymmetric import ec, rsa
```
benbz: Can we do these imports just once at the top, unless there's a compelling reason to do them per-fixture?

scripts/migration/tests/test_mas_secrets.py
gaelgatelement: This is now done!

PR #1087, opened 2026-03-01
Add Caddy reverse proxy example

moritzdietz:
This PR adds an example for Caddy when ESS is put behind a reverse proxy.

As I have seen others and me in the ESS Matrix room trying to come up with our own Caddy configuration, I thought it would be valuable to add it to the docs itself and iterate on it there if needed.

As Caddyfiles are tiny and don't require much tweaking because of sane defaults, this configuration fully works and don't bloat the README too much IMO.

Some notes:
- Caddy always does TLS 1.2 and TLS 1.3 unless told to in a TLS directive, so this won't break Calling on Element X iOS
- Caddy also set the correct headers by default when you add a reverse proxy directive
- I also wanted to add a second example to show how delegation works to those who never have set up Synapse before
- The well known files are also passed through to ESS 

I didn't coordinate in the chat before opening this PR. Hope this is still ok :)

### DIFF
--- a/newsfragments/1087.added.md
+++ b/newsfragments/1087.added.md
@@ -1,1 +1,1 @@
-Add example config for Caddy reverse proxy. (#1087)
+Add example config for Caddy reverse proxy.


moritzdietz:
I woke up and realized: we can probably just use a reverse proxy to handle the well-known files as well from the main block:
```diff
diff --git i/README.md w/README.md
index 734a7c55..3d2bcad4 100644
--- i/README.md
+++ w/README.md
@@ -467,10 +467,7 @@ example.com matrix.example.com account.example.com mrtc.example.com chat.example
 example.com {
   # Other directives or matcher definitions i.e. your personal homepage or a blog under example.com/blog/ could be in here as well
   #
-  header /.well-known/matrix/* Content-Type application/json
-  header /.well-known/matrix/* Access-Control-Allow-Origin *
-  respond /.well-known/matrix/server `{"m.server":"matrix.example.com:443"}`
-  respond /.well-known/matrix/client `{"m.homeserver":{"base_url":"https://matrix.example.com"},"org.matrix.msc2965.authentication":{"account":"https://account.example.com/account","issuer":"https://account.example.com/"},"org.matrix.msc4143.rtc_foci":[{"livekit_service_url":"https://mrtc.example.com","type":"livekit"}]}`
+  reverse_proxy /.well-known/matrix/* http://127.0.0.1:8080
 }

 admin.example.com chat.example.com account.example.com mrtc.example.com matrix.example.com {
```

That would probably be better in terms of who-manages-what. I think letting ESS handle the well-known files seems like the better approach.

benbz:
> I woke up and realized: we can probably just use a reverse proxy to handle the well-known files as well from the main block:
> 
> ```diff
> diff --git i/README.md w/README.md
> index 734a7c55..3d2bcad4 100644
> --- i/README.md
> +++ w/README.md
> @@ -467,10 +467,7 @@ example.com matrix.example.com account.example.com mrtc.example.com chat.example
>  example.com {
>    # Other directives or matcher definitions i.e. your personal homepage or a blog under example.com/blog/ could be in here as well
>    #
> -  header /.well-known/matrix/* Content-Type application/json
> -  header /.well-known/matrix/* Access-Control-Allow-Origin *
> -  respond /.well-known/matrix/server `{"m.server":"matrix.example.com:443"}`
> -  respond /.well-known/matrix/client `{"m.homeserver":{"base_url":"https://matrix.example.com"},"org.matrix.msc2965.authentication":{"account":"https://account.example.com/account","issuer":"https://account.example.com/"},"org.matrix.msc4143.rtc_foci":[{"livekit_service_url":"https://mrtc.example.com","type":"livekit"}]}`
> +  reverse_proxy /.well-known/matrix/* http://127.0.0.1:8080
>  }
> 
>  admin.example.com chat.example.com account.example.com mrtc.example.com matrix.example.com {
> ```
> 
> That would probably be better in terms of who-manages-what. I think letting ESS handle the well-known files seems like the better approach.

As per https://docs.element.io/latest/element-server-suite-pro/advanced/#external-well-knowns (for Pro but the same principals apply) we strongly recommend proxying through to ESS' well-knowns rather than reimplementing them elsewhere. If you could update this PR to do the proxying that'd be appreicated

moritzdietz:
I just updated the PR as suggested 👍🏻

benbz:
https://github.com/element-hq/ess-helm/pull/1096 to fix the `helm-lint` failure for you.

newsfragments/1087.added.md
```diff
@@ -0,0 +1 @@
+Add example config for Caddy reverse proxy. (#1087)
```
benbz:
```suggestion
Add example config for Caddy reverse proxy.
```

This gets added into the release notes automatically

PR #1098, opened 2026-03-03
Fix indentation and remove tab character

moritzdietz:
Noticed this when browsing the docs. I checked this with the live output of the command from my setup to get the indentation right.

### DIFF
--- a/README.md
+++ b/README.md
@@ -496,8 +496,8 @@ This should give you the following output:
 Defaulted container "matrix-authentication-service" out of: matrix-authentication-service, render-config (init), db-wait (init), config (init)
 ✔ Username · alice
 User attributes
-    	Username: alice
-   	Matrix ID: @alice:thisservername.tld
+         Username: alice
+        Matrix ID: @alice:thisservername.tld
 No email address provided, user will be prompted to add one
 No password or upstream provider mapping provided, user will not be able to log in
 
@@ -508,9 +508,9 @@ Non-interactive equivalent to create this user:
 ✔ What do you want to do next? (<Esc> to abort) · Set a password
 ✔ Password · ********
 User attributes
-    	Username: alice
-   	Matrix ID: @alice:thisservername.tld
-    	Password: ********
+         Username: alice
+        Matrix ID: @alice:thisservername.tld
+         Password: ********
 No email address provided, user will be prompted to add one
 ```
 


README.md
```diff
@@ -496,8 +496,8 @@ This should give you the following output:
 Defaulted container "matrix-authentication-service" out of: matrix-authentication-service, render-config (init), db-wait (init), config (init)
 ✔ Username · alice
 User attributes
-    	Username: alice
-   	Matrix ID: @alice:thisservername.tld
+         Username: alice
+        Matrix ID: @alice:thisservername.tld
```
benbz, line 500:
```suggestion
         Matrix ID: @alice:thisservername.tld
```

README.md
```diff
@@ -508,9 +508,9 @@ Non-interactive equivalent to create this user:
 ✔ What do you want to do next? (<Esc> to abort) · Set a password
 ✔ Password · ********
 User attributes
-    	Username: alice
-   	Matrix ID: @alice:thisservername.tld
-    	Password: ********
+         Username: alice
+        Matrix ID: @alice:thisservername.tld
```
benbz, line 512:
```suggestion
         Matrix ID: @alice:thisservername.tld
```

benbz (commented):
Appreciate these weren't well aligned before, but if we're fixing them let's align them now

README.md
moritzdietz: But they are right aligned and not left aligned in the output, no?
```
✔ What do you want to do next? (<Esc> to abort) · Set a password
✔ Password · ********
User attributes
         Username: administrator
        Matrix ID: @administrator:828f80fb.ess.lab.moritzdietz.com
         Password: ********
No email address provided, user will be prompted to add one
Can request admin: true
```

README.md
moritzdietz: Same as above
```
register-user
Defaulted container "matrix-authentication-service" out of: matrix-authentication-service, render-config (init), db-wait (init), database-migrate (init)
✔ Username · administrator
User attributes
         Username: administrator
        Matrix ID: @administrator:828f80fb.ess.lab.moritzdietz.com
No email address provided, user will be prompted to add one
No password or upstream provider mapping provided, user will not be able to log in
```

README.md
benbz: Ah I hadn't noticed the alignment on `:`. I just noticed the mis-alignment on the field names. This is fine then

moritzdietz:
Does this need a newsfile entry or is this good as is? If so, I'll adhere to the advice to from the other PR to not include the PR number in the text itself.

If not I'll add one and squash it into one commit. Or do you prefer to squash merge at the end of the PR?

benbz:
> Does this need a newsfile entry or is this good as is? If so, I'll adhere to the advice to from the other PR to not include the PR number in the text itself.
> 
> If not I'll add one and squash it into one commit. Or do you prefer to squash merge at the end of the PR?

I'm happy to take this one without a changelog fragment

PR #1103, opened 2026-03-05
Add optional file pattern argument to ct-lint.sh

gaelgatelement:
This allows selective linting of CI values files by passing a pattern argument, similar to how `assemble_ci_values_files_from_fragments.sh` works.

When a pattern is provided, the script temporarily moves non-matching `*-values.yaml` files from `charts/matrix-stack/ci/ `to a temporary directory, runs `ct lint` with only the matching files, then restores all files.

The script now accepts an optional file pattern that filters which CI values files are processed by ct lint.

Built with mistral codestral.

### DIFF
--- a/scripts/ct-lint.sh
+++ b/scripts/ct-lint.sh
@@ -1,11 +1,68 @@
 #!/usr/bin/env bash
 
 # Copyright 2024 New Vector Ltd
-# Copyright 2025 Element Creations Ltd
+# Copyright 2025-2026 Element Creations Ltd
 #
 # SPDX-License-Identifier: AGPL-3.0-only
 
 set -euo pipefail
+
+# Parse arguments: first argument is optional file pattern, rest are passed to ct lint
+file_pattern="*"
+ct_args=()
+
+# If first argument doesn't start with -, it's a file pattern
+if [[ "$#" -gt 0 && "${1:0:1}" != "-" ]]; then
+  file_pattern="$1"
+  shift  # Remove file pattern from arguments
+fi
+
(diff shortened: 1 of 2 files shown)


scripts/ct-lint.sh
```diff
@@ -1,26 +1,83 @@
...
+    done
+  fi
+fi
+
 temp_output_file=$(mktemp)
 
 error=1
 
-find . -type f -name '*.tpl' -exec grep -nE '\{\{[^}]*\$[^a-zA-Z0-9_][^}]*\}\}' {} + && {
+find . -type f -name '*.tpl' -path "*/$file_pattern*" -exec grep -nE '\{\{[^}]*\$[^a-zA-Z0-9_][^}]*\}\}' {} + && {
```
benbz: I don't think this should change as it isn't related to values files

scripts/ct-lint.sh
```diff
@@ -1,26 +1,83 @@
...
   echo 'Error: $ is used in a .tpl files, but helm passes the local context to the special variable $ in included templates.'; exit 1 
 } || echo "OK."
 
-find . '(' -type f -name '*.tpl' -o -name '*.yaml' ')' -exec grep -nE '\{\{[^}]*merge\s[^}]*\}\}' {} + && {
+find . '(' -type f -name '*.tpl' -o -name '*.yaml' ')' -path "*/$file_pattern*" -exec grep -nE '\{\{[^}]*merge\s[^}]*\}\}' {} + && {
```
benbz: I don't think this should change as it isn't related to values files

Issue #1060, opened 2026-02-22
Traefik config doesn't work


LovesTha:
**Describe the bug**
Doing the config for k3s behind a reverse proxy, creating `/var/lib/rancher/k3s/server/manifests/traefik-config.yaml` does nothing to which ports traefik was binding to.

I noticed `/var/lib/rancher/k3s/server/manifests/traefik.yaml` exists, but merging the suggested config into that file also did nothing.

**To Reproduce**
Steps to reproduce the behaviour:
1. Follow instructions for `Using an existing reverse proxy`
2. Output of `kubectl get svc -n kube-system | grep traefik` is still ports 80 & 443

NB: traefik service in that output indicates it hasn't been restarted, the uptime is large, but so is the uptime in the example output.

**Expected behaviour**
Traefik is reloaded with customised exposed ports

**Desktop (please complete the following information):**
N/A

**Server:**
 - OS: OpenSuse Tumbleweed 20250220

gaelgatelement:
Hello,

I tried again on a local VM and it works just fine : 

```
[arch@archlinux ~]$  kubectl get svc/traefik -n kube-system
NAME      TYPE           CLUSTER-IP    EXTERNAL-IP       PORT(S)                      AGE
traefik   LoadBalancer   10.43.66.56   192.168.200.251   80:30356/TCP,443:32320/TCP   34s
[arch@archlinux ~]$ sudo vim /var/lib/rancher/k3s/server/manifests/traefik-config.yaml
[arch@archlinux ~]$  kubectl get svc/traefik -n kube-system -w
NAME      TYPE           CLUSTER-IP    EXTERNAL-IP       PORT(S)                      AGE
traefik   LoadBalancer   10.43.66.56   192.168.200.251   80:30356/TCP,443:32320/TCP   65s
traefik   LoadBalancer   10.43.66.56   192.168.200.251,192.168.200.251   8080:30356/TCP,8443:32320/TCP   71s
```

There might be something wrong in what you re doing maybe?

MonstraG:
Happened for me too, started googling, found this issue, went back to open it again just to check, it started working without any changes. Delay on config change?

Loonesbury:
I also had this issue. Two mistakes I made:
- Checking that I only using spaces for indentation. I pasted into vim, which happily added in tabs while I was editing the file.
- I didn't realize the grave accents in the example `externalIPs` were meant to be replaced as well.
```yaml
apiVersion: helm.cattle.io/v1
[...]

PR #1110, opened 2026-03-07
Update README.md Clarification on port 443, revisions to external proxy instructions

AndyQuantonium:
Clarified that port 443 is used with MatrixRTC. I ran into an issue with video chat not working, due to the certificate for the mRTC domain missing. I was under the impression that the mRTC domain was only used for ports 30881 and 30882, however it handles HTTPS requests as well.

Revised traefik-config.yaml instructions as the previous file content was not working (related to #1060). This is identical to the file in my config, which has updated the ports traefik was listening on appropriately. I am not sure if this is the definitive fix. This is an issue with k3s which I found reported elsewhere (though I cannot find the issue I saw this in currently).

Also changed step 3 with `kubectl apply ...` which directly applies the file, rather than assuming k3s will update automatically.


README.md
```diff
@@ -348,13 +348,10 @@ spec:
...
       websecure:
         exposedPort: 8443
-    service:
-      spec:
-        externalIPs:
-        - `<external IP returned by the command above>`
 ```
 
-3. K3s will apply the file content automatically. You can verify its ports using the command :
+3. Update the LoadBalancer service with `kubectl apply -f /var/lib/rancher/k3s/server/manifests/traefik-config.yaml`
```
gaelgatelement: This should not be necessary, k3s is watching the manifests in this directory https://docs.k3s.io/installation/packaged-components#auto-deploying-manifests-addons

> On server nodes, any file found in /var/lib/rancher/k3s/server/manifests will automatically be deployed to Kubernetes in a manner similar to kubectl apply

README.md
```diff
@@ -178,7 +178,7 @@ You need to create DNS records (A or CNAME, for example) to set up ESS Community
 
 For this simple setup you need to open the following ports:
  - TCP 80: This port will be used for the HTTP connections of all services, which will redirect to the HTTPS connection.
- - TCP 443: This port will be used for the HTTPS connections of all services.
+ - TCP 443: This port will be used for the HTTPS connections of all services. This port is also used by the Matrix RTC Backend for authentication over HTTPS.
```
gaelgatelement: It feels a bit odd to clarify it only for Matrix RTC Backend here. What was the source of confusion in the first place?

README.md
AndyQuantonium: The k3s docs don't specify how the files are automatically deployed, but perhaps in setting my deployment up, that wasn't the culprit of the issue with traefik's ports not changing. This change could be left out if the automatic updating documented in the k3s docs works consistently.

README.md
AndyQuantonium: The source of confusion about the port was what the domains would point to. In my setup I configured k3s to run alongside an existing traefik docker container. I thought that there would be no traffic on the mRTC domain on https, so I didn't configure traefik to reverse proxy connections under the domain, and relied on port forwarding on the host to route the mRTC ports. After finding out via the browser console that http requests are sent to the mRTC domain to authenticate, I updated traefik to handle the additional domain.

In a more advanced setup, it's possible to host the mRTC backend on a different machine, which would mean the domain would route to a different IP. As is, the documentation may lead someone to expose only ports 30881 and 30882 on the other machine, when at least port 443 is also needed.

gaelgatelement:
```suggestion
 - TCP 443: This port will be used for the HTTPS connections of all services (including Matrix RTC for authentication).
```

gaelgatelement: I'm suggesting a minor rephrasing to try to keep it succinct then.

README.md
gaelgatelement: Can you remove these changes from your branch? Happy to press the merge button once done! Thanks!

AndyQuantonium: I updated the file with the previous step 3.

README.md
```diff
@@ -348,10 +348,6 @@ spec:
         exposedPort: 8080
       websecure:
         exposedPort: 8443
-    service:
-      spec:
-        externalIPs:
-        - `<external IP returned by the command above>`
```
gaelgatelement, line 387: Is that really not required ? 🤔

AndyQuantonium: Here's the config in my server:
```
~# cat /var/lib/rancher/k3s/server/manifests/traefik-config.yaml
apiVersion: helm.cattle.io/v1
kind: HelmChartConfig
metadata:
  name: traefik
  namespace: kube-system
spec:
  valuesContent: |-
    ports:
      web:
        exposedPort: 8888
      websecure:
        exposedPort: 4430
```

Like with issue #1060 I was getting the issue where the port wouldn't change despite the file being present. I tried different things, what I know is this config I have works. Though one thing in peculiar, the external IP traefik's service is on is the external IP to the server (ideally I would want this set manually to an internal IP).

gaelgatelement: I tried without setting an external IP and it would not bind here AFAICT. That's a bit surprising.

I agree that it would be great to make k3s traefik to bind on alternatively on a local IP. If you are able to document it that would be a very valuable contribution!

PR #1114, opened 2026-03-09
ci: fix dyff comment broken after hardening

gaelgatelement:


.github/workflows/templates-dyff-comment.yml
```diff
@@ -19,25 +19,43 @@ jobs:
...
     if: github.event.workflow_run.conclusion == 'success'
     steps:
+    - name: Fetch workflow run details
+      uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd  # v8
+      id: fetch-workflow
+      with:
+        script: |
+          const runId = context.payload.workflow_run.id;
+          const owner = context.repo.owner;
+          const repo = context.repo.repo;
```
benbz, line 29: I assume for a PR this is the current owner/repo and not the owner/repo of the PR fork?

gaelgatelement: The owner & repo are built from `GITHUB_REPOSITORY` env var : https://github.com/actions/toolkit/blob/main/packages/github/src/context.ts#L66-L83

As far as I can tell, this variable is always going to be `element-hq/ess-helm` in the context of our CI.

PR #1118, opened 2026-03-10
Matrix RTC: Change default Static ports to move them to Kubernetes defaults

gaelgatelement:
Kubernetes [default static range](https://kubernetes.io/docs/concepts/services-networking/service/#avoid-nodeport-collisions) is 
> When using the default NodePort range 30000-32767, the bands are partitioned as follows:
>
>    Static band: 30000-30085
>    Dynamic band: 30086-32767

We have been wrongly using a default static port in the dynamic band. This fixes this wrong defaults.

Users can choose to update their firewalls rules, or keep using the previous defaults.

This also changes the ports used in CI and relies on dynamic port allocation. This should remove Port conflicts failures from CI Runs, and should allow running multiple setups of pytest Matrix RTC tests runs locally.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -117,6 +117,9 @@ jobs:
     - name: Deploy with pytest once
       env:
         PYTEST_CI_FIRST_STEP: "1"
+      # Drop me after releasing next 26.3.x
+      # We are changing the k3d ports range, so we cannot setup matrix-rtc with the old values
+      if: ${{ !contains(env.MATRIX_TEST_COMPONENT, 'matrix-rtc') }}
       run: |
         if [ -f "tests/integration/env/${MATRIX_TEST_COMPONENT}.rc" ]; then
           # shellcheck source=/dev/null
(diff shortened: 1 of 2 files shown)


tests/integration/test_element_call.py
```diff
@@ -80,3 +97,88 @@ async def _assert_tls_socket():
...
+    # Verify it has a NodePort assigned
+    rtc_udp_port = None
+    if rtc_udp_service.spec and rtc_udp_service.spec.ports:
+        for port in rtc_udp_service.spec.ports:
+            if port.name == "rtc-muxed-udp":
+                rtc_udp_port = port.nodePort
+                break
+
+    assert rtc_udp_port is not None, "UDP service should have NodePort assigned"
+    assert 30000 <= rtc_udp_port <= 32767, "NodePort should be in valid range"
```
benbz: Should this assertion be in the other 2 tests too? Should the range ensure that NodePort is in the dynamic part not the static part?

newsfragments/1118.removed.md
benbz, line 1: Should be `breaking` and not `changed`

gaelgatelement: Mmh there's no such type : 

```
towncrier create
Issue number (`+` if none): 
Fragment type (security, removed, deprecated, added, changed, fixed, doc, internal):
```

I think it's `removed` 

```

[[tool.towncrier.type]]
directory = "removed"
name = "Removed / Breaking Changes"
showcontent = true
```

Which feels confusing. I'll duplicate `removed` and add `breaking`

PR #1119, opened 2026-03-10
CI: Persist credentials when releasing

gaelgatelement:
We are creating a commit to change the next chart version with a PR. We have to persist git credentials to make it work.

### DIFF
--- a/.github/workflows/releasing.yml
+++ b/.github/workflows/releasing.yml
@@ -105,7 +105,8 @@ jobs:
     - name: Checkout
       uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6
       with:
-        persist-credentials: false
+        # We are creating a commit changing the chart version for the next version PR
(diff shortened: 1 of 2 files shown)


.github/workflows/releasing.yml
```diff
@@ -105,7 +105,7 @@ jobs:
     - name: Checkout
       uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6
       with:
-        persist-credentials: false
+        persist-credentials: true
```
benbz, line 109: Can we have a comment here as to why

gaelgatelement: Well actually I wanted to comment with zizmor ignore, but ... It did not raise any error :thinking:

gaelgatelement: Ah it is expected actually : https://docs.zizmor.sh/audits/#remediation_2

PR #1129, opened 2026-03-11
Replace poetry with uv

gaelgatelement:
- UV is faster, allows for local workspace and in general is the preferred python package manager nowadays


.github/workflows/build-test.yml
```diff
@@ -27,13 +27,17 @@ jobs:
...
         persist-credentials: false
 
-    - uses: matrix-org/setup-python-poetry@5bbf6603c5c930615ec8a29f1b5d7d258d905aa4  # v2
+    - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # v6
       with:
         python-version: "3.x"
 
-    - name: Load poetry path
-      run: |
-        echo "$(poetry env info -p)/bin" >> "${GITHUB_PATH}"
```
benbz, line 36: Can we do something similar in the `Set up UV environment` step so that we don't need to prefix everything with `uv run`?

pyproject.toml
```diff
@@ -3,54 +3,55 @@
...
-
-[build-system]
-requires = ["poetry-core>=2.1.0"]
-build-backend = "poetry.core.masonry.api"
+requires-python = ">=3.12,<4.0"
+dependencies = [
+    "frozendict>=2.4.7",
+    "types-jsonschema>=4.26.0.20260202",
+    "rapidfuzz>=3.14.3",
+]
```
benbz: I don't think these are deliberately here and not `dev-dependencies`?

.github/workflows/build-test.yml
gaelgatelement: `setup-uv` logs show : 

```

Added /home/runner/.local/bin to the path
Added /opt/hostedtoolcache/uv/0.10.9/x86_64 to the path
Set UV_PYTHON_INSTALL_DIR to /home/runner/work/_temp/uv-python-dir
Added /home/runner/work/_temp/uv-python-dir to the path
Set UV_CACHE_DIR to /home/runner/work/_temp/setup-uv-cache

```

So I would have expected the bin path to already by in github path.

gaelgatelement: Yes it is working properly after enabling the venv activate flag!

.github/workflows/changelog_check.yml
```diff
@@ -25,21 +25,25 @@ jobs:
...
         with:
           python-version: "3.x"
 
-      - name: Load poetry path
-        run: |
-          echo "$(poetry env info -p)/bin" >> "${GITHUB_PATH}"
+      - name: Install uv
+        uses: astral-sh/setup-uv@6ee6290f1cbc4156c0bdd66691b2c144ef8df19a  # v7
+        with:
+          enable-cache: true
```
sandhose, line 31: setting `activate-environment: true` creates a venv automatically and activates it

.github/workflows/changelog_check.yml
```diff
@@ -25,13 +25,17 @@ jobs:
           persist-credentials: false
           fetch-depth: 0
 
-      - uses: matrix-org/setup-python-poetry@5bbf6603c5c930615ec8a29f1b5d7d258d905aa4  # v2
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # v6
         with:
           python-version: "3.x"
```
sandhose: I think `uv` will download a python for you. Either set in `python-version` input in the `setup-uv` action or a `.python-version` file in the repo

PR #1130, opened 2026-03-12
update project to rely on python 3.14 explicitly

gaelgatelement:
This has been the case for a little while actually, but it was never updated in docs nor in the python project file.

### DIFF
--- a/scripts/migration/utils.py
+++ b/scripts/migration/utils.py
@@ -110,7 +110,7 @@ def get_nested_value(config: dict[str, Any], path: str) -> Any:
             elif isinstance(current, list):
                 try:
                     current = current[int(part)]
-                except (IndexError, ValueError):
+                except IndexError, ValueError:
                     return None
             elif part != parts[-1]:
                 return None  # Can't navigate further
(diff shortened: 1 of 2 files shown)


pyproject.toml
sandhose, line 1: Careful with this, if you're looking to ship tools that people can easily install on any setup (for the migration to ESS script), you probably want to support all the current Python versions unfortunately… which right now means 3.10 minimum. That's what ships in a still-supported Ubuntu LTS (22.04)

That's also what we support as part of our support policy on the Synapse side

gaelgatelement: Can we have distinct python version between the main project and a given workspace?

sandhose: Yes, but that means your tests will still run with the most recent version of python. It should although make `ruff` (and maybe `mypy`?) recognise the target python version and maybe tell you if you're using unsupported syntax/libraries?

If you're not using specific syntax/features requiring a more modern python version, I would just stick to 3.10

gaelgatelement: Well, I am, so I'll have to make the code work with older python version. But I'd like to do it only for the migration scripts.

gaelgatelement: I wll handle python version for migration package in https://github.com/element-hq/ess-helm/pull/1131

gaelgatelement:
In the end, we are going to stay compatible with python > 3.11 so that the migration scripts are available to as much OS as possible.

PR #1131, opened 2026-03-12
Move ess-migration-tool to a packages directory

gaelgatelement:
- We should distinguish what are packages we intend end-users to install (application/libraries) vs internal development scripts

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,17 +15,13 @@ dependencies = []
 
 [tool.uv]
 [dependency-groups]
-# This should be kept in sync with packages/ess-migration-tool
-dev-tools = [
-        "ruff>=0.15.4",
-        "pytest>=8.4.2",
-        "mypy>=1.19.1",
-        "types-pyyaml>=6.0.12.20250915",
-        "types-cryptography>=3.3.23.2",
-        "types-jsonschema>=4.26.0.20260202",
-]
 dev = [
-  {include-group = "dev-tools"},
+  "ruff>=0.15.4",
+  "pytest>=8.4.2",
+  "mypy>=1.19.1",
+  "types-pyyaml>=6.0.12.20250915",
+  "types-cryptography>=3.3.23.2",
+  "types-jsonschema>=4.26.0.20260202",
   # For https://github.com/azimuth-cloud/pyhelm3/pull/36 & https://github.com/azimuth-cloud/pyhelm3/pull/37 & https://github.com/azimuth-cloud/pyhelm3/pull/39
   "pyhelm3 @ git+https://github.com/element-hq/pyhelm3.git@efa509e7501edeb9d04f2a9e688f6e0088305954",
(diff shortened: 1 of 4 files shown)


.github/workflows/scripts-linting.yml
```diff
@@ -70,4 +69,4 @@ jobs:
       with:
         persist-credentials: false
 
-    - uses: streetsidesoftware/cspell-action@9cd41bb518a24fefdafd9880cbab8f0ceba04d28  # v8.3.0
+    - uses: streetsidesoftware/cspell-action@9cd41bb518a24fefdafd9880cbab8f0ceba04d28  # v8
```
benbz: Revert me

tests/integration/test_synapse.py
```diff
@@ -180,9 +180,10 @@ async def make_request_and_assert_backend_used(path, backend, ssl_context, logs_
 
             attempts += 1
             await asyncio.sleep(1)
+        newline = chr(10)
```
benbz, line 183: Why are we doing this?

tests/manifests/test_configs_consistency.py
```diff
@@ -643,14 +643,15 @@ def check_paths_used_in_content(self):
                         break
                 else:
                     paths_not_found.append((str(MountPath(parent_mount, mount_node)), source))
+        newline = chr(10)
```
benbz, line 646: Why are we doing this?

pyproject.toml
```diff
@@ -10,49 +10,56 @@ version = "0.1.0"
...
-        "lightkube>=0.19.1",
-        "pytz>=2025.2",
-        "cryptography>=46.0.5",
-        "platformdirs>=4.9.2",
-        "python-on-whales>=0.80.0",
-        "aiohttp>=3.13.3",
-        "yamale>=6.1.0",
-        "typer>=0.24.1",
-        "pytest>=8.4.2",
+# This should be kept in sync with packages/ess-migration-tool
```
benbz: This feels fraught. Is there no way to get renovate to manage them both? Or to have `uv` manage them both in a single place

tests/integration/test_synapse.py
gaelgatelement: Because python 3.11 does not support `'\n'` in format strings

pyproject.toml
benbz: I see the comment has gone but both versions are currently in sync. Is that deliberate/accidental and required to carry on being the case or not?

gaelgatelement: The explicit dependencies are deliberate - they are dev dependencies of the root package, but runtime dependencies of the ess-migration-tool. I expect that they will be maintained properly by dependabot

benbz: What happens if dependabot doesn't keep the versions in sync? Or someone edits one place but not the other and they're out of sync until the next dependabot run?

PR #1142, opened 2026-03-17
ci: fix matrix-tools release being flaky

gaelgatelement:
I think this is what have been causing flakes with pushing images to the registry.

https://docs.docker.com/reference/cli/docker/buildx/build/#registry

> The registry exporter is a shortcut for type=image,push=true.

https://github.com/docker/bake-action

> push Bool Push is a shorthand for --set=*.output=type=registry (default false)

So according to this, I think we should not set output type in our docker bake action.

### DIFF
--- a/.github/workflows/matrix-tools.yml
+++ b/.github/workflows/matrix-tools.yml
@@ -126,7 +126,6 @@ jobs:
           cwd://${{ steps.meta.outputs.bake-file }}
         targets: matrix-tools
         set: |
-          *.output=type=image
           *.platform=linux/amd64,linux/arm64
         # https://docs.zizmor.sh/audits/#cache-poisoning
         # This is not catched by zizmor but arguably we need to disable cache for docker builds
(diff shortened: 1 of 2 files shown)


.github/workflows/matrix-tools.yml
```diff
@@ -126,7 +126,6 @@ jobs:
           cwd://${{ steps.meta.outputs.bake-file }}
         targets: matrix-tools
         set: |
-          *.output=type=image
```
benbz, line 129: Do we need / want to include this if `github.ref_type != 'tag'` ?

gaelgatelement: I dont think so. Current CI does the build but ends up with a warning : 

> WARNING: No output specified for matrix-tools target(s) with docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load

Which I think is fine

https://github.com/element-hq/ess-helm/actions/runs/23200708648/job/67423209951?pr=1142

benbz (approved):

PR #1149, opened 2026-03-18
Migration Tool : Support PermissionError

gaelgatelement:
Fixes : 

```
ess-migration-tool --synapse-config /etc/synapse/homeserver.yaml 
🚀 Starting Matrix Stack to ESS Migration
📦 Step 0/4 (0%): Loading and validating input files
📦 Step 1/4 (25%): Migrating configuration to ESS values
❌ Migration failed!
💥 Error: [Errno 13] Permission denied: '/etc/synapse/legacy.localhost.signing.key'
📚 Check logs for details and try again.
ERROR:root:Migration failed: [Errno 13] Permission denied: '/etc/synapse/legacy.localhost.signing.key'

```

### DIFF
--- a/packages/ess-migration-tool/newsfragments/1149.misc.md
+++ b/packages/ess-migration-tool/newsfragments/1149.misc.md
@@ -1,0 +1,1 @@
+Prepare `ess-migration-tool` for pypi publishing.
--- a/packages/ess-migration-tool/src/ess_migration_tool/extra_files.py
+++ b/packages/ess-migration-tool/src/ess_migration_tool/extra_files.py
@@ -406,6 +406,9 @@ class ExtraFilesDiscovery:
         if self._is_binary_file(file_path):
             extra_file.cleartext = False
 
-        with open(file_path):
-            extra_file.content = file_path.read_bytes()
+        try:
+            with open(file_path):
+                extra_file.content = file_path.read_bytes()
+        except PermissionError as err:
+            raise ExtraFilesError(f"Permission denied when reading file: {file_path}") from err
         return extra_file
--- a/packages/ess-migration-tool/src/ess_migration_tool/secrets.py
+++ b/packages/ess-migration-tool/src/ess_migration_tool/secrets.py
@@ -76,6 +76,8 @@ class SecretDiscovery:
                         logging.info(f"Found file value for {secret_key}")
                     except FileNotFoundError:
                         logger.warning(f"File not found: {file_path}")
+                    except PermissionError:
+                        logger.warning(f"Permission denied when reading file: {file_path}")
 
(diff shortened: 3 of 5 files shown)


gaelgatelement:
> Does this need to detect if any critical files could not be migrated?
> 
> IIRC changing the signing key when migrating a Synapse homeserver breaks federation ([related docs](https://element-hq.github.io/synapse/latest/usage/administration/backups.html#server-signing-key)). So maybe the migration should warn if it could not be found / suggest how it may be (securely) migrated manually.



> Does this need to detect if any critical files could not be migrated?
> 
> IIRC changing the signing key when migrating a Synapse homeserver breaks federation ([related docs](https://element-hq.github.io/synapse/latest/usage/administration/backups.html#server-signing-key)). So maybe the migration should warn if it could not be found / suggest how it may be (securely) migrated manually.

So this is what it looks like : 

```
 ess-migration-tool --synapse-config /etc/synapse/homeserver.yaml 
🚀 Starting Matrix Stack to ESS Migration
📦 Step 0/4 (0%): Loading and validating input files
📦 Step 1/4 (25%): Migrating configuration to ESS values

============================================================
🔐 SYNAPSE SECRETS REQUIRED FOR MIGRATION
============================================================
The following SYNAPSE secrets are required but could not be automatically
discovered from your configuration files. Please provide them:
📝 Signing key
   Secret path: synapse.signingKey
   Please paste the secret value: 
```

The user cannot skip this : 

```
   Please paste the secret value: 
   ❌ Value cannot be empty. Please try again.
   Please paste the secret value: 
   ❌ Value cannot be empty. Please try again.
   Please paste the secret value: 
   ❌ Value cannot be empty. Please try again.
   Please paste the secret value: 
```

I think we might want to double-check the content of the signing key, but that would be the role of another PR.

gaelgatelement:
To expand a bit more on the explanations here : 

 - The script does not skip this secret because it has [init_if_missing_from_source_cfg=False](https://github.com/element-hq/ess-helm/blob/1f05d65ffef2058804822400ded98077f7460d6e/packages/ess-migration-tool/src/ess_migration_tool/synapse.py#L217). This is set to False on purpose because, as you said, we **do not want** to automatically generate a new signing key 
- On the contrary, the script would continue and jsut warn that the registration shared secret will be re-generated. It has [init_if_missing_from_source_cfg=True](https://github.com/element-hq/ess-helm/blob/1f05d65ffef2058804822400ded98077f7460d6e/packages/ess-migration-tool/src/ess_migration_tool/synapse.py#L210) and we do not consider it critical if it breaks external script relying on this shared secret. Especially since the user can always figure out where the issue is, and restore the previous secret later on.

PR #1150, opened 2026-03-19
migration: ignore synapse pid file

gaelgatelement:
This should not be imported.

### DIFF
--- a/packages/ess-migration-tool/src/ess_migration_tool/synapse.py
+++ b/packages/ess-migration-tool/src/ess_migration_tool/synapse.py
@@ -250,6 +250,7 @@ class SynapseExtraFileDiscovery(ExtraFilesDiscoveryStrategy):
         return [
             "media_store_path",  # Synapse media store path should be ignored
             "log_config",  # Log configuration is handled by the chart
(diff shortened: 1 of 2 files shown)


packages/ess-migration-tool/src/ess_migration_tool/synapse.py
```diff
@@ -250,6 +250,7 @@ def ignored_config_keys(self) -> list[str]:
         return [
             "media_store_path",  # Synapse media store path should be ignored
             "log_config",  # Log configuration is handled by the chart
+            "pid_file",  # We do not care of importing pid files
```
AndrewFerr, line 253: This can be ignored because containerized deployments of Synapse don't need a PID file, right?

gaelgatelement: > This can be ignored because containerized deployments of Synapse don't need a PID file, right?

Yes and no. This part of the code is here to include the content pointed by the path of these config paths in a ConfigMap that will be made availables to pods in clusters. In this regards, we do not need to import any pid file.

PR #1155, opened 2026-03-19
Migration: Allow the user to use ESS postgres database

gaelgatelement:
This allows the user to decide between using their existing postgres database, or migrate everything to ESS-managed databases.

### DIFF
--- a/packages/ess-migration-tool/src/ess_migration_tool/engine.py
+++ b/packages/ess-migration-tool/src/ess_migration_tool/engine.py
@@ -38,8 +38,12 @@ class MigrationEngine:
     def __post_init__(self) -> None:
         """Initialize the migration engine."""
         for migration, secret_discovery_strategy, extra_file_strategy in [
-            (SynapseMigration(self.global_options), SynapseSecretDiscovery(), SynapseExtraFileDiscovery()),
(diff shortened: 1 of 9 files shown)


packages/ess-migration-tool/src/ess_migration_tool/__main__.py
```diff
@@ -313,6 +332,144 @@ def main() -> int:
         pretty_logger.info("📚 For more details on deployment and data migration, refer to the ESS documentation.")
         pretty_logger.info("")
 
+        # Add database-specific instructions
+        if not engine.global_options.use_existing_database:
```
AndrewFerr, line 340: Is this meant to cover the `None` / "unselected" case too?  Or is it guaranteed to have a choice by this point?

If not:
```suggestion
        if engine.global_options.use_existing_database is False:
```

gaelgatelement: > Is this meant to cover the None / "unselected" case too? Or is it guaranteed to have a choice by this point?

We are guaranteed to have a choice by this point due to ` self.prompt_for_database_choice()` in the migration engine. But I agree it's not very clear when reading at the code here, I'll see for moving it here.

PR #1217, opened 2026-04-07
Improve credential schema

benbz:
Previously we always setup empty dictionaries for each credential. This meant we couldn't do strict schema validation that one of `value` or `secret`+`secretKey` is present. We had a helper (sometimes) present to check for some of these issues but it wasn't consistently applied either.

~~Draft as exploring whether we want to do this or not~~ I think this is worthwhile as more things can be validate by the schema and it hasn't imposed too much change on the chart itself.


jaywink (approved):
This feels cleaner 👍

PR #1238, opened 2026-04-17
Fix formatting error in DEVELOPERS.md

AndrewFerr:


benbz (approved):
Doesn't need a changelog fragment IMO

PR #1242, opened 2026-04-20
Improve k3s docs and tests around X-Forwarded-For

benbz:
Fixes #1216

The Trusted IP CIDRs come from https://docs.k3s.io/networking/basic-network-options. We don't technically need for this Traefik in k3d in CI as we use `externalTrafficPolicy: Local` but it doesn't hurt either

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,19 +211,50 @@ chown "$USER:$USER" "$KUBECONFIG"
 
 3. Add `export KUBECONFIG=~/.kube/config` to `~/.bashrc` to make it persistent
 
-4. Install Helm, the Kubernetes Package Manager. You can use your [OS repository](https://helm.sh/docs/intro/install/) or call the following command:
+4. Customise Traefik by creating the file `traefik-config.yaml` in `/var/lib/rancher/k3s/server/manifests`
+
+```yaml
+apiVersion: helm.cattle.io/v1
+kind: HelmChartConfig
+metadata:
+  name: traefik
+  namespace: kube-system
+spec:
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -211,19 +211,50 @@ chown "$USER:$USER" "$KUBECONFIG"
 
 3. Add `export KUBECONFIG=~/.kube/config` to `~/.bashrc` to make it persistent
 
-4. Install Helm, the Kubernetes Package Manager. You can use your [OS repository](https://helm.sh/docs/intro/install/) or call the following command:
+4. Customise Traefik
+
```
gaelgatelement, line 215: We probably need to explain where this file should be created?

benbz: Done

tests/integration/fixtures/cluster.py
```diff
@@ -146,12 +146,6 @@ async def ingress(cluster, kube_client: AsyncClient):
             assert service
             assert service.spec
             assert service.spec.clusterIP
-
-            # So that we see the real connecting IP
-            # As per https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/#preserving-the-client-source-ip
-            service.spec.externalTrafficPolicy = "Local"
-            await kube_client.patch(Service, name="traefik", namespace="kube-system", obj=service)
```
gaelgatelement, line 153: Should we add a test assertion that we see the proper IP in Synapse logs ? (or at least, not the internal IP) ?

gaelgatelement (approved):

PR #1243, opened 2026-04-20
ESS Migration: Rename "component_root_key" to strategy name

gaelgatelement:
- A strategy does not necessarily manage only 1 component root key so the attribute name is misleading
- In the future we are going to have strategies managing multiple values of the chart

### DIFF
--- a/packages/ess-migration-tool/src/ess_migration_tool/synapse.py
+++ b/packages/ess-migration-tool/src/ess_migration_tool/synapse.py
@@ -46,6 +46,8 @@ worker_types = [
     "user-dir",
 ]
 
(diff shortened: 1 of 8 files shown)


packages/ess-migration-tool/src/ess_migration_tool/synapse.py
```diff
@@ -331,10 +333,6 @@ def transformations(self) -> list[TransformationSpec]:
                 )
             ]
 
-    @property
-    def component_config_extras(self) -> dict[str, Any]:
```
benbz, line 335: Why is this disappearing?

gaelgatelement:
This has been superseded by https://github.com/element-hq/ess-helm/pull/1248

PR #1244, opened 2026-04-20
Default Synapse URL previews to on

benbz:
cc @Half-Shot 

The render-config process appends to arrays by default. This is desirable behaviour here as emptying the denylist is non-desirable. Adding specific ranges to the allowlist is much better

### DIFF
--- a/charts/matrix-stack/configs/synapse/synapse-01-shared-underrides.yaml.tpl
+++ b/charts/matrix-stack/configs/synapse/synapse-01-shared-underrides.yaml.tpl
@@ -19,6 +19,28 @@ federation_client_minimum_tls_version: '1.2'
 experimental_features:
   msc4028_push_encrypted_events: true
 
+url_preview_enabled: true
+# Both these lists are append only. If there are specific ranges within the denylist
+# that should be allowed, then add them to the allowlist.
+url_preview_ip_range_whitelist: []
+url_preview_ip_range_blacklist:
+- '192.168.0.0/16'
+- '100.64.0.0/10'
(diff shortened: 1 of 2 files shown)


Half-Shot (approved):
LGTM!

charts/matrix-stack/configs/synapse/synapse-01-shared-underrides.yaml.tpl
```diff
@@ -19,6 +19,28 @@ federation_client_minimum_tls_version: '1.2'
 experimental_features:
   msc4028_push_encrypted_events: true
 
+url_preview_enabled: true
+# Both these lists are append only. If there are specific ranges within the denylist
+# that should be allowed, then add them to the allowlist.
+url_preview_ip_range_whitelist: []
```
Half-Shot, line 25: Empty means we allow all not in the blacklist?

benbz: Empty means no ranges in the deny list are allowed. https://element-hq.github.io/synapse/latest/usage/configuration/config_documentation.html?#url_preview_ip_range_whitelist
> (array) This option sets a list of IP address CIDR ranges that the URL preview spider is allowed to access even if they are specified in url_preview_ip_range_blacklist. This is useful for specifying exceptions to wide-ranging blacklisted target IP ranges – e.g. for enabling URL previews for a specific private website only visible in your network. Defaults to [].

gaelgatelement (approved):

PR #1245, opened 2026-04-20
Upgrade MatrixRTC authoriser to v0.4.4

benbz:
https://github.com/element-hq/lk-jwt-service/releases/tag/v0.4.4


newsfragments/1245.changed.md
```diff
@@ -0,0 +1,10 @@
+Upgrade MatrixRTC Authoriser to v0.4.4.
```
jaywink, line 1: We're going to filter out the other changelog before release?

Is there a reason we don't just delete 1237 now? Or at least feels odd to update it, since 1237 did not update the version.

jaywink (approved):
Q about changelogs, otherwise good to go!

newsfragments/1245.changed.md
benbz: By doing it this way the release notes link to both PRs with a single changelog text

PR #1246, opened 2026-04-21
ESS Migration: Implement `additional` config as a TransformationSpec transformer

gaelgatelement:
`additional` is now handled with a dedicated transformer for the attributes with additional configuration. This allows us to support component without any `additional` values, or to have other names for additional value key path.

- The first argument of transformers is changed from `logging.Logger` to the `ConfigValueTransformer` using the transformer in order to access tracked values in additional transformer
- Add support for kwargs to `TransformationSpec` transformers
- This allows us to pass arbitrary named arguments to transformers. In the case of `additional`, this contains `component_root_key` and the `extra_file_discovery` service

Built with Mistral Devstral 2.


packages/ess-migration-tool/src/ess_migration_tool/mas.py
```diff
@@ -91,8 +94,9 @@ def filter_mas_listeners(pretty_logger: logging.Logger, listeners: list[dict] |
     returning them as a dictionary structure for additional config.
 
     Args:
-        pretty_logger: Logger for user-friendly output
+        config_value_transformer: ConfigValueTransformer instance
```
benbz: Unused?

packages/ess-migration-tool/src/ess_migration_tool/synapse.py
```diff
@@ -170,8 +186,9 @@ def filter_listeners(pretty_logger: logging.Logger, listeners: list[dict] | None
     returning them as a dictionary structure for additional config.
 
     Args:
-        pretty_logger: Logger for user-friendly output
+        config_value_transformer: ConfigValueTransformer instance
```
benbz: Unused?

benbz (approved):

PR #1248, opened 2026-04-21
ESS Migration: Refactor to drop component root key from strategies

gaelgatelement:
This removes `component_root_key` from strategies so that a given strategy can manage any component

Only the `ExtraFileDiscovery` strategy needs a `component_root_key` to be able to figure the key where to inject `extraVolumes` and `extraVolumeMounts`.

Strategies now have a name, we refer to this name through constants.

The transformer which builds the additional section is now specific to each component through a closure passing in component-specific parameters.

Built with Mistral Devstral 2.

### DIFF
--- a/packages/ess-migration-tool/src/ess_migration_tool/migration.py
+++ b/packages/ess-migration-tool/src/ess_migration_tool/migration.py
@@ -50,12 +50,14 @@ def additional_config_transformer(
         **kwargs: Context parameters. See below.
 
     kwargs uses:
-    - component_root_key: str containing the root key of the component
     - extra_files_discovery: ExtraFilesDiscovery | None containing the extra files discovery service
     """
     source_config = value  # value is the source config (when src_key is None)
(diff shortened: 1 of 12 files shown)


packages/ess-migration-tool/src/ess_migration_tool/migration.py
```diff
@@ -63,11 +68,23 @@ def additional_config_transformer(
     for source_path in config_value_transformer.tracked_values:
         remove_nested_value(filtered_config, source_path)
 
+    # Override checking - moved from MigrationService._check_overrides
```
benbz: Don't think we need this. It doesn't help us after this PR is merged

benbz (approved):

packages/ess-migration-tool/src/ess_migration_tool/migration.py
gaelgatelement: Good catch, that's one painful tendency of LLMs to track changes as comment.

PR #1256, opened 2026-04-24
Support wildcard in secrets handling

gaelgatelement:
This helps us support a bit more complex structures where the secret path is discovered dynamically in an array.

Built with Mistral devstral 2.

### DIFF
--- a/packages/ess-migration-tool/tests/test_nested_values.py
+++ b/packages/ess-migration-tool/tests/test_nested_values.py
@@ -219,6 +219,11 @@ def test_find_matching_schema_key_wildcard_match():
     assert find_matching_schema_key("a.999.c", schema) == "a.*.c"
 
 
+def test_find_matching_schema_key_wildcard_no_match():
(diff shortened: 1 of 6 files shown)


packages/ess-migration-tool/tests/test_nested_values.py
```diff
@@ -111,3 +118,120 @@ def test_remove_nested_value_empty_config():
...
+
+# Tests for find_matching_schema_key
+
+
+def test_find_matching_schema_key_exact_match():
+    schema = {"a.b.c": "config"}
+    assert find_matching_schema_key("a.b.c", schema) == "a.b.c"
+
+
+def test_find_matching_schema_key_wildcard_match():
```
benbz, line 217: Can we have a negative version of this test, in the same way that `test_find_matching_schema_key_no_match` is the negative version of `test_find_matching_schema_key_exact_match`

benbz (approved):

PR #1262, opened 2026-04-29
Enable url previews by default in Element Web for unencrypted messages

Half-Shot:
Ref https://github.com/element-hq/element-web/blob/develop/apps/web/src/settings/Settings.tsx#L1122


benbz (approved):

benbz (changes requested):
Only applies with MAS

benbz (approved):

PR #1263, opened 2026-04-29
Default homeserver usage statistics to on

benbz:
Draft as awaiting product decision


benbz:
As discussed just now, we're not progressing this PR in this form at this time

PR #1268, opened 2026-04-30
Migrate Element Web config.json to ESS Community

gaelgatelement:
This adds support for Element Web configuration migration to ESS Community.

### DIFF
--- a/packages/ess-migration-tool/src/ess_migration_tool/element_web.py
+++ b/packages/ess-migration-tool/src/ess_migration_tool/element_web.py
@@ -32,7 +32,14 @@ class ElementWebMigration(MigrationStrategy):
 
     @property
     def override_configs(self) -> set[str]:
-        """Config keys that are managed by ESS and should be filtered out."""
+        """Config keys that are managed by ESS (users should not override these)."""
+        # Element Web uses underrides (defaults) rather than true overrides.
+        # All configs are ESS defaults that users can override via additional config.
+        return set()
+
+    @property
+    def underride_configs(self) -> set[str]:
+        """Config keys that are ESS defaults (users can override these via additional config)."""
         return {
             "setting_defaults.urlPreviewsEnabled",
             "setting_defaults.UIFeature.registration",
@@ -71,6 +78,7 @@ class ElementWebMigration(MigrationStrategy):
                 value,
                 component_root_key=ELEMENT_WEB_COMPONENT_ROOT_KEY,
                 override_configs=self.override_configs,
+                underride_configs=self.underride_configs,
                 component_name=ELEMENT_WEB_STRATEGY_NAME,
                 serialization_format="json",
                 use_file_object_format=False,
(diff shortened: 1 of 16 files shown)


packages/ess-migration-tool/src/ess_migration_tool/element_web.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def __init__(self, global_options: GlobalOptions):
+        self.global_options = global_options
+
+    @property
+    def name(self) -> str:
+        return ELEMENT_WEB_STRATEGY_NAME
+
+    @property
+    def override_configs(self) -> set[str]:
```
benbz, line 34: All ESS EW settings are underrides, so I don't think I agree with this

packages/ess-migration-tool/src/ess_migration_tool/element_web.py
```diff
@@ -0,0 +1,174 @@
...
+                required=False,
+            ),
+        ]
+
+    @property
+    def component_config_extras(self) -> dict[str, Any]:
+        return {"enabled": True}
+
+
+class ElementWebSecretDiscovery(SecretDiscoveryStrategy):
```
benbz: Would it be better to add a stub/no-op `SecretDiscoveryStrategy` to handle e.g. Well Knowns too?

packages/ess-migration-tool/src/ess_migration_tool/element_web.py
```diff
@@ -0,0 +1,174 @@
...
+            source_file: Source configuration file name
+            config_data: Element Web configuration data
+
+        Returns:
+            Tuple of (empty dict, empty list) - no specialized secret discovery for Element Web
+        """
+        return ({}, [])
+
+
+class ElementWebExtraFileDiscovery(ExtraFilesDiscoveryStrategy):
```
benbz: Same question about stun/no-op here too

packages/ess-migration-tool/src/ess_migration_tool/element_web.py
gaelgatelement: That's true, I guess we can just silently ignore the config keys for ElementWeb then.

packages/ess-migration-tool/src/ess_migration_tool/element_web.py
gaelgatelement: I think the correct approach here is to make SecretDiscovery optional, like `ExtraFileDiscovery`.

packages/ess-migration-tool/src/ess_migration_tool/element_web.py
gaelgatelement: It's going to be useful as a generic ExtraFileDiscovery mechanism indeed, so let's add it.

packages/ess-migration-tool/src/ess_migration_tool/element_web.py
gaelgatelement: I went ahead and implemented a new list for underrides. I'll update the list for Synapse & MAS in a distinct PR.

PR #1273, opened 2026-05-04
Add underride support in Synapse & MAS configuration

gaelgatelement:


packages/ess-migration-tool/src/ess_migration_tool/mas.py
```diff
@@ -225,9 +225,12 @@ def override_configs(self) -> set[str]:
     @property
     def underride_configs(self) -> set[str]:
         """Config keys that are ESS defaults (users can override these via additional config)."""
-        # MAS primarily uses overrides (ESS-managed values).
-        # No underride configs currently defined.
-        return set()
+        return {
```
benbz, line 228: I feel like I don't understand `underride_configs` in the migration tool. https://github.com/element-hq/ess-helm/pull/1275/changes appears to set Hookshot underride to ESS Community values properties that happen to control config that users can adjust. This is MAS config that can be changed via MAS additional, no direct values file support

gaelgatelement: Yeah, it's totally wrong indeed 🤦  Adjusted for Hookshot, will adjust for MAS&Synapse properly as well in a dedicated PR.

gaelgatelement: So this was incorrect in Hookshot but is correct in MAS/Synapse/ Underrides contains value that the user-provided additional can use to customize.

PR #1274, opened 2026-05-04
Well-Known migration support

gaelgatelement:
Add support for migrating Well Known delegation configuration files (client.json, server.json, support.json) to ESS. 

Users can now specify well-known files via `--well-known-dir` for a directory containing the files, or individual files via `--well-known-client`, `--well-known-server`, and `--well-known-support`.

 Files can be with or without the `.json` extension.

### DIFF
--- a/packages/ess-migration-tool/src/ess_migration_tool/inputs.py
+++ b/packages/ess-migration-tool/src/ess_migration_tool/inputs.py
@@ -17,6 +17,7 @@ from typing import Any
 import yaml
 
 from .models import MigrationError, MigrationInput
+from .well_known import WELL_KNOWN_FILE_PATTERNS, WELL_KNOWN_STRATEGY_NAMES
 
 logger = logging.getLogger("migration")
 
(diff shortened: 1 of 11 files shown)


packages/ess-migration-tool/src/ess_migration_tool/inputs.py
```diff
@@ -177,3 +178,85 @@ def load_migration_input(
...
+                    results.append((strategy_name, arg_path, config))
+                except Exception as e:
+                    logger.warning(f"Failed to load {wk_type} file {arg_path}: {e}")
+
+        # Load from directory (lower precedence, only if not already loaded via CLI arg)
+        if dir_path:
+            dir_path_obj = Path(dir_path)
+            loaded_strategies: set[str] = {r[0] for r in results}
+
+            for wk_type, file_names in WELL_KNOWN_FILE_PATTERNS.items():
```
benbz, line 253: Do we need any error handling if no files match this pattern?

packages/ess-migration-tool/src/ess_migration_tool/well_known.py
```diff
@@ -0,0 +1,107 @@
...
+            # Enable well-known delegation (harmless if set multiple times)
+            TransformationSpec(
+                src_key=None,
+                target_key="wellKnownDelegation.enabled",
+                transformer=lambda *_, **__: True,
+            ),
+        ]
+
+        # For client well-known type, extract server_name and base_url
+        if self.well_known_type == "client":
```
benbz, line 69: Do we want to do a similar thing with the server well-known for the Synapse Ingress?

packages/ess-migration-tool/tests/test_main_e2e.py
```diff
@@ -1008,4 +1009,67 @@ def test_main_e2e_synapse_with_element_web(
...
+    values = yaml.safe_load(values_file.read_text())
+
+    assert values["wellKnownDelegation"]["enabled"] is True
+    assert "additional" in values["wellKnownDelegation"]
+
+    client_config = json.loads(values["wellKnownDelegation"]["additional"]["client"])
+    assert client_config["m.homeserver"]["base_url"] == "https://matrix.example.com"
+
+    server_config = json.loads(values["wellKnownDelegation"]["additional"]["server"])
+    assert server_config["m.server"] == "test.example.com:443"
```
benbz: Neither of these should be in additional if Synapse is configured (and it appears to be)

gaelgatelement: Good catch, it's a sign of missing a `TransformationSpec` !

packages/ess-migration-tool/src/ess_migration_tool/inputs.py
benbz: Sorry not seeing where this happens?

gaelgatelement: Above, see https://github.com/element-hq/ess-helm/pull/1274/changes/07992044c2261045e95601674862ad88769a133b

benbz (approved):

PR #1290, opened 2026-05-11
Add msc4452_enabled to experimental_features.

Half-Shot:
Uses https://github.com/element-hq/synapse/pull/19715 (and will be in Synapse v1.154.0).

We've evaluated the risks to clients and this shouldn't pose any problems, it just changes a status code for when the feature is off and adds an extra capability for Element Web to view.


benbz (approved):
LGTM. While I suspect operationally this is safe to merge now, we don't want to do this yet given the changelog won't be true until Synapse v1.154.0 is presesnt here

benbz:
#1358 will do the Synapse upgrade

PR #1299, opened 2026-05-12
Deep copy remove redundant deepcopy in update_paths_in_config

gaelgatelement:


benbz (approved):
This makes me uncomfortable but I'd hope the tests would catch issues

gaelgatelement:
> This makes me uncomfortable but I'd hope the tests would catch issues

Well, your comment made me double-check to prove the redundant copy, and it turns-out filtered config was just dead code - everything is in `additional_config_transformer` now.

I fixed this PR content to adjust to the actual redundant `deepcopy`.

PR #1301, opened 2026-05-13
pin pytest 8 for now

gaelgatelement:
We are seeing issues with pytest 9. Let's pin it for now.


benbz (approved):
Are there no changes required to lock file to encoded the new constraint?

benbz:
> Are there no changes required to lock file to encoded the new constraint?

Ha! I didn't realise `no-changes-after-building` would detect this

PR #1306, opened 2026-05-13
Synapse: Discover MAS properties from configuration

gaelgatelement:
Trigger a values conflict resolution if Synapse `homeserver.yaml` does not have `matrix_authentication_service` configured despite a MAS `config.yaml` being passed.

Adding the test discovered a secrets duplication issue that is being fixed in https://github.com/element-hq/ess-helm/pull/1313

### DIFF
--- a/packages/ess-migration-tool/src/ess_migration_tool/synapse.py
+++ b/packages/ess-migration-tool/src/ess_migration_tool/synapse.py
@@ -345,6 +345,12 @@ class SynapseMigration(MigrationStrategy):
                 transformer=filter_listeners,
                 required=False,
             ),  # Filter out chart-managed listeners and output to additional config
+            TransformationSpec(
+                src_key="matrix_authentication_service",
+                target_key="matrixAuthenticationService.enabled",
+                transformer=lambda _, mas, **__: mas.get("enabled") if mas else None,
+                required=False,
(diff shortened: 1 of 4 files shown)


packages/ess-migration-tool/src/ess_migration_tool/synapse.py
```diff
@@ -351,6 +351,12 @@ def synapse_additional_transformer(
                 transformer=synapse_additional_transformer,
                 required=False,
             ),  # Generic additional config generation (src_key=None passes full config)
+            TransformationSpec(
+                src_key="matrix_authentication_service",
```
benbz, line 349: I'm confused as why this is in the Synapse file?

gaelgatelement: The main reason is to remove it from the `additional` section. But it also allows us to make sure that the Synapse `homeserver.yaml` and MAS `config.yaml` are consistent together. Typically, if `config.yaml` is passed but Synapse `homeserver.yaml` does not have `matrix_authentication_service` configured, it will trigger a values conflict resolution on `matrixAuthenticationService.enabled`. This should allow the user to stop here and investigate what's going on.

benbz (approved):

PR #1309, opened 2026-05-15
Document traffic policies on exposed port ranges

benbz:
They're supported in the schema, tests and templates, just missing this documentation


benbz:
I think the change from `<nil>` to `Cluster` is fine as `Cluster` is the default

gaelgatelement (approved):

PR #1313, opened 2026-05-18
Deduplicate discovered secrets when multiple strategies can discover it.

gaelgatelement:
This fixes a minor issue where `synapseSharedSecret` value was rendered in 2 secrets (MAS & Synapse), despite being referred only once in the values file.


packages/ess-migration-tool/src/ess_migration_tool/models.py
```diff
@@ -248,3 +255,43 @@ def get_strategies_for_secret(self, secret_key: str) -> list[str]:
...
+        # If secret was not discovered, return None
+        if not self.is_discovered(secret_key):
+            return None
+
+        # Get all sources (strategies that discovered this secret)
+        sources = self.sources.get(secret_key, [])
+
+        # Find the strategy with takes_precedence=True
+        owning_strategy: str | None = None
+        for source in sources:
```
benbz, line 287: Also check for no source taking precedence if multiple sources?

benbz (approved):

PR #1370, opened 2026-06-02
Migration fixtures : Use frozendict

gaelgatelement:
- Instead of calling `deepcopy` everywhere, let's use `frozendict` 
- We also now have a single `write_config` fixture

### DIFF
--- a/packages/ess-migration-tool/tests/conftest.py
+++ b/packages/ess-migration-tool/tests/conftest.py
@@ -17,8 +17,13 @@ from cryptography.hazmat.primitives import serialization
 from cryptography.hazmat.primitives.asymmetric import dsa, ec, rsa
 from ess_migration_tool.migration import ConfigValueTransformer
 from ess_migration_tool.models import GlobalOptions, ValueSourceTracking
+from frozendict import frozendict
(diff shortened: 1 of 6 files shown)


packages/ess-migration-tool/tests/conftest.py
```diff
@@ -293,23 +285,9 @@ def basic_mas_config_with_keys(tmp_path, basic_mas_config):
...
-
-    def _write_config(config_data):
-        mas_config_file = tmp_path / "mas.yaml"
-        with open(mas_config_file, "w") as f:
-            yaml.dump(config_data, f)
-        return mas_config_file
-
-    return _write_config
+    return frozendict(basic_mas_config) | {
+        "secrets": {"keys_dir": str(keys_dir), "encryption": basic_mas_config["secrets"]["encryption"]}
```
benbz: Why is `encryption` appearing here?

gaelgatelement: Because `secrets` is being overridden by the `|`. I'll do it differently.

benbz (approved):

PR #1385, opened 2026-06-09
Well Known: Import MAS ingress from client properties

gaelgatelement:
This will help discover any issuer if well-known delegation was misconfigured.

### DIFF
--- a/packages/ess-migration-tool/tests/test_main_e2e.py
+++ b/packages/ess-migration-tool/tests/test_main_e2e.py
@@ -409,7 +409,7 @@ def test_main_e2e_synapse_with_mas(
             else:
                 pytest.fail(f"Unexpected secret file: {secret_file.name}")
     mas_additional_config = yaml.safe_load(mas_config["additional"]["00-imported.yaml"]["config"])
-    assert "keys_dir" not in mas_additional_config["secrets"]
+    assert "secrets" not in mas_additional_config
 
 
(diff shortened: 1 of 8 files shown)


packages/ess-migration-tool/tests/test_main_e2e.py
```diff
@@ -1088,10 +1092,10 @@ def test_well_known_migration(
...
     client_config = json.loads(values["wellKnownDelegation"]["additional"]["client"])
     # m.homeserver.base_url and m.homeserver.server_name should be filtered out
-    assert "m.homeserver" not in client_config or (
-        "base_url" not in client_config.get("m.homeserver", {})
-        and "server_name" not in client_config.get("m.homeserver", {})
-    )
+    assert "base_url" not in client_config.get("m.homeserver", {})
+    assert "server_name" not in client_config.get("m.homeserver", {})
+    assert "account" not in client_config.get("org.matrix.msc2965.authentication", {})
+    assert "issuer" not in client_config.get("org.matrix.msc2965.authentication", {})
```
benbz: Arguably I think we should be assert `"m.homeserver" not in client_config` and `"org.matrix.msc2965.authentication" not in client_config`

gaelgatelement: I tried hard to find any reason *not* to do that, but they all felt rather pointless :laughing: This is now implemented in the community tool.

benbz (approved):

PR #1388, opened 2026-06-09
Refactor secret conflict resolution to rely on strategy object

gaelgatelement:
It allows us to show the strategy name instead of the secret name.


packages/ess-migration-tool/src/ess_migration_tool/conflicts.py
```diff
@@ -98,68 +111,160 @@ def resolve_value_conflicts(
...
+                )
 
-        selected_value, is_custom = prompt_for_conflict_resolution(
-            summary_logger, secret_key, value_to_strategies, global_options
-        )
+            selected_value, _ = prompt_for_conflict_resolution(
+                summary_logger, secret_key, value_to_strategies, global_options
+            )
 
-        if is_custom:
```
benbz, line 149: Do we not need this anymore?

gaelgatelement: It was actually never called. The `prompt_for_conflict_resolution` never passed `enable_custom=True`, so we never had this case at all for secrets. That's intended, but I missed this `if` branch when I initially implemented it.

benbz (approved):

packages/ess-migration-tool/src/ess_migration_tool/conflicts.py
benbz: Are there any other callers of `prompt_for_conflict_resolution`? If not let's remove the 2nd return value?

PR #1394, opened 2026-06-10
Secrets conflicts resolution: fix debug logs

gaelgatelement:
Noticed a couple of wrong logs while testing the tool in debug mode.

### DIFF
--- a/packages/ess-migration-tool/src/ess_migration_tool/conflicts.py
+++ b/packages/ess-migration-tool/src/ess_migration_tool/conflicts.py
@@ -199,14 +199,14 @@ class DiscoveredSecretTracking:
                 owning_strategy = source.strategy
 
             if source.takes_precedence:
-                found_precedence = True
                 # This strategy wants to own this secret
-                if owning_strategy is not None:
+                if found_precedence:
                     # More than one strategy wants to own it - this is a conflict
                     logger.error(
                         f"Multiple strategies claim precedence for secret {secret_key}: "
-                        f"{owning_strategy} and {source.strategy.name}. "
+                        f"{owning_strategy.name} and {source.strategy.name}. "
                     )
+                found_precedence = True
                 owning_strategy = source.strategy
 
         # If no strategy has takes_precedence,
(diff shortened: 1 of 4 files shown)


packages/ess-migration-tool/src/ess_migration_tool/conflicts.py
```diff
@@ -199,14 +199,14 @@ def get_secret_owner(self, secret_key: str) -> SecretDiscoveryStrategy | None:
...
                 # This strategy wants to own this secret
-                if owning_strategy is not None:
+                if found_precedence:
                     # More than one strategy wants to own it - this is a conflict
                     logger.error(
                         f"Multiple strategies claim precedence for secret {secret_key}: "
-                        f"{owning_strategy} and {source.strategy.name}. "
+                        f"{owning_strategy.name} and {source.strategy.name}. "
                     )
+                found_precedence = True
```
benbz, line 209: I'm struggling to see how this changes anything other than logging. After the `if source.takes_precedence` block `found_precedence` is still `True` and `owning_strategy` has been updated

benbz (commented):
No test is changing or being added?

packages/ess-migration-tool/src/ess_migration_tool/conflicts.py
gaelgatelement: Yes, the logs were totally wrong. It was being repeated for each conflict when a secret was having precedence (since it  was marking `found_precedence: True` before testing its value).

gaelgatelement:
> No test is changing or being added?

Added tests!

packages/ess-migration-tool/src/ess_migration_tool/conflicts.py
benbz: Ack, so the changelog fragment is wrong?

gaelgatelement: I linked it to the same PR as the secret refactoring, as it's related in logging the strategies names. ( https://github.com/element-hq/ess-helm/pull/1393 )

benbz (approved):

PR #1411, opened 2026-06-17
Pin pytest to <9 in the integration tests package

sandhose:
PR #1301 pinned pytest to <9 in the root pyproject.toml / uv.lock, but the ess-community-integration-tests package in tests/pyproject.toml still declared an unbounded `pytest>=8.4.2`.

Downstream consumers (e.g. matrix-authentication-service, synapse) install the suite via `uv tool install git+...#subdirectory=tests`, which resolves fresh from tests/pyproject.toml and ignores the workspace uv.lock — so they pull pytest 9, which breaks pytest-asyncio-cooperative (sync tests depending on async fixtures became a hard error, and the cluster fixture trips the "more than one yield" check).

Apply the same <9 bound to the test package itself.


gaelgatelement (approved):

newsfragments/1411.internal.md
```diff
@@ -0,0 +1 @@
+Pin pytest to `<9` in the `ess-community-integration-tests` package as well, so downstream consumers that `uv tool install` the test package (rather than running from the locked workspace) don't pull pytest 9, which breaks `pytest-asyncio-cooperative`. Completes the pin from #1301.
```
gaelgatelement, line 1: Can you rename the file to `1411.internal.md` ?

sandhose: Done!

PR #1425, opened 2026-06-30
Fix non-breaking whitespace

gaelgatelement:
My editor noticed this non-breaking whitespace.


benbz (approved):
Don't see a need for a changelog entry here

PR #1438, opened 2026-07-06
feat(matrix-stack): add priorityClassName support to component workloads

### LINKED ISSUE #1422: Support setting priorityClass
I want Matrix/Element have a higher prio than certian other containers

I would like to have the option to set the priorityClass. Its set under spec.template.spec.priorityClassName.

I considered using a kustomize patch

**Additional context**
See this Helm chart as example:
Template: https://github.com/cloudnative-pg/charts/blob/cacd294e54dcb87461d315795dda4f735a4c7b54/charts/cloudnative-pg/templates/deployment.yaml#L146
Values: https://github.com/cloudnative-pg/charts/blob/cacd294e54dcb87461d315795dda4f735a4c7b54/charts/cloudnative-pg/values.yaml#L144

somaz94:
Adds an optional `priorityClassName` value to every matrix-stack component workload, mirroring the existing `nodeSelector` pattern. Default is unset so rendered output is unchanged for existing users. Resolves #1422.

Validation (run locally against the repo's own tooling):
- `helm lint` clean; `helm template` renders `priorityClassName` when set, byte-identical to `main` when unset
- Full `pytest tests/manifests` passes (2377 passed), including a new `test_pod_priorityClassName.py` mirroring `test_pod_nodeSelector.py`
- `reuse lint` compliant, `ruff` clean, the assemble scripts reproduce with no drift

This is a per-component knob (like `nodeSelector`). If you'd prefer a global default with per-component overrides (like `tolerations`/`topologySpreadConstraints`), happy to switch.

### DIFF
--- a/tests/manifests/test_pod_priorityClassName.py
+++ b/tests/manifests/test_pod_priorityClassName.py
@@ -1,0 +1,87 @@
+# Copyright 2025 New Vector Ltd
+# Copyright 2025-2026 Element Creations Ltd
+#
+# SPDX-License-Identifier: AGPL-3.0-only
+
+import random
+import string
+
+import pytest
+
+from . import DeployableDetails, PropertyType, values_files_to_test
+from .utils import (
+    iterate_deployables_workload_parts,
+    template_id,
+    template_to_deployable_details,
+)
+
+
+@pytest.mark.parametrize("values_file", values_files_to_test)
+@pytest.mark.asyncio_cooperative
+async def test_pod_has_no_priorityClassName_by_default(templates):
+    for template in templates:
+        if template["kind"] in ["Deployment", "StatefulSet", "Job"]:
+            pod_spec = template["spec"]["template"]["spec"]
+            assert "priorityClassName" not in pod_spec, (
+                f"{template_id(template)} has a default priorityClassName when one isn't configured"
+            )
+
+
+@pytest.mark.parametrize("values_file", values_files_to_test)
+@pytest.mark.asyncio_cooperative
+async def test_pod_gets_configured_priorityClassName(values, make_templates, release_name):
+    def set_priorityClassName(deployable_details: DeployableDetails):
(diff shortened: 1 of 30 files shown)


benbz (commented):
Hi, thanks for this. It looks basically perfect. Ideally we would like a global `priorityClassName`, that's default unset (to avoid changes to existing deploys as you've done for the per-component values) but can be used if people want to change the `priorityClassName` for everything at once.

To document: all Synapse workers use the same `priorityClassName` as the main process. I'm not requesting that as a change, I think that's fine for now and we can add per-worker support later if needed.

somaz94:
Thanks @benbz! Added a global `priorityClassName` (default unset, so existing deploys are unaffected) that all components fall back to, with a per-component value overriding it. Also documented that Synapse workers inherit the main process's `priorityClassName`. Ready for another look.

tests/manifests/test_pod_priorityClassName.py
```diff
@@ -0,0 +1,87 @@
+# Copyright 2025 New Vector Ltd
+# Copyright 2025-2026 Element Creations Ltd
```
benbz:
```suggestion
# Copyright 2026 Element Creations Ltd
```

And then we're good to merge, thanks

somaz94:
Done — updated the copyright header to `# Copyright 2026 Element Creations Ltd`. Thanks @benbz!

benbz:
> Done — updated the copyright header to `# Copyright 2026 Element Creations Ltd`. Thanks @benbz!

As per the suggestion the New Vector Ltd copyright line needed to go too, not just changing the date on the Element Creations Ltd line

somaz94:
Done — dropped the New Vector Ltd copyright line, the header on the new test file is now just `# Copyright 2026 Element Creations Ltd`. Thanks!

benbz (approved):

benbz:
Thank you very much for your contribution!

PR #1442, opened 2026-07-09
feat(matrix-stack): add affinity support to component workloads

### LINKED ISSUE #368: Node Affinity adding
So on my cluster i have nodes with public and private IP. And it will be great if you can add Node affinities to choose by node tag or/and label in values

Just adding it like option will be enough like:

```yaml
elementWeb:
  enabled: true
  additional: {}
  replicas: 1
  image:
    registry: ghcr.io
    repository:
    tag: <WEB_DOCKER_TAG>
    pullSecrets: 
      - name: github-auth
  ingress:
    host: <WEBHOST>
    annotations: {}
    tlsEnabled: true
    service: {}
  labels: {}
  extraEnv: []
  containersSecurityContext:
    allowPrivilegeEscalation: true
    capabilities:
      drop:
        - ALL
    readOnlyRootFilesystem: true
  podSecurityContext:
    fsGroup: 10004
    runAsGroup: 10004
    runAsNonRoot: true
    runAsUser: 10004
    seccompProfile:
      type: RuntimeDefault
    supplementalGroups: []
  resources:
    requests:
      memory: 50Mi
      cpu: 50m
    limits:
      memory: 200Mi
  serviceAccount:
    create: true
    name: ""
    annotations: {}
  tolerations: []
  topologySpreadConstraints: []
 PodAffinity[]
 NodeAffinity[]
 //     nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
              - matchExpressions:
                  - key: role
                    operator: In
                    values:
                      - Private 
                    ```

### PR DESCRIPTION
**Pull request by somaz94**
### What this does

Adds support for configuring `affinity` (nodeAffinity, podAffinity and podAntiAffinity) on component workloads, plus a global `affinity` applied to all components. A per-component `affinity` overrides the global one.

This follows the exact pattern of the already-merged `priorityClassName` support (#1438): a single `affinity` object passthrough, wired through every component's source fragment (`.json` + `.yaml.j2`), the global block in `sub_schema_values.yaml.j2`, and the `_pods.tpl` render (`{{- with (coalesce .affinity $root.Values.affinity) }}`).

### Why

Operators need to control pod scheduling (node/pod affinity and anti-affinity) for HA spreading, dedicated node pools, and co-location/anti-co-location — currently only `nodeSelector`, `tolerations`, `topologySpreadConstraints` and `priorityClassName` are exposed.

related: #368

### Notes

- `source/common/affinity.json` uses `additionalProperties: false` at the top level (only the three affinity sub-keys allowed) but `additionalProperties: true` inside each sub-object, so the full Kubernetes affinity content passes through while typos in the three top-level keys are still caught.
- Synapse worker pods inherit the affinity of the Synapse main process; per-worker affinity is not yet configurable (documented in the values comment).
- Source fragments were edited and `values.yaml` / `values.schema.json` regenerated via `scripts/assemble_helm_charts_from_fragments.sh` (idempotent — regeneration produces no drift).

### Testing

- New `tests/manifests/test_pod_affinity.py` (default-absent / per-component / global / component-overrides-global).
- Full `tests/manifests` suite green locally (2485 passed, 26 skipped).
- `reuse lint`, `ruff check`, `towncrier check` all pass.

### CODE REVIEW

File: charts/matrix-stack/source/common/affinity.json [hunk 1]
```diff
@@ -0,0 +1,18 @@
+{
+  "type": "object",
+  "properties": {
+    "nodeAffinity": {
+      "type": "object",
+      "additionalProperties": true
```
benbz: I'm fairly strongly against `additionalProperties: true` for these 3 objects. Unlike e.g. `extraVolumes` the potential options are fairly constrained. Having a strict schema that validates the structure of these is desirable IMO

charts/matrix-stack/source/common/sub_schema_values.yaml.j2
```diff
@@ -230,6 +231,19 @@ networking:
...
+{%- if global %}
+## Default affinity rules applied to the pods of all components.
+## An affinity set on an individual component overrides this global value for that component.
+## Synapse worker pods inherit the affinity of the Synapse main process; per-worker affinity is not yet configurable.
+{%- else %}
+## If specified, the pod's scheduling constraints.
+{%- endif %}
+## Supports nodeAffinity, podAffinity and podAntiAffinity.
+## More info: https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity
+{{ key }}: {}
```
benbz, line 281:
```suggestion
{{ key }}:
  nodeAffinity: {}
  podAffinity: {}
  podAntiAffinity: {}
```

With commented examples above each sub-key.

Potentially the default of `{}` needs to differ between globally and per-component so that you've a way of unsetting/blanking out a global e.g. `nodeAffinity`

charts/matrix-stack/templates/ess-library/_pods.tpl
```diff
@@ -31,6 +31,10 @@ nodeSelector:
 {{- with (coalesce .priorityClassName $root.Values.priorityClassName) }}
 priorityClassName: {{ . }}
 {{- end }}
+{{- with (coalesce .affinity $root.Values.affinity) }}
```
benbz: I think the 3 affinity sub-keys need to be rendered individually. At present setting `$.Values.affinity.nodeAffinity....` and `$.Values.synapse.affinity.podAntiAffinity...` means that Synapse loses the `nodeAffinity`. I don't think that's correct.

There should be a way to apply a global `nodeAffinity` and blank it it out, be it with `null` or empty dict and the tests should cover it.

Ideally `podAntiAffinity` should have `labelSelector` and `matchLabelKeys` set by default like how we do with `topologySpreadConstraints` but that's a nice to have and can be added later

tests/manifests/test_pod_affinity.py
benbz, line 1: These tests:
* Need to be done independently for each of the 3 affinities
* Should show that setting one of the affinities at a component level doesn't impact others at the global level
* Show that a global setting of one of the affinities can be blanked out on a per-component basis

somaz94:
Thanks @benbz — reworked per your review, pushed:

- **Strict schema**: dropped `additionalProperties: true`; `nodeAffinity`/`podAffinity`/`podAntiAffinity` are now validated structurally (only those 3 keys, and the `requiredDuringSchedulingIgnoredDuringExecution`/`preferredDuringSchedulingIgnoredDuringExecution` split per type). I kept the deeply-nested scheduling-term objects permissive since those are the full K8s affinity API and the API server validates them — happy to tighten further if you'd prefer.
- **Per-type rendering**: each affinity type is now resolved independently in `_pods.tpl`, so setting e.g. a component `podAntiAffinity` no longer drops a global `nodeAffinity`.
- **Blanking**: a component can blank an inherited global type by setting it to `{}` (or null). I kept the per-component default as `{}` (rather than materialising the 3 sub-keys) so "inherit" and "blank" stay distinguishable.
- **Tests**: each of the 3 affinity types is covered independently, plus override-isolation (the other types keep their global value) and per-component blank-out.

tests/manifests/test_pod_affinity.py
```diff
@@ -0,0 +1,196 @@
...
+        deployable_details.set_helm_values(values, PropertyType.Affinity, {affinity_type: {}})
+
+    iterate_deployables_workload_parts(blank_affinity)
+    for template in await make_templates(values):
+        if template["kind"] in ["Deployment", "StatefulSet", "Job"]:
+            pod_spec = template["spec"]["template"]["spec"]
+            deployable_details = template_to_deployable_details(template)
+            component_affinity = deployable_details.get_helm_values(values, PropertyType.Affinity)
+            affinity = pod_spec.get("affinity", {})
+            if component_affinity and component_affinity.get(affinity_type) == {}:
```
benbz: Why do we need this guard here?

charts/matrix-stack/source/common/sub_schema_values.yaml.j2
benbz: Commented examples are still outstanding

benbz:
>     * **Strict schema**: dropped `additionalProperties: true`; `nodeAffinity`/`podAffinity`/`podAntiAffinity` are now validated structurally (only those 3 keys, and the `requiredDuringSchedulingIgnoredDuringExecution`/`preferredDuringSchedulingIgnoredDuringExecution` split per type). I kept the deeply-nested scheduling-term objects permissive since those are the full K8s affinity API and the API server validates them — happy to tighten further if you'd prefer.

Thanks for addressing the comments. I would indeed like whole deeply nested structure encoded in the schema here. It will only be validated by the API server if Server-Side Apply is in use. Otherwise it will be silently ignored. And unlike with `extraVolumes` anything typo'ed here would be unlikely to cause `Pods` to be invalid

charts/matrix-stack/source/common/affinity.json
somaz94: Done — dropped `additionalProperties: true` entirely. Each of `nodeAffinity`/`podAffinity`/`podAntiAffinity` now uses `additionalProperties: false` and only permits its real sub-keys, and `nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution` is now strict with just `nodeSelectorTerms`. I deliberately left the individual term arrays untyped rather than fully modelling `matchExpressions`/`podAffinityTerm`: the construct step injects `additionalProperties: false` onto any object we describe, so a partial term schema rejects valid Kubernetes affinity. This validates the structure (allowed keys per level) without over-constraining term contents — happy to model the terms fully if you'd prefer.

charts/matrix-stack/source/common/sub_schema_values.yaml.j2
somaz94: Added commented per-type examples for all three sub-keys in both the global and per-component `affinity` blocks, each noting what it configures and (per-component) that `{}` blanks out the inherited global value. I kept them as a commented example block above `affinity: {}` rather than rendered children because values.yaml is round-tripped through ruamel during assembly and rendered-but-commented children weren't stable across a second `assemble_helm_charts_from_fragments.sh` run. On "default differs global vs per-component": inherit-vs-blank is keyed off presence, not the default value — a component inherits a global type while the key is absent and blanks it only when explicitly set to `{}` (`hasKey` in `_pods.tpl`), so both can default to `{}` safely. Can render the sub-keys out despite the round-trip caveat if you'd rather.

charts/matrix-stack/templates/ess-library/_pods.tpl
somaz94: This is covered now — each affinity type is resolved independently in `_pods.tpl`, so a component setting one type overrides only that type and still inherits the others from the global affinity; setting a type to `{}`/`null` blanks the inherited global for that type. Covered by `test_component_override_is_isolated_to_its_type` and `test_component_can_blank_global_affinity_type`. The default `labelSelector`/`matchLabelKeys` on `podAntiAffinity` I've left as the noted nice-to-have for a follow-up.

tests/manifests/test_pod_affinity.py
somaz94: Good catch — removed the guard; the blank-out assertion is now unconditional (`assert affinity_type not in affinity`), a stronger check, and the full suite passes with it unconditional so the guard was masking nothing. (The similar guard in `test_component_override_is_isolated_to_its_type` stays — there it legitimately skips the "other type inherits global" check when the component also overrode that other type.)

charts/matrix-stack/source/common/sub_schema_values.yaml.j2
benbz: e.g. https://github.com/somaz94/ess-helm/blob/7166df7327acc583c3ee45dd0abedd42db29e22d/charts/matrix-stack/source/common/sub_schema_values.yaml.j2#L120-L125 seems to work as a commented out complex setting with the blank, uncommented default below. It then round trips correctly at e.g. https://github.com/somaz94/ess-helm/blob/7166df7327acc583c3ee45dd0abedd42db29e22d/charts/matrix-stack/values.yaml#L620-L626

Potentially I'm not understanding the problem you're seeing

charts/matrix-stack/source/common/sub_schema_values.yaml.j2
```diff
@@ -230,6 +231,30 @@ networking:
...
+## Default affinity rules applied to the pods of all components.
+## Configured independently per affinity type (nodeAffinity, podAffinity, podAntiAffinity).
+## A component that sets one of these affinity types overrides only that type for that component; the other types are still inherited from here.
+## Synapse worker pods inherit the affinity of the Synapse main process; per-worker affinity is not yet configurable.
+## Supports the three Kubernetes affinity types, each configured as a sub-key, e.g.:
+##   nodeAffinity: {}       ## node-scheduling rules, e.g. requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms
+##   podAffinity: {}        ## co-scheduling rules, e.g. preferredDuringSchedulingIgnoredDuringExecution[].podAffinityTerm
+##   podAntiAffinity: {}    ## spreading rules, e.g. preferredDuringSchedulingIgnoredDuringExecution[].podAffinityTerm
+## More info: https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity
+{{ key }}: {}
```
benbz:
```suggestion
{{ key }}:
  nodeAffinity: {}
  podAffinity: {}
  podAntiAffinity: {}
```

And then the commented examples for each can be slotted above each one

charts/matrix-stack/source/common/affinity.json
```diff
@@ -0,0 +1,48 @@
...
+    "nodeAffinity": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "requiredDuringSchedulingIgnoredDuringExecution": {
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
+            "nodeSelectorTerms": {
+              "type": "array"
```
benbz: I really want to the see the full schema all the way down into these arrays. e.g. https://github.com/somaz94/ess-helm/blob/7166df7327acc583c3ee45dd0abedd42db29e22d/charts/matrix-stack/source/common/extraVolumeMountsWithContext.json, https://github.com/somaz94/ess-helm/blob/7166df7327acc583c3ee45dd0abedd42db29e22d/charts/matrix-stack/source/common/topologySpreadConstraints.json or https://github.com/somaz94/ess-helm/blob/7166df7327acc583c3ee45dd0abedd42db29e22d/charts/matrix-stack/source/common/image.json#L61-L70

charts/matrix-stack/source/common/affinity.json
```diff
@@ -0,0 +1,48 @@
+{
+  "type": "object",
+  "properties": {
+    "nodeAffinity": {
+      "type": "object",
+      "additionalProperties": false,
```
benbz: You don't need to specify any `additionalProperties: false` - the schema construction script injects this for any properties that don't have `additionalProperties: true` in them

benbz (commented):
Thanks for the updates. I'm happy with everything bar the schema and common values fragments now

somaz94:
Thanks @benbz — pushed `2fc297ab` addressing both the schema and the values fragments.

**Full affinity schema:** modelled the whole affinity API down into the arrays, mirroring the reference schemas you pointed at. `nodeSelectorTerm.json` covers `matchExpressions`/`matchFields` (node operator enum incl. `Gt`/`Lt`); `podAffinityTerm.json` covers `topologyKey` (required), `labelSelector`/`namespaceSelector` (`$ref file://common/labelSelector.json`), `namespaces`, `matchLabelKeys`, `mismatchLabelKeys`; the `preferred…` terms carry `weight` (1–100) + `preference`/`podAffinityTerm`. Dropped every explicit `additionalProperties: false` — the construction script injects it (confirmed in the assembled `values.schema.json`).

**Values fragment:** the global `affinity` now renders the three types as `{}` sub-keys with a commented example above each, per your suggestion. I deliberately kept the per-component default as `affinity: {}` (no rendered sub-keys): `_pods.tpl` resolves each type via `hasKey $componentAffinity <type>` — an absent key inherits the global, a present key (even `{}`) blanks/overrides it. Rendering `nodeAffinity: {}` etc. per-component would make every component blank all three global types by default and break inheritance — that is the global-vs-per-component default difference you flagged. The per-component doc block keeps the sub-keys as commented examples so users still see how to override/blank each type.

On the round-trip: assembling twice is now stable (`values.yaml` run2 == run1), so the nested rendered children survive re-assembly with this structure.

somaz94:
Thanks @benbz — you were right, and I've now modelled the affinity schema all the way down (superseding my earlier "left the term arrays untyped" note). In `2fc297ab`:

- **Full schema, matching the `topologySpreadConstraints.json` / `image.json` style**:
  - `nodeSelectorTerm.json` — `matchExpressions` / `matchFields` with `key` / `operator` (enum `In,NotIn,Exists,DoesNotExist,Gt,Lt`) / `values`.
  - `podAffinityTerm.json` — `labelSelector`, `namespaceSelector`, `namespaces`, `topologyKey` (required), `matchLabelKeys`, `mismatchLabelKeys`.
  - `labelSelector.json` — `matchExpressions` + `matchLabels` (string map).
  - `affinity.json` wires the three affinity types to these via `$ref`, with `weight` (1–100) on the preferred terms.
- **No `additionalProperties: false` anywhere** — as you noted, the schema-construction script injects it, so I only kept the legitimate `matchLabels` string-map `additionalProperties` in `labelSelector.json`.
- **Per-type global defaults** — the global `affinity` now renders `nodeAffinity: {}` / `podAffinity: {}` / `podAntiAffinity: {}` with a commented example above each (per your suggestion); the per-component default stays `affinity: {}`.

Regenerated `values.yaml` / `values.schema.json` (zero drift) and the manifests suite is green (234 affinity tests). PTAL 🙏

benbz (approved):
LGTM. Thank you very much for implementing this

PR #1452, opened 2026-07-16
Allow customizing the well-known delegation response headers

### LINKED ISSUE #1450: well-known delegation: no way to customise the served responses (e.g. add a Cache-Control header)
The haproxy backend serving /.well-known/matrix/{server,client,support} ([partial-haproxy.cfg.tpl](https://github.com/element-hq/ess-helm/blob/26.7.0/charts/matrix-stack/configs/well-known/partial-haproxy.cfg.tpl#L47-L62)) serves the JSON with a fixed set of response headers. Beyond the JSON content (additional), nothing about these responses can be influenced from values (checked on 26.7.0).
Short of ingress-controller-specific annotation hacks, the only route is patching the template.

My concrete case: Cache-Control. The responses are static JSON, yet are served with no Cache-Control and no validators (ETag/Last-Modified):

Proposal
A generic way to set extra response headers on the well-known responses, defaulting to none so existing installs are unaffected:

```
wellKnownDelegation:
  ## Additional response headers to set on /.well-known/matrix/* responses
  ## e.g.
  ## extraHeaders:
  ##   Cache-Control: "public, max-age=3600"
  extraHeaders: {}
```

rendered in partial-haproxy.cfg.tpl as:
```
{{- range $name, $value := .extraHeaders }}
  http-after-response set-header {{ $name }} "{{ $value }}"
{{- end }}
```

An even more general escape hatch (raw haproxy lines appended to the well-known-static backend) would cover further cases, but headers feel like the right level of abstraction for what this backend does - happy to go either way and send a PR for whichever direction you'd prefer.

ad35cm:
Closes [#1450](https://github.com/element-hq/ess-helm/issues/1450), implementing what was proposed in https://github.com/element-hq/ess-helm/issues/1450#issuecomment-4955761438.

Adds wellKnownDelegation.headers, applied to all /.well-known/matrix/* responses.
Defaults to the previously hardcoded headers plus Cache-Control: "public, max-age=3600".
Entries can be modified, added, or removed (key: null).

Default render matches the previous output (plus Cache-Control); modify/add/remove verified

### DIFF
--- a/charts/matrix-stack/configs/well-known/partial-haproxy.cfg.tpl
+++ b/charts/matrix-stack/configs/well-known/partial-haproxy.cfg.tpl
@@ -1,6 +1,6 @@
 {{- /*
 Copyright 2024-2025 New Vector Ltd
-Copyright 2025 Element Creations Ltd
+Copyright 2025-2026 Element Creations Ltd
 
 SPDX-License-Identifier: AGPL-3.0-only
 */ -}}
--- a/charts/matrix-stack/configs/well-known/partial-haproxy.cfg.tpl
+++ b/charts/matrix-stack/configs/well-known/partial-haproxy.cfg.tpl
@@ -47,15 +47,9 @@ frontend well-known-in
 backend well-known-static
   mode http
 
-  http-after-response set-header X-Frame-Options SAMEORIGIN
-  http-after-response set-header X-Content-Type-Options nosniff
-  http-after-response set-header X-XSS-Protection "1; mode=block"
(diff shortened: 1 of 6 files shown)


benbz (commented):
Thanks. This all LGTM and as discussed. The copyright dates just need adjusting on a couple of files as per https://github.com/element-hq/ess-helm/actions/runs/29465183988/job/87568472788?pr=1452 and then we're good to merge

benbz (approved):

benbz:
Thank you very much for implementing this!

PR #1456, opened 2026-07-16
fix MAS pointing to Synapse main

gaelgatelement:
Nowadays the `mas-helper` worker exists, and we should profit from it.

### DIFF
--- a/charts/matrix-stack/configs/matrix-authentication-service/config-overrides.yaml.tpl
+++ b/charts/matrix-stack/configs/matrix-authentication-service/config-overrides.yaml.tpl
@@ -82,7 +82,7 @@ matrix:
                         "defaultSecretKey" "SYNAPSE_SHARED_SECRET"
                       )
                   ) }}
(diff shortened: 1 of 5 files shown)


charts/matrix-stack/configs/matrix-authentication-service/config-overrides.yaml.tpl
```diff
@@ -82,7 +82,7 @@ matrix:
                         "defaultSecretKey" "SYNAPSE_SHARED_SECRET"
                       )
                   ) }}
-  endpoint: "http://{{ include "element-io.synapse.internal-hostport" (dict "root" $root "context" (dict "targetProcessType" "main")) }}"
+  endpoint: "http://{{ include "element-io.synapse.internal-hostport" (dict "root" $root "context" (dict "targetProcessType" "")) }}"
```
benbz: This appears to be the only usage of `targetProcessType`, so remove that functionality?

benbz (approved):

PR #1461, opened 2026-07-20
Add schedulerName and runtimeClassName support to component workloads

somaz94:
Adds `schedulerName` and `runtimeClassName` as pod-scheduling knobs on every
component workload, plus matching global values applied to all components
(a per-component value overrides the global one).

This mirrors the existing `priorityClassName` (#1438) and `affinity` (#1442)
support, filling out the remaining common pod-spec scheduling fields. Both are
plain string scalars rendered via the same coalesce(component, global) pattern
in _pods.tpl.

Validation (local):
- Regenerated values.yaml + values.schema.json from source fragments
  (scripts/assemble_helm_charts_from_fragments.sh); git diff clean.
- New manifest tests test_pod_schedulerName.py + test_pod_runtimeClassName.py
  pass (full pytest run: 144 passed).
- reuse lint: 656/656 files compliant.

A newsfragment will be added once the PR number is assigned.

### DIFF
--- a/charts/matrix-stack/source/common/sub_schema_values.yaml.j2
+++ b/charts/matrix-stack/source/common/sub_schema_values.yaml.j2
@@ -46,6 +46,8 @@ image:
 {{ tolerations(global=true) }}
 {{ topologySpreadConstraints(global=true) }}
 {{ priorityClassName(global=true) }}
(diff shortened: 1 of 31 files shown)


benbz (approved):
Thanks for implementing these

PR #1462, opened 2026-07-20
Unify Pod Spec handling in manifest tests

benbz:
Unify handling of workload manifests to ensure that we don't accidentally miss a `kind` in some places.

Inside `PodTemplateDetails` we use `manifest` and `manifest_id` rather than `template` and `template_id` because we've been incorrectly naming things until now. `template` isn't a Helm template, it is a rendered Helm template which is a fully fledged manifest. `template_id`, `template_to_deployable_details`, `make_templates`, etc will be renamed separately but I've not introduced any further tech debt.

Draft as WIP but want to see what I've broken so far

### DIFF
--- a/tests/manifests/test_labels.py
+++ b/tests/manifests/test_labels.py
@@ -1,5 +1,5 @@
 # Copyright 2025 New Vector Ltd
-# Copyright 2025 Element Creations Ltd
+# Copyright 2025-2026 Element Creations Ltd
 #
 # SPDX-License-Identifier: AGPL-3.0-only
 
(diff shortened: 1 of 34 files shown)


tests/manifests/test_labels.py
```diff
@@ -9,7 +9,7 @@
 import pytest
 
 from . import PropertyType, secret_values_files_to_test, values_files_to_test
-from .utils import template_id, template_to_deployable_details
+from .utils import PERISTENT_WORKLOAD_KINDS, iterate_pod_template, template_id
```
gaelgatelement:
```suggestion
from .utils import PERSISTENT_WORKLOAD_KINDS, iterate_pod_template, template_id
```

gaelgatelement (approved):

tests/manifests/test_labels.py
benbz: Whops. Great catch. Thanks

PR #1465, opened 2026-07-22
Allow RTC services to be of type ClusterIP

### LINKED ISSUE #1464: Allow possibility to use `ClusterIP` and one Service for rtc-sfu component
I'm currently running the latest ess-helm chart on a Kubernetes in the cloud. The nodes do not have public IPs and it is thus expected to expose the UDP and TCP port through a LoadBalancer. 

TURN-TLS is also expected to be done through Traefik or any other ingress controller that supports TCP routes.

for the UDP and TCP port the current setup force me to deploy 2 LB (one for rtcTCP and one for rtcUDPMux) which costs more and creates 2 different IPs which won't work as I need a single manualIP for the sfu.

Regarding TURN-TLS currently I'm forced to create a service with a NodePort that is not used at all since traffic goes  Browser -> Traefik -> Service -> Pod. A ClusterIP type would be enough

The solution I'd like for TURN-TLS would be to allow for the service type to be `ClusterIP` as well when needed.

For the double LB the solution would be to be able to have only one service opening both the TCP and UDP ports on the same LoadBalancer allowing me to have one external floating IP.

For TURN-TLS I currently juste let it take a NodePort that is then useless.

My current workaround for the LoadBalancer is to basically disable all the exposed services and do it manually by adding:

```
 matrixRTC:
    sfu:
      additional:
        0-rtc:
          config: |
            rtc:
              tcp_port: 30001
              udp_port: 30002
      exposedServices:
        rtcTcp:
          enabled: false
        rtcMuxedUdp:
          enabled: false
```

and a manifest for the service:
```
---
apiVersion: v1
kind: Service
metadata:
  name: {{ $.Release.Name }}-matrix-rtc-sfu-external
spec:
  type: LoadBalancer
  ipFamilyPolicy: PreferDualStack
  loadBalancerIP: "{{ $.Values.matrixStack.matrixRTC.sfu.manualIP }}"
  ports:
  - name: "rtc-tcp"
    protocol: "TCP"
    port: 30001
    targetPort: 30001
    nodePort: 30001
  - name: "rtc-muxed-udp"
    protocol: "UDP"
    port: 30002
    targetPort: 30002
    nodePort: 30002
  selector:
    app.kubernetes.io/instance: "{{ $.Release.Name }}-matrix-rtc-sfu"
```

**Additional context**

I could work on a PR if that's something you are interested in

jocelynthode:
Let a user choose `ClusterIP` as the service type when they don't need to be exposed directly.

Fixes #1464

### DIFF
--- a/newsfragments/1465.added.md
+++ b/newsfragments/1465.added.md
@@ -1,1 +1,1 @@
-Add support for setting `portType` in `matrixRTC.sfu.exposedServices` to `ClusterIP`
+Add support for setting `portType` in `matrixRTC.sfu.exposedServices` to `ClusterIP`.
--- a/newsfragments/1465.added.md
+++ b/newsfragments/1465.added.md
@@ -1,0 +1,1 @@
+Add support for setting `portType` in `matrixRTC.sfu.exposedServices` to `ClusterIP`
--- a/charts/matrix-stack/source/common/exposedServicePort.json
+++ b/charts/matrix-stack/source/common/exposedServicePort.json
@@ -18,6 +18,7 @@
     "portType": {
       "type": "string",
       "enum": [
+        "ClusterIP",
         "NodePort",
         "HostPort",
         "LoadBalancer"
--- a/charts/matrix-stack/source/common/exposedServicePortRange.json
+++ b/charts/matrix-stack/source/common/exposedServicePortRange.json
@@ -18,6 +18,7 @@
     "portType": {
       "type": "string",
       "enum": [
+        "ClusterIP",
         "NodePort",
         "HostPort",
         "LoadBalancer"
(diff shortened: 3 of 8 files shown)


newsfragments/1465.added.md
```diff
@@ -0,0 +1 @@
+Add support for setting `portType` in `matrixRTC.sfu.exposedServices` to `ClusterIP`
```
benbz:
```suggestion
Add support for setting `portType` in `matrixRTC.sfu.exposedServices` to `ClusterIP`.
```

To appease CI

benbz (approved):

benbz:
Thanks for the contribution!

PR #1497, opened 2026-08-04
Add command to truncate e2e_one_time_keys_json

mgr43:
Added command to truncate e2e_one_time_keys_json table in PostgreSQL to the backup instructions in docs/maintenance.md.
This brings it in line with what's currently present at https://github.com/element-hq/synapse/blob/develop/docs/usage/administration/backups.md

### DIFF
--- a/docs/maintenance.md
+++ b/docs/maintenance.md
@@ -66,6 +66,7 @@ kubectl exec -n ess sts/ess-postgres -- psql -U postgres -c 'DROP ROLE synapse_u
 kubectl exec -n ess sts/ess-postgres -- psql -U postgres -c 'DROP ROLE matrixauthenticationservice_user'
 kubectl cp dump.sql ess-postgres-0:/tmp -n ess
 kubectl exec -n ess sts/ess-postgres -- bash -c "psql -U postgres -d postgres < /tmp/dump.sql"
+kubectl exec -n ess sts/ess-postgres -- psql -U postgres -d synapse -c 'TRUNCATE e2e_one_time_keys_json'
 ```
 Adjust to your own kubernetes namespace and release name if required.
 


benbz:
Thanks for this. Is there a reason in your mind to prefer truncating after restoration vs `--exclude-table-data e2e_one_time_keys_json` on the dump, which is the approach https://github.com/element-hq/synapse/blob/develop/docs/usage/administration/backups.md#quick-and-easy-database-backup-and-restore takes

mgr43:
> Thanks for this. Is there a reason in your mind to prefer truncating after restoration vs `--exclude-table-data e2e_one_time_keys_json` on the dump, which is the approach https://github.com/element-hq/synapse/blob/develop/docs/usage/administration/backups.md#quick-and-easy-database-backup-and-restore takes

The reason why truncating the table is done after is because we're using pg_dumpall
https://github.com/element-hq/ess-helm/blob/fcc89baae53ab99f2255cac056e85d3ec3364751/docs/maintenance.md?plain=1#L38
Plain Synapse has only one DB to take care of, so it uses pg_dump, which lets you specify to exclude certain tables and many other things. Which isn't possible with pg_dumpall.
The docs for the respective commands [pg_dump](https://www.postgresql.org/docs/current/app-pgdump.html) and [pg_dumpall](https://www.postgresql.org/docs/current/app-pg-dumpall.html) are useful to look over in this context.

benbz:
> > Thanks for this. Is there a reason in your mind to prefer truncating after restoration vs `--exclude-table-data e2e_one_time_keys_json` on the dump, which is the approach https://github.com/element-hq/synapse/blob/develop/docs/usage/administration/backups.md#quick-and-easy-database-backup-and-restore takes
> 
> The reason why truncating the table is done after is because we're using pg_dumpall
> 
> https://github.com/element-hq/ess-helm/blob/fcc89baae53ab99f2255cac056e85d3ec3364751/docs/maintenance.md?plain=1#L38
> 
> Plain Synapse has only one DB to take care of, so it uses pg_dump, which lets you specify to exclude certain tables and many other things. Which isn't possible with pg_dumpall.
> The docs for the respective commands [pg_dump](https://www.postgresql.org/docs/current/app-pgdump.html) and [pg_dumpall](https://www.postgresql.org/docs/current/app-pg-dumpall.html) are useful to look over in this context.

:+1: makes sense

benbz (approved):

PR #1519, opened 2026-08-19
Upgrade MatrixRTC Authoriser to v0.6.0

benbz:
Brings in https://github.com/element-hq/lk-jwt-service/releases/tag/v0.5.0 & https://github.com/element-hq/lk-jwt-service/releases/tag/v0.6.0.

This doesn't make use of Redis yet, I think that change is chunky enough to be its own PR.

~~The internal routing for MatrixRTC authoriser in CI/the test cluster is removed as `LIVEKIT_CS_API_URL_OVERRIDES` is set appropriately so no well-known lookups happen and the Synapse path is known. Unclear whether the authoriser still needs to make requests to itself.~~ `LIVEKIT_CS_API_URL_OVERRIDES` is only for the CS API. Well-Knowns and Synapse interaction still happens for the SS API.

The situation with the SFU webhook key is a bit interesting. I'm not sure this is going to work with `keysYaml` support and I might need to tighten restrictions there. Edit: we've removed `keysYaml`. The Authoriser crashes with >1 key and so there's no value in supporting `keysYaml`, only complexity.


gaelgatelement:
> The situation with the SFU webhook key is a bit interesting. I'm not sure this is going to work with keysYaml support and I might need to tighten restrictions there

I'm not sure to follow what might be going wrong here? Can you explain the issue you think could happen?

benbz:
> > The situation with the SFU webhook key is a bit interesting. I'm not sure this is going to work with keysYaml support and I might need to tighten restrictions there
> 
> I'm not sure to follow what might be going wrong here? Can you explain the issue you think could happen?

With `keysYaml` you don't necessarily have a single key in the YAML and any key known to both the authoriser and the SFU could be used. I suspect `LIVEKIT_KEY` is unset in this scenario too

gaelgatelement (approved):

PR #1529, opened 2026-08-25
matrix-tools: Add support for x25519 keys

gaelgatelement:
This adds support for x25519 private-keys.


benbz (approved):
From an SPT PoV this looks good (though needs `BUILD_MATRIX_TOOLS` testing).

I am explicitly not commenting on the crypto

benbz:
As discussed this is a good point to also bump to Golang 1.27

gaelgatelement:
> As discussed this is a good point to also bump to Golang 1.27

We have reconsidered this and will wait a couple of time for golang 1.27 to leave edge repositories of distributions before moving to 1.27.

benbz (approved):
Happy for the `matrix-tools` dance to proceed

PR #1534, opened 2026-08-26
haproxy: `element-io.synapse.ingress.additionalPaths` helper

gaelgatelement:
This makes the templates more consistent with how we refer to Synapse additional paths.


charts/matrix-stack/templates/synapse/synapse_ingress.yaml
```diff
@@ -26,6 +26,9 @@ spec:
   - host: {{ (tpl .ingress.host $) | quote }}
     http:
       paths:
+{/*- `only_externally` and `blocked` path are handled in the public Ingress,
```
benbz:
```suggestion
{/*- `only_externally` and `blocked` path are handled in the public Ingress only so that internal routing isn't impacted,
```

benbz (approved):

PR #1540, opened 2026-08-26
fix deployment markers annotations field empty.

nlamamra-acton:


gaelgatelement:
Thank you for your contribution @nlamamra-acton ! Can you create a newsfragments in the `newsfragments` directory please ? You can use the command `towncrier create` to do so.

benbz:
FYI we're going to ignore the `no-changes-after-building` failing check as we believe / hopeful it will be resolved with https://github.com/element-hq/ess-helm/pull/1541

benbz (approved):

benbz:
Thanks for this. I'm going to follow-up shortly with a PR that introduces a test to catch this going forward and test the Pod annotations too (there's at least 2 similar issues there)

benbz:
> Thanks for this. I'm going to follow-up shortly with a PR that introduces a test to catch this going forward and test the Pod annotations too (there's at least 2 similar issues there)

#1542 adds this test and corrects 2 instances within pod templates as well

PR #1545, opened 2026-08-27
Add support for emptyDir sizeLimit

gaelgatelement:
Initialized with GLM 5.2, and reworked manually to adjust to our tests architecture.

This adds support for configure `sizeLimit` on any `emptyDir` property of our workloads.


charts/matrix-stack/source/common/ephemeralStorage.json
```diff
@@ -0,0 +1,9 @@
+{
+  "type": "object",
+  "properties": {
+    "sizeLimit": {
```
benbz, line 4: Do we want to adapt the regex from resources into here?

tests/manifests/__init__.py
```diff
@@ -131,6 +132,7 @@ class DeployableDetails(abc.ABC):
     is_hook: bool = field(default=False, hash=False)
     has_mount_context: bool = field(default=None, hash=False)  # type: ignore[assignment]
     is_synapse_process: bool = field(default=False, hash=False)
+    has_ephemeral_storage: bool = field(default=False, hash=False)
```
benbz: Feels like most components have some sort of ephemeral storage, invert the default?

charts/matrix-stack/templates/synapse/_synapse_pod.tpl
```diff
@@ -210,10 +210,16 @@ We have an init container to render & merge the config for several reasons:
...
+        sizeLimit: {{ . }}
+        {{- end }}
       name: "media"
 {{- end }}
     - emptyDir:
         medium: Memory
+        {{- with $root.Values.synapse.ephemeralStorages.tmp.sizeLimit }}
+        sizeLimit: {{ . }}
+        {{- end }}
       name: "tmp"
```
benbz: What's this volume for?

charts/matrix-stack/source/common/ephemeralStorage.json
gaelgatelement: Good idea, I added the regex for quantity strings from https://github.com/kubernetes/apimachinery/blob/master/pkg/api/resource/quantity.go#L143

charts/matrix-stack/templates/synapse/_synapse_pod.tpl
gaelgatelement: This volume was added as it was needed for some modules to work properly, such as SAML authentication.

Now that we have support for `extraVolumes` and `extraVolumeMounts`, we should just get rid of it. Users are free to add this if they need it.

tests/manifests/__init__.py
```diff
@@ -410,11 +417,11 @@ def make_synapse_worker_sub_component(worker_name: str, worker_type: str) -> Sub
         has_ingress=False,
         is_synapse_process=True,
         is_singleton=(worker_type != "scalable"),
-        ignore_unreferenced_mounts={"synapse": ("/tmp",)},
```
benbz, line 413: Heh, I guess this was telling us something :p

tests/manifests/test_ephemeral_storages.py
```diff
@@ -0,0 +1,105 @@
...
+    for pod_template_details in iterate_pod_template(templates):
+        volumes = pod_template_details.pod_template["spec"].get("volumes", [])
+        assert (
+            any("emptyDir" in v for v in volumes) == pod_template_details.deployable_details().has_ephemeral_storage
+        ), f"{pod_template_details.manifest_id}: emptyDir volumes is not consistent with has_ephemeral_storage"
+        for volume in volumes:
+            if "emptyDir" not in volume:
+                continue
+            volume_name = volume["name"]
+            assert volume["emptyDir"].get("sizeLimit"), (
```
benbz, line 70: Can we move the `emptyDir.medium == memory` test from `test_volumes.py` in here?

tests/manifests/__init__.py
```diff
@@ -592,6 +605,7 @@ def make_synapse_worker_sub_component(worker_name: str, worker_type: str) -> Sub
         has_service_monitor=False,
         makes_outbound_requests=False,
         ignore_unreferenced_mounts={"element-admin": ("/tmp",)},
+        ephemeral_storages={"nginx-tmp": "nginxTemporaryFiles"},
```
benbz: Do we want to take the opportunity before we add values relating to this to change this to be "temp": "temp" like other places?

tests/manifests/__init__.py
```diff
@@ -621,6 +635,7 @@ def make_synapse_worker_sub_component(worker_name: str, worker_type: str) -> Sub
             )
         },
         content_volumes_mapping={"/tmp": ("element-web-config",)},
+        ephemeral_storages={"nginx-tmp": "nginxTemporaryFiles"},
```
benbz: Same Q about unifying on `temp`?

tests/manifests/__init__.py
```diff
@@ -683,6 +699,7 @@ def make_synapse_worker_sub_component(worker_name: str, worker_type: str) -> Sub
                 is_hook=True,
                 has_mount_context=False,
                 makes_outbound_requests=False,
+                ephemeral_storages={"tmp-mas-cli": "tmpMasCli"},
```
benbz, line 702: Same Q about unifying on `temp`?

gaelgatelement: While `nginx-tmp` is actually mounted on `/tmp`, `tmp-mas-cli` is mounted on `/tmp-mas-cli`. I think we should keep it like this.

Because we are mounting it on `/tmp`, should we use `tmp` everywhere instead of `temp` ?

benbz: I'd be inclined to keep `/tmp` as the mount point as that's a standard but use `temp` in the values file (given human readable). I have minimal feeling about `tmp` vs `temp` as the volume name.

On the specific MAS CLI volume, I'm happy to keep that as `/tmp-mas-cli` but `tempMasCLI` in the values file?

benbz (approved):

PR #1546, opened 2026-08-27
MatrixRTC: restrict authoriser to 1 replica

benbz:
https://github.com/element-hq/ess-helm/pull/1519 introduced delayed events support into MatrixRTC. This requires the authoriser service to keep state. Whilst the authoriser service does optionally support Redis (and ESS is planning on wiring this up soon), Redis is only consulted on startup to allow the authoriser service to restart and not totally drop its state.

Until state is looked up from Redis/another store at other points than startup, the authoriser must not allow > 1 replica.


gaelgatelement (approved):

benbz:
Marked as a breaking change with https://github.com/element-hq/ess-helm/commit/da253021c8efb867b440816778825e08eac4eeec

PR #1554, opened 2026-08-28
Add support for `emptyDir.medium` configurability

gaelgatelement:
Built with GLM 5.2.

All `ephemeralStorages` get a new `medium` key. It can either be an empty string or `Memory`. If it is an empty string, it will result in on-disk ephemeral storage on the node. All our `ephemeralStorages` default to `Memory` for now.

### DIFF
--- a/charts/matrix-stack/templates/element-admin/deployment.yaml
+++ b/charts/matrix-stack/templates/element-admin/deployment.yaml
@@ -71,14 +71,7 @@ spec:
         - {{ (. | toYaml) | nindent 10 }}
 {{- end }}
       volumes:
-      - emptyDir:
-          {{- with $.Values.elementAdmin.ephemeralStorages.tmp.medium }}
-          medium: {{ . }}
-          {{- end }}
-          {{- with $.Values.elementAdmin.ephemeralStorages.tmp.sizeLimit }}
-          sizeLimit: {{ . }}
-          {{- end }}
-        name: tmp
+{{- include "element-io.ess-library.empty-dir" (dict "root" $ "context" (dict "componentValues" $.Values.elementAdmin "ephemeralStorage" "tmp" "volumeName" "tmp")) | nindent 6 }}
 {{- range .extraVolumes }}
       - {{- (tpl (. | toYaml) $) | nindent 8 }}
 {{- end }}
--- a/charts/matrix-stack/templates/element-admin/deployment.yaml
+++ b/charts/matrix-stack/templates/element-admin/deployment.yaml
@@ -72,7 +72,9 @@ spec:
 {{- end }}
       volumes:
       - emptyDir:
-          medium: Memory
+          {{- with $.Values.elementAdmin.ephemeralStorages.tmp.medium }}
+          medium: {{ . }}
+          {{- end }}
           {{- with $.Values.elementAdmin.ephemeralStorages.tmp.sizeLimit }}
           sizeLimit: {{ . }}
           {{- end }}
--- a/charts/matrix-stack/templates/element-admin/deployment.yaml
+++ b/charts/matrix-stack/templates/element-admin/deployment.yaml
@@ -71,7 +71,8 @@ spec:
         - {{ (. | toYaml) | nindent 10 }}
 {{- end }}
       volumes:
-{{- include "element-io.ess-library.empty-dir" (dict "root" $ "context" (dict "componentValues" $.Values.elementAdmin "ephemeralStorage" "tmp" "volumeName" "tmp")) | nindent 6 }}
+      - emptyDir: {{- $.Values.elementAdmin.ephemeralStorages.tmp | toYaml | nindent 10 }}
+        name: tmp
 {{- range .extraVolumes }}
       - {{- (tpl (. | toYaml) $) | nindent 8 }}
 {{- end }}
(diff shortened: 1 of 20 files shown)


charts/matrix-stack/templates/element-admin/deployment.yaml
```diff
@@ -71,12 +71,7 @@ spec:
         - {{ (. | toYaml) | nindent 10 }}
 {{- end }}
       volumes:
-      - emptyDir:
-          medium: Memory
-          {{- with $.Values.elementAdmin.ephemeralStorages.tmp.sizeLimit }}
-          sizeLimit: {{ . }}
-          {{- end }}
-        name: tmp
+{{- include "element-io.ess-library.empty-dir" (dict "root" $ "context" (dict "componentValues" $.Values.elementAdmin "ephemeralStorage" "tmp" "volumeName" "tmp")) | nindent 6 }}
```
benbz: Why don't we just do

```yaml
      - emptyDir: {{- $.Values.elementAdmin.ephemeralStorages.tmp | toYaml | nindent 8 }}
        name: tmp
```

tests/manifests/test_ephemeral_storages.py
```diff
@@ -70,12 +70,6 @@ async def test_ephemeral_storage_defaults(templates, values, base_values):
             assert volume["emptyDir"].get("sizeLimit"), (
                 f"{pod_template_details.manifest_id}: emptyDir volume '{volume_name}' is lacking sizeLimit"
             )
-            assert "medium" in volume["emptyDir"], (
```
benbz, line 73: We still want both of these in this test I'd have thought

tests/manifests/test_ephemeral_storages.py
```diff
@@ -109,3 +103,73 @@ def set_ephemeral_storage_empty_dir_size_limit(deployable_details: DeployableDet
...
+    """
+    # Configure the shared rendered-config volume directly, preserving its sizeLimit.
+    rendered_config_default_medium = base_values["matrixTools"]["ephemeralStorages"]["renderedConfig"]["medium"]
+    rendered_config_non_default = _non_default_medium(rendered_config_default_medium)
+    values.setdefault("matrixTools", {}).setdefault("ephemeralStorages", {}).setdefault("renderedConfig", {})[
+        "medium"
+    ] = rendered_config_non_default
+
+    # Configure each component-specific ephemeral storage's medium via set_helm_values.
+    def set_ephemeral_storage_medium(deployable_details: DeployableDetails):
```
benbz:
```suggestion
    # Configure each component-specific ephemeral storage'
    def set_ephemeral_storage(deployable_details: DeployableDetails):
```

tests/manifests/test_ephemeral_storages.py
```diff
@@ -109,3 +103,73 @@ def set_ephemeral_storage_empty_dir_size_limit(deployable_details: DeployableDet
...
+        ephemeral_storages = deployable_details.get_helm_values(base_values, PropertyType.EphemeralStorages)
+        if not ephemeral_storages:
+            return
```
benbz: Why's this needed given `iterate_deployables_parts` uses `lambda deployable_details: deployable_details.has_ephemeral_storage`

charts/matrix-stack/templates/element-admin/deployment.yaml
gaelgatelement: The problem is that we cannot set `medium: ""` in `emptyDir`, we have to unset the medium key, which is rather hard with how helm values are merged ?

If we set : 

```
ephemeralStorages:
  tmp:
    sizeLimit: 10Mi
```

It will still be rendered with `medium: Memory`.

tests/manifests/test_ephemeral_storages.py
gaelgatelement: I think its not necessary now that we explicitly define it in the values?

benbz: We still want to assert that those are our defaults

charts/matrix-stack/templates/element-admin/deployment.yaml
benbz: > If we set :
> 
> ```
> ephemeralStorages:
>   tmp:
>     sizeLimit: 10Mi
> ```
> 
> It will still be rendered with `medium: Memory`.

That's desirable as it is the default in the chart's values file.

If we want to set `medium: ""` (or omit which given `emptyDir`'s default is the same thing then we need to do:

```yaml
ephemeralStorages:
  tmp:
    medium: ""
    sizeLimit: 10Mi
```

This is currently true with the helper given

```
{{- with $storage.medium }}
    medium: {{ . }}
{{- end }}
```

gaelgatelement: Yes but if I do your strategy: `{{- $.Values.elementAdmin.ephemeralStorages.tmp | toYaml | nindent 8 }}`

`toYaml` will render `medium: Memory` when we do not set `medium` in the values file ? And if we set `medium: ""` ,it will render it as-is, which is also incorrect?

benbz: The `medium: ""` from user values will override the `medium: Memory` from the chart values?

gaelgatelement: Oh I was certain that `medium: ""` would be an invalid value, but actually it seems valid : 
https://github.com/kubernetes-client/c/blob/master/kubernetes/docs/v1_empty_dir_volume_source.md

I'll adjust to your suggestion then.

tests/manifests/test_ephemeral_storages.py
gaelgatelement: An assertion should make it clearer.

tests/manifests/test_ephemeral_storages.py
gaelgatelement: I renamed to a more specific function name for clarity.

benbz (approved):

PR #1555, opened 2026-08-28
MatrixRTC: configure Redis

benbz:
#1519 supported delayed events. It also adds Redis support to provide restart-safety for delayed events when the authoriser pod restarts. As per #1546 Redis is only used on startup and so multiple replicas aren't possible.

As `LIVEKIT_REDIS_URL` is a single string and so we can't support loading the password from a file. For this reason we're temporarily not allowing an external Redis with the authoriser as we'd have to restrict the auth options. The plan is to update the authoriser to support separate env vars for auth and so an external Redis can be fully supported.

Given the number of distinct usages of Redis make sure each component that uses it uses its own distinct db.

And given that Redis is being spun up for the authoriser we might as well configure it on the SFU too

### DIFF
--- a/charts/matrix-stack/configs/matrix-rtc/sfu/config-overrides.yaml.tpl
+++ b/charts/matrix-stack/configs/matrix-rtc/sfu/config-overrides.yaml.tpl
@@ -74,6 +74,20 @@ turn:
 {{- end }}
 {{- end }}
 
+redis:
+{{- if .redis }}
+  address: {{ tpl .redis.host $root }}:{{ .redis.port | default 6379 }}
+{{- with .redis.db }}
+  db: {{ . }}
+{{- end }}
+{{- if .redis.password }}
+  password: ${SFU_REDIS_PASSWORD}
+{{- end }}
+{{- else }}
+  address: "{{ $root.Release.Name }}-redis.{{ $root.Release.Namespace }}.svc.{{ $root.Values.clusterDomain }}:6379"
+  db: 3
+{{- end }}
+
 room:
   auto_create: false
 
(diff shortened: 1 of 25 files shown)


charts/matrix-stack/configs/matrix-rtc/sfu/config-overrides.yaml.tpl
```diff
@@ -74,6 +74,20 @@ turn:
...
+  address: {{ tpl .redis.host $root }}:{{ .redis.port | default 6379 }}
+{{- with .redis.db }}
+  db: {{ . }}
+{{- end }}
+{{- if .redis.password }}
+  password: ${SFU_REDIS_PASSWORD}
+{{- end }}
+{{- else }}
+  address: "{{ $root.Release.Name }}-redis.{{ $root.Release.Namespace }}.svc.{{ $root.Values.clusterDomain }}:6379"
+  db: 3
```
gaelgatelement, line 88: Can we double-check with backend that it is fine to change this with a progressive rollout, considering the issues we saw with contamination before?

benbz: This is the SFU. Synapse is very deliberately still using 0 albeit made explicit

gaelgatelement: Mmmh I misplace my comment. But OK, that's fine then!

gaelgatelement (approved):

charts/matrix-stack/configs/matrix-rtc/sfu/config-overrides.yaml.tpl
benbz: Triple checking with Backend though

benbz: Confirmed 0 is what Synapse uses by default and so this is just made explicit

PR #1558, opened 2026-09-02
Replace Redis container image with latest Valkey image

benbz:
Simply replace the built-in Redis with Valkey. No schema changes, no values changes, no naming changes at this point in time. Those will come in later PRs, each of which should be self-contained and provide value in and of themselves

### DIFF
--- a/.github/element-docs-words.txt
+++ b/.github/element-docs-words.txt
@@ -51,5 +51,6 @@ templatable
 threepids
 towncrier
 Traefik
+Valkey
 Wifi
 zizmor
--- a/charts/matrix-stack/source/redis.yaml.j2
+++ b/charts/matrix-stack/source/redis.yaml.j2
@@ -12,7 +12,7 @@ maxMemoryPolicy: allkeys-lru
 # Use 0 replicas to stop Redis
 replicas: 1
 
-{{- sub_schema_values.image(registry='docker.io', repository='library/redis', tag='7.4-alpine') }}
+{{- sub_schema_values.image(registry='docker.io', repository='valkey/valkey', tag='9.1-alpine') }}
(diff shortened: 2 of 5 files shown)


gaelgatelement (approved):

.github/element-docs-words.txt
```diff
@@ -51,5 +51,6 @@ templatable
 threepids
```
MadLittleMods, line 51: Any background for why we migrated to Valkey?

benbz: We'd got ourselves a bit stuck on an old Redis version due to its license changes. So moving to Valkey was wanted from that angle and we could then start updating again rather then relying on still getting security releases for the old version.

I'd previously raised https://github.com/element-hq/ess-helm/pull/1368 that changed Redis from a `Deployment` to `StatefulSet` but this was blocked due to not being able to change Redis' `Service` to a headless one (`spec.clusterIP: None`) as this is immutable once the `Service` is in cluster. Helm provides no mechanism for recreating individual resources (just `helm uninstall` then `helm install` on the whole stack). 

The priority of https://github.com/element-hq/ess-helm/pull/1368 jumped up the other week when we realised there were situations were we could be running > 1 Redis `Pod`. https://github.com/element-hq/ess-helm/pull/1557 was done quickly to workaround the most likely cause but 1368 was the better / more complete fix.

This series then occurred so that 1368 could be updated to do `Deployment` -> `StatefulSet`, `ClusterIP` `Service` -> headless `Service` whilst renaming and so from Helm's PoV there was the deletion of all the resources named `{{ Release.Name }}-redis` and coincidental creation of resources named `{{ Release.Name }}-valkey`.

PR #1562, opened 2026-09-02
Introduce `redisOrValkey` for components that support external Redis/Valkey

benbz:
This PR introduces a replacement for `{hookshot,synapse,matrixRTC.sfu}.redis` in the form of `{hookshot,synapse,matrixRTC.sfu}.redisOrValkey`. Only 1 will be used, and `redisOrValkey` will be preferred.

The CI values files and fragments are not touched in this PR to prove that the `coalesce` works and is a no-op. A final PR will rename those


gaelgatelement (approved):

charts/matrix-stack/configs/matrix-rtc/sfu/config-overrides.yaml.tpl
```diff
@@ -75,14 +75,16 @@ turn:
 {{- end }}
 
 redis:
-{{- if .redis }}
-  address: {{ tpl .redis.host $root }}:{{ .redis.port | default 6379 }}
-{{- with .redis.db }}
+{{- if or .redisOrValkey .redis }}
+{{- with coalesce .redisOrValkey .redis }}
```
gaelgatelement, line 79: Why is `if or` needed before the `with coalesce` OOI ? It might deserve a comment explaining it as it feels surprising

benbz: Because we need to branch into the built-in Valkey if neither are set and Helm doesn't support `with`/`else` AFAICT.

We didn't have the `with` before as it did `.redis.host`, `.redis.port`, etc, but doing `(coalesce .redisOrValkey .redis).host` is horrible, hence the addition of the `with`

PR #1566, opened 2026-09-03
Synapse: Restore `/tmp` directory

### LINKED ISSUE #1565: 26.9.0 regression: synapse pod lost writable /tmp — media uploads 502 with readOnlyRootFilesystem
### Describe the bug

Media uploads (`POST /_matrix/media/v3/upload`) return 502 after upgrading from matrix-stack 26.8.1 to 26.9.0. Synapse crashes each request with:

```
builtins.FileNotFoundError: [Errno 2] No usable temporary directory found in ['/tmp', '/var/tmp', '/usr/tmp', '/']
```

### Root cause

In 26.8.1 the synapse pod mounted a writable `/tmp` emptyDir:

```yaml
      - mountPath: /tmp
        name: tmp
        readOnly: false
...
    - emptyDir:
        medium: Memory
      name: "tmp"
```

The 26.9.0 refactor moved all emptyDirs to the new `ephemeralStorages.*` value knobs and added default `tmp` (10Mi tmpfs) volumes for haproxy / element-web / element-admin / MAS / matrix-RTC / postgres / hookshot — but the synapse pod template lost its `/tmp` mount entirely. `templates/synapse/_synapse_pod.tpl` in 26.9.0 only references `synapse.ephemeralStorages.media`; there is no `tmp` knob for synapse at all.

Since the chart sets `readOnlyRootFilesystem: true` on the synapse container, Twisted's `tempfile.TemporaryFile()` (used to buffer request bodies above the in-memory threshold) fails, breaking every large upload. Small bodies still pass, so the stack otherwise looks healthy — haproxy just logs a stream of `502 SH--` while clients retry.

### Reproduce

- matrix-stack 26.9.0, synapse enabled
- `kubectl exec <synapse-pod> -- touch /tmp/test` → `Read-only file system`
- any media upload > in-memory threshold → 502 + the traceback above

### Versions

- chart matrix-stack 26.9.0 (worked on 26.8.1)
- diff between the two template versions confirms the volume+mount removal

### Workaround

Restore the volume via the existing extension points:

```yaml
synapse:
  extraVolumes:
    - name: synapse-tmp
      emptyDir:
        sizeLimit: 1Gi
  extraVolumeMounts:
    - name: synapse-tmp
      mountPath: /tmp
      readOnly: false
```

Would suggest wiring `synapse.ephemeralStorages.tmp` (default like the other components) so the workaround can be removed.

gaelgatelement:
Twisted requires a `/tmp` directory to buffer files larger than 100KB: https://github.com/twisted/twisted/blob/64c03b8b43da8c3c63f5c2b4c5a4506b5373f6b7/src/twisted/web/http.py#L785-L798

This restores the `/tmp` directory which was wrongly removed from 26.9.0 and caused #1565 

This adds 2 new values:

 - `synapse.ephemeralStorages.nonMediaTmp` handles the `/tmp` directory properties of non-media workers.
 - `synapse.media.ephemeralStorages.media.tmp` handles the `/tmp` directory properties of media workers.


charts/matrix-stack/source/synapse.yaml.j2
```diff
@@ -107,7 +109,7 @@ replicas: 1
 {{- sub_schema_values.affinity() }}
 {{- sub_schema_values.podSecurityContext(user_id='10091', group_id='10091') }}
 {{- sub_schema_values.resources(requests_memory='100Mi', requests_cpu='100m', limits_memory='4Gi') }}
-{{- sub_schema_values.ephemeralStorages([("media", "media store emptyDir fallback (/media)", "100Mi", "Memory"),]) }}
+{{- sub_schema_values.ephemeralStorages([("media", "media store emptyDir fallback (/media)", "100Mi", "Memory"),("nonMediaTmp", "non-media tmp directory. See .media.ephemeralStorages.tmp for tmp directory for media uploads", "10Mi", "Memory"),]) }}
```
benbz: The `media` `emptyDir` now looks quite bonkers. I wonder whether we should move `media_store_path` from `synapse-04-homeserver-overrides.yaml.tpl` into `synapse-05-process-specific.yaml.tpl` and vary it between `/media/media_media_store` and `/tmp/media_store` depending on whether that process is responsible for media or not. That way we only need 1 ephemeral storage here

tests/manifests/test_ephemeral_storages.py
```diff
@@ -19,37 +24,54 @@ def _assert_empty_dir_volumes(
...
 
     for volume in volumes:
         if "emptyDir" not in volume:
             continue
         volume_name = volume["name"]
 
-        assert deployable_details.has_ephemeral_storage, (
-            f"{pod_template_details.manifest_id}: found an emptyDir volume '{volume_name}' but deployable is not "
-            "defined with ephemeral storage"
-        )
```
benbz, line 31: Not quite sure I understand why this has moved

tests/manifests/test_ephemeral_storages.py
```diff
@@ -92,86 +114,36 @@ async def test_ephemeral_storage_passed_through(values, base_values, make_templa
...
-
-
-def _non_default_medium(default_medium: str) -> str:
-    # "Memory" -> "" ; "" -> "Memory"
-    return "" if default_medium == "Memory" else "Memory"
-
-
-@pytest.mark.parametrize("values_file", values_files_to_test)
-@pytest.mark.asyncio_cooperative
-async def test_ephemeral_storage_medium_configurable(values, base_values, make_templates):
```
benbz, line 121: This is disappearing as you've merged all it tests with `test_ephemeral_storage_passed_through`? Does that test need a rename to make this clearer?

tests/manifests/test_ephemeral_storages.py
```diff
@@ -92,86 +114,36 @@ async def test_ephemeral_storage_passed_through(values, base_values, make_templa
...
-    """Each ephemeral storage's medium is configurable in both directions.
-
-    Every emptyDir volume is set to the non-default of its own default medium
-    """
-    # Configure the shared rendered-config volume directly, preserving its sizeLimit.
-    rendered_config_default_medium = base_values["matrixTools"]["ephemeralStorages"]["renderedConfig"]["medium"]
-    rendered_config_non_default = _non_default_medium(rendered_config_default_medium)
-    values.setdefault("matrixTools", {}).setdefault("ephemeralStorages", {}).setdefault("renderedConfig", {})[
-        "medium"
-    ] = rendered_config_non_default
```
benbz, line 131: I think we've lost toggling the rendered config medium

tests/manifests/test_ephemeral_storages.py
```diff
@@ -19,37 +24,54 @@ def _assert_empty_dir_volumes(
...
+            expected_ephemeral_volume = expected_ephemeral_storages[deployable_details.ephemeral_storages[volume_name]]
+
+            # Special case for Synapse: The process responsible for media needs a larger /tmp to
+            # store buffered uploaded media. It is configured through synase.media.ephemeralStorages.tmp
+            if (
+                deployable_details.name == "synapse" and not synapse_media_worker_enabled
+            ) or deployable_details.name == "synapse-media-repository":
+                expected_tmp_dir = base_values["synapse"]["media"]["ephemeralStorages"]["tmp"] | values.get(
+                    "synapse", {}
+                ).get("media", {}).get("ephemeralStorages", {}).get("tmp", {})
```
benbz:
```suggestion
                default_media_tmp_dir_config = base_values["synapse"]["media"]["ephemeralStorages"]["tmp"]
                configured_media_tmp_dir_config = values.get("synapse", {}).get("media", {}).get("ephemeralStorages", {}).get("tmp", {})
                expected_tmp_dir = default_media_tmp_dir_config | configured_media_tmp_dir_config
```

Or something similar as I'm finding this not particularly readable.

tests/manifests/test_ephemeral_storages.py
```diff
@@ -92,86 +114,36 @@ async def test_ephemeral_storage_passed_through(values, base_values, make_templa
...
     iterate_deployables_parts(
-        set_ephemeral_storage_with_non_default_medium,
-        lambda deployable_details: deployable_details.has_ephemeral_storage,
+        set_ephemeral_storage, lambda deployable_details: deployable_details.has_ephemeral_storage
     )
 
+    counter += 1
+    expected_synapse_media_tmp_dir = base_values["synapse"]["media"]["ephemeralStorages"]["tmp"] | values.get(
+        "synapse", {}
+    ).get("media", {}).get("ephemeralStorages", {}).get("tmp", {})
```
benbz: Same comment about splitting into intermediate variables as above

tests/manifests/test_ephemeral_storages.py
gaelgatelement: `test_ephemeral_storage_passed_through` feels clear enough to me? I think it was a mistake to distinguish them in the first place.

charts/matrix-stack/source/synapse.yaml.j2
```diff
@@ -21,9 +21,11 @@ checkConfigHook:
 ## Configures the media store for Synapse
 media:
 {{- sub_schema_values.persistentVolumeClaim("storage") | indent(2) }}
+{{- sub_schema_values.ephemeralStorages([("tmp", "media tmp directory, used to upload files larger than 100KB", "1Gi", "Memory"),]) | indent(2) }}
 
   ## The maximum size (in bytes ending in M or K) that Synapse will accept for media uploads
   ## You may need to adjust your ingress controller to also allow uploads of this size
+  ## You may also need to adjust your .media.ephemeralStorages.tmp directory sizeLimit to support buffering of large upload files
```
benbz:
```suggestion
  ## You may also need to adjust your synapse.media.ephemeralStorages.tmp directory sizeLimit to support buffering of large upload files
```

charts/matrix-stack/source/synapse.yaml.j2
```diff
@@ -107,7 +109,7 @@ replicas: 1
 {{- sub_schema_values.affinity() }}
 {{- sub_schema_values.podSecurityContext(user_id='10091', group_id='10091') }}
 {{- sub_schema_values.resources(requests_memory='100Mi', requests_cpu='100m', limits_memory='4Gi') }}
-{{- sub_schema_values.ephemeralStorages([("media", "media store emptyDir fallback (/media)", "100Mi", "Memory"),]) }}
+{{- sub_schema_values.ephemeralStorages([(("nonMediaTmp", "non-media tmp directory. See .media.ephemeralStorages.tmp for tmp directory for media uploads", "10Mi", "Memory"),]) }}
```
benbz:
```suggestion
{{- sub_schema_values.ephemeralStorages([(("nonMediaTmp", "non-media tmp directory. See synapse.media.ephemeralStorages.tmp for tmp directory for media uploads", "10Mi", "Memory"),]) }}
```

tests/manifests/test_ephemeral_storages.py
gaelgatelement: No good reason, let's rollback this move.

newsfragments/1566.deprecated.md
```diff
@@ -0,0 +1 @@
+Synapse: Drop `.synapse.ephemeralStorages.media` has non-media processes now mount the media store in `/tmp`.
```
benbz:
```suggestion
Synapse: Drop `synapse.ephemeralStorages.media` as non-media processes now mount the media store in `/tmp`.
```

charts/matrix-stack/templates/synapse/_helpers.tpl
```diff
@@ -258,11 +258,11 @@ env:
 04-homeserver-overrides.yaml: |
 {{- (tpl ($root.Files.Get "configs/synapse/synapse-04-homeserver-overrides.yaml.tpl") (dict "root" $root "context" (mustMergeOverwrite ($root.Values.synapse | deepCopy) (dict "isHook" $isHook)))) | nindent 2 }}
 05-main.yaml: |
-{{- (tpl ($root.Files.Get "configs/synapse/synapse-05-process-specific.yaml.tpl") (dict "root" $root "context" (dict "processType" "main"))) | nindent 2 }}
+{{- (tpl ($root.Files.Get "configs/synapse/synapse-05-process-specific.yaml.tpl") (dict "root" $root "context" (dict "processType" "main" "isHook" $isHook))) | nindent 2 }}
```
benbz, line 261: This isn't needed AFAICT?

gaelgatelement: The variable was marked as required : https://github.com/element-hq/ess-helm/pull/1566/changes#diff-0b384671229dc9966e5f44595cf5c5a8b702c8ca402b53cdfa88773f92f433fbR10

benbz: Oh right I didn't read the filename fully, I thought it was a different one. Sorry

benbz (approved):

PR #1580, opened 2026-09-10
Upgrade Synapse to v1.161.0

benbz:
Draft as RC. Changelog written as per release.

https://github.com/element-hq/synapse/blob/release-v1.161/CHANGES.md


charts/matrix-stack/configs/synapse/path_map_file_get.tpl
```diff
@@ -10,6 +10,7 @@ SPDX-License-Identifier: AGPL-3.0-only
 # A map file that is used in haproxy config to map from matrix paths to the
 # named backend. The format is: path_regexp backend_name
 {{ if dig "client-reader" "enabled" false $root.Values.synapse.workers }}
+^/_matrix/client/unstable/org.matrix.msc4140/delayed_events/[^/]+$
```
benbz: This isn't in the public documentation yet. I've raised this in https://github.com/element-hq/synapse/pull/20210 as my reading of https://github.com/element-hq/synapse/pull/19926 is that the `GET` version of this path is workerisable

benbz: https://github.com/element-hq/synapse/pull/20210 has been merged and documents that this is workerisable

newsfragments/1580.changed.md
```diff
@@ -0,0 +1,9 @@
+Update Synapse to v1.161.0.
+
+Highlights:
```
jaywink, line 3: Is there a reason to leave out the two other features, ie:

* Don't validate signatures with unknown algorithms for master keys, and allow updates to signatures. (https://github.com/element-hq/synapse/issues/19915)
* Add a config option that limits the time period in which local users can redact their own messages. (https://github.com/element-hq/synapse/issues/20138)

Maybe the first one feels a bit internal, but the second one feels like it might be worth drawing attention to?

newsfragments/1580.changed.md
```diff
@@ -0,0 +1,9 @@
+Update Synapse to v1.161.0.
```
jaywink, line 1: As per https://github.com/element-hq/synapse/blob/develop/docs/upgrade.md#deprecation-of-matrix_rtclivekit_service_url should we already add the new `url` key, along with the deprecated one, so we support both early?

newsfragments/1580.changed.md
benbz: I'll update to include the 2nd. What to include in `Highlights` is a really subjective thing for the PR author/reviewer, i.e. what might be people/our clients find interesting

newsfragments/1580.changed.md
benbz: Good catch, will do so

benbz: Ah no.

> Clients that support url will use the Client-Server API to (indirectly) interact with the LiveKit authorization service. The service needs to be set up as an application service in order to support these endpoints. See https://github.com/element-hq/lk-jwt-service for further details.

The authoriser service isn't an AS in ESS yet. We should absolutely do this as part of that PR

jaywink (approved):

PR #1584, opened 2026-09-10
Allow the syn2mas process to stay in stage 2 for multiple deployments

benbz:
Update the `syn2mas` migration process (as originally implemented in https://github.com/element-hq/ess-helm/pull/454) so that `helm upgrade` works for multiple deployments in [stage 2](https://github.com/element-hq/ess-helm/blob/main/docs/syn2mas.md) without having to move immediately onto stage 3 (set `matrixAuthenticationService.syn2mas.enabled: false`).

We must move/gate the `syn2mas check` as it complains about a non-empty database. We also move the `syn2mas --dry-run` so that a) it is after the `syn2mas check` and b) also runs immediately before a real migration.

Requires a `matrix-tools` dance to complete.

Based on top of #1583 so as to benefit from the refactoring and improve clarity of the deployment markers states

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -127,6 +127,8 @@ jobs:
         else
           echo "The integration test does not exist in previous ref, skipping..."
         fi
+      # syn2mas is a special case that cannot run twice by design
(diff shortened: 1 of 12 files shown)


jaywink (approved):
Looks good as far as I can tell stepping through the commits 👍

PR #1586, opened 2026-09-10
Upgrade to MatrixRTC Authoriser v0.7.0 and support external Redis/Valkey against it

benbz:
https://github.com/element-hq/lk-jwt-service/releases/tag/v0.7.0.

Then uses https://github.com/element-hq/lk-jwt-service/pull/233 so that we can fully support an external Redis/Valkey with auth. i.e. what was left off of https://github.com/element-hq/ess-helm/pull/1555. Unlike with https://github.com/element-hq/ess-helm/pull/1236 we do not need to quote or URL encode the password as Rust's URL `set_password` does that for us

Given that this is being implemented after https://github.com/element-hq/ess-helm/pull/1562 we do not introduce a deprecated `matrixRTC.redis`, we only support `matrixRTC.redisOrValkey`


jaywink (approved):
Looks good

PR #1589, opened 2026-09-11
Well Known: always use HTTPS when redirecting to Element Web

benbz:
https://github.com/element-hq/ess-helm/pull/231 introduced the option to redirect to Element Web when a path on `<serverName>` wasn't matched by a well-known file.

It included a conditional as to whether it redirected to HTTP or HTTPS. However a) Element Web requires TLS (to run things like service workers in a secure context), b) all other links in ESS Pro are HTTPS only and c) misunderstands why `element-io.ess-library.ingress.tlsHostsSecret` might not return anything. For c) this because TLS is expected to be terminated upstream of the cluster but TLS will still be used by users connecting to the endpoint.

Remove this conditional and always use HTTPS


jaywink (approved):
Looks cleaner 👍

